PR #29, opened 2017-09-14
Add .travis.yml file

murgatroid99:


nicolasnoble:
Seems like gulp needs ES6, so this won't work with node 4 and 5.

PR #31, opened 2017-09-14
Add Kokoro test script

murgatroid99:
CC @matt-kwong

### DIFF
--- a/kokoro-linux.sh
+++ b/kokoro-linux.sh
@@ -1,3 +1,18 @@
+#!/bin/sh
+# Copyright 2017 gRPC authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 # Install NVM
 curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.4/install.sh | bash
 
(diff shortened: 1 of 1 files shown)


nicolasnoble:
cc @matt-kwong

kokoro-linux.sh
```diff
@@ -0,0 +1,20 @@
+# Install NVM
```
nicolasnoble: We'll need the copyright blurb (that's bureaucracy), and #!/bin/sh (that's more about making this work properly).

kokoro-linux.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Install dependencies and link packages together
+npm install
+gulp setup
+
+for version in ${node_versions[@]}
+do
+  nvm install $version
+  # Rebuild libraries and run tests
+  gulp native.test
```
nicolasnoble: Wouldn't you need to do npm install between each change of node version ?

matt-kwong:
This should work for both Linux and MacOS, right?

kokoro-linux.sh
nicolasnoble: The file you're adding is 644. It needs to be 755.

nicolasnoble:
It should work on both Linux and Mac, yes. Windows might be a bit more tricky right off the bat.

murgatroid99:
I think so

kokoro-linux.sh
murgatroid99: I think I forget that more often than I remember.

kokoro-linux.sh
murgatroid99: As mentioned, that command includes a rebuild, and I think that the only thing that depends on the Node version is the grpc extension build itself.

nicolasnoble:
So, #29 shows that gulp won't work without ES6 support, so this script isn't looking like it's going to work before Node 6.

murgatroid99:
It looks like it's a specific issue with some ES6 features. I might just be able to fix it.

nicolasnoble:
I see, it's a problem in our gulp script files, not with gulp itself. Let's remove node 4 and 5 from that PR for now then, and fix it later.

PR #33, opened 2017-09-15
Add Jenkins XML reporting to tests

murgatroid99:
The XML file names are going to be based on timestamp and not anything useful, but it's better than nothing.

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -35,6 +35,7 @@ mkdir -p reports
 
 for version in ${node_versions}
 do
+  git clean -f -d -x
   # Install and setup node for the version we want.
   set +ex
   echo "Switching to node version $version"
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -42,7 +42,7 @@ do
   nvm use $version
   set -ex
 
-  mkdir p "reports/node$version"
+  mkdir -p "reports/node$version"
 
   # Install dependencies and link packages together.
   npm install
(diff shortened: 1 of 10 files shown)


murgatroid99:
I redid this change with the new script

run-tests.sh
```diff
@@ -46,7 +46,8 @@ do
   ./node_modules/.bin/gulp setup
 
   # Rebuild libraries and run tests.
-  ./node_modules/.bin/gulp native.test || FAILED="true"
+  JUNIT_REPORT_PATH="node$version/" JUNIT_REPORT_STACK=1 ./node_modules/.bin/gulp native.test || FAILED="true"
+  rename 's/([^\/]+).xml/$1_sponge_log.xml/' "node$version"/*
```
nicolasnoble: ./run-tests.sh: line 50: rename: command not found

nicolasnoble: Suggest

```
cd node$version
for file in * ; do
  mv $f `echo $f | sed 's/\(.*\)\.xml/\1_sponge_log.xml/'`
fi
```

(also `.xml` in the regex is wrong, it should be `\.xml` :-) )

matt-kwong:
It looks like Sponge has a weird way of parsing the .xml files. I think it interprets multiple .xml in the same directory as reruns of a flaky test. Can this be changed so that each .xml file has its own subdirectory or that the .xml files are combined?

matt-kwong:
You can see the current directory structure of the reports generated here: https://pantheon.corp.google.com/storage/browser/grpc-testing-kokoro-prod/test_result_public/prod/grpc/node/pull_request/linux/220/20170919-142854/github/grpc-node/reports/

PR #37, opened 2017-09-16
Adding support for Node 4 and 5.

nicolasnoble:


packages/grpc-js-core/gulpfile.js
```diff
@@ -31,7 +31,7 @@ const files = !util.env.file ? [] :
       Array.isArray(util.env.file) ? util.env.file : [util.env.file];
 
 // If --dev is passed, override certain ts config options
-let tsDevOptions = {};
+var tsDevOptions = {};
```
kjin: Doesn't Node 4 support `let`? I always considered `const`/`let` a single feature.

packages/grpc-js-core/gulpfile.js
```diff
@@ -58,10 +58,12 @@ function makeCompileFn(globs) {
   const copyGlob = globs.copy || '!(**/*)';
   return () => {
     const tsProject = typescript.createProject(tsconfigPath, tsDevOptions)();
-    const { dts, js } = gulp.src(transpileGlob, { base: jsCoreDir })
+    const data = gulp.src(transpileGlob, { base: jsCoreDir })
           .pipe(sourcemaps.init())
           .pipe(tsProject)
           .on('error', onError);
+    const dts = data.dts
```
kjin: Lines 65/66 need semicolons

murgatroid99:
Instead of changing `let` to `var`, you should add `"use strict";` to the beginning of each file. The error messages on Node 4 and 5 mentioned that `let` is not available "outside of strict mode".

nicolasnoble:
Wouldn't using strict cause more breakages however ? I can try that way instead.

murgatroid99:
Strict mode is like compiling with `-Werror`. It might cause more failures, but it's still a good idea.

nicolasnoble:
Kokoro linux and mac both green, Travis about to be green too.

PR #40, opened 2017-09-18
Adding Windows-Kokoro support.

nicolasnoble:


nicolasnoble:
We'll indeed do that later.

PR #41, opened 2017-09-19
Remove some unnecessary files from the native npm package

murgatroid99:
This is for grpc/grpc#12590, but it may not fix it because of npm/npm#4479.


packages/grpc-native-core/package.json
```diff
@@ -71,9 +70,13 @@
     "deps/grpc/src/core",
     "deps/grpc/src/boringssl",
     "deps/grpc/src/zlib",
-    "deps/grpc/third_party/nanopb",
+    "deps/grpc/third_party/nanopb/pb_common.c",
+    "deps/grpc/third_party/nanopb/pb_decode.c",
+    "deps/grpc/third_party/nanopb/pb_encode.c",
```
nicolasnoble: Is it possible to use globs here instead ? Like deps/grpc/third_party/nanopb/**.{c,h}

PR #44, opened 2017-09-19
Script changes to allow testing from grpc/grpc

matt-kwong:


murgatroid99:
You might also have to add "--unsafe-perm" to npm install calls in a couple of the gulpfiles

matt-kwong:
@murgatroid99 With the changes from #33 , should I just remove `npm install --unsafe-perm` all together from `run-tests.sh`? The npm installs in the gulpfiles have been changed to add `--unsafe-perm`.

murgatroid99:
The `npm install` in `run-tests.sh` was moved, but it's still there. You should still change it just to be safe.

matt-kwong:
I was able to get testing from grpc/grpc to work locally with this branch, PTAL.

PR #45, opened 2017-09-20
Disable warning in recent GCC versions that hits BoringSSL

murgatroid99:
This fixes grpc/grpc#12536


nicolasnoble:
We should monitor this when they fix it upstream properly and then disable the flag.

PR #49, opened 2017-09-22
Add license notice to several script files

murgatroid99:


nicolasnoble:
The upstream copyright checker is a good thing that works fine. We might be able to invoke it in a sanity test suite we should add. cc @matt-kwong

PR #51, opened 2017-09-28
Separate out packages. Upgrade new protobufjs package to Protobuf.js 6.8

murgatroid99:
This adds the basic skeleton of the future package structure. It also uses the Protobuf.js 6 code from grpc@1.3.0 to depend on Protobuf.js 6.8 instead of Protobuf.js 5.

Pretty much all of the code is copied from grpc-native-core. The code duplication should be temporary, just until we publish grpc 2.0


packages/grpc-surface/index.js
```diff
@@ -0,0 +1,147 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+'use strict';
+
+const util = require(util);
```
kjin: `require('util')`

kjin:
We should have a script that automatically populates dependencies from relative paths within the repo (e.g. grpc-js-core on grpc-js).

packages/grpc-js/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "grpc-js",
+  "name": "@grpc/js",
```
kjin: Are we really going to be dropping the `grpc-` prefix? I feel like any package that could conceivably be brought as a dependency should retain the prefix, even if it's part of the namespace.

murgatroid99: The scope takes the place of the prefix. Why do you think we would need both?

kjin: It just seems strange that `js` essentially becomes the name of the module after we look past the org scope. It's possible that the gRPC org publishes other, related tools under the `@grpc` scope in the future -- say, for example, our own protobuf as `@grpc/protobuf`; if that ends up being the case, it'd be strange that a module called 'js' co-exists at the same level of citizenship as `protobuf` within the `@grpc` scope.

if `@grpc` is used *purely* for helper packages I'm OK with this. But I still think the unscoped package names are more clear.

kjin: Talked offline -- I'm OK with this change

murgatroid99:
The gulp scripts still don't create some of the links, because of problems caused by npm/npm#18835. But I think it's good enough to merge.

PR #52, opened 2017-09-29
Add TypeScript typings to gRPC native core package

Crevil:
Following @murgatroid99's request I've moved the PR https://github.com/grpc/grpc/pull/12755 to this repo instead.

**From the original PR:**

This is a follow up of https://github.com/grpc/grpc/pull/11020. I've accidentally deleted the old gRPC fork I had as I had totally forgotten this PR.

This is a remake with all the latest API changes and the comments from the former PR added.
Further more, I've created a test repo to verify that the type comply with the documented API. It is located here for now: https://github.com/lunarway/grpc-ts-test
If it can fit into the structure of the repo, we could incorporate it to make sure the type and tests stay up to date? Let me know if there is anything I can do in that case.

From the initial PR:

> This PR closes https://github.com/grpc/grpc/issues/8233
> It adds the most basic typings for TypeScript of the public API of grpc for Node.js.
> 
> All details are not defined in this, as I'm no expert in reading the native parts of the module. I've used these definitions for a while and trusts the API. Let me know if anything needs to be corrected or otherwise modified.
> 
> All comments etc. are from the existing documentations.
> 
> Example usage:
> 
> import * as grpc from 'grpc';
> 
> const server = new grpc.Server();
> server.bind('0.0.0.0:7200', grpc.ServerCredentials.createInsecure());
> server.start();

### DIFF
--- a/packages/grpc-native-core/index.d.ts
+++ b/packages/grpc-native-core/index.d.ts
@@ -1,7 +1,7 @@
 declare module "grpc" {
-  /// <reference types="node" />
   import { Message, Service } from "protobufjs";
   import { SecureContext } from "tls";
+  import { Duplex, Readable, Writable } from "stream";
 
   /**
    * Load a ProtoBuf.js object as a gRPC object.
@@ -308,7 +308,7 @@ declare module "grpc" {
    * A stream that the server can read from. Used for calls that are streaming
    * from the client side.
    */
-  export class ServerReadableStream {
+  export class ServerReadableStream extends Readable {
     /**
      * Indicates if the call has been cancelled
      */
@@ -343,7 +343,7 @@ declare module "grpc" {
    * A stream that the server can write to. Used for calls that are streaming
    * from the server side.
    */
-  export class ServerWriteableStream {
+  export class ServerWriteableStream extends Writable {
     /**
      * Indicates if the call has been cancelled
      */
@@ -383,7 +383,7 @@ declare module "grpc" {
    * A stream that the server can read from or write to. Used for calls
    * with duplex streaming.
    */
-  export class ServerDuplexStream {
+  export class ServerDuplexStream extends Duplex {
     private constructor();
 
     /**
@@ -1106,7 +1106,7 @@ declare module "grpc" {
    * A stream that the client can read from. Used for calls that are streaming
    * from the server side.
    */
-  export class ClientReadableStream {
+  export class ClientReadableStream extends Readable {
     private constructor();
 
     /**
--- a/packages/grpc-native-core/index.d.ts
+++ b/packages/grpc-native-core/index.d.ts
@@ -1,0 +1,1214 @@
+declare module "grpc" {
+  import { Service } from "protobufjs";
+
+  /**
+   * Load a ProtoBuf.js object as a gRPC object.
+   * @param value The ProtoBuf.js reflection object to load
+   * @param options Options to apply to the loaded file
+   * @return The resulting gRPC object.
+   */
+  export function loadObject(value: object, options?: LoadObjectOptions): GrpcObject;
+
+  /**
+   * Options for loading proto object as gRPC object
+   * @param {(number|string)=} [options.protobufjsVersion='detect'] 5 and 6
+   *     respectively indicate that an object from the corresponding version of
+   *     Protobuf.js is provided in the value argument. If the option is 'detect',
+   *     gRPC will guess what the version is based on the structure of the value.
+   */
+  export interface LoadObjectOptions {
+    /**
+     * Deserialize bytes values as base64 strings instead of Buffers.
+     * Defaults to `false`.
+     */
+    binaryAsBase64?: boolean;
+
+    /**
+     * Deserialize long values as strings instead of objects.
+     * Defaults to `true`.
+     */
+    longsAsStrings?: boolean;
+
+    /**
+     * Deserialize enum values as strings instead of numbers. Only works with
+     * Protobuf.js 6 values.
+     * Defaults to `true`.
+     */
+    enumsAsStrings?: boolean;
+
+    /**
+     * use the beta method argument order for client methods, with optional
+     * arguments after the callback. This option is only a temporary stopgap
+     * measure to smooth an API breakage. It is deprecated, and new code
+     * should not use it.
+     * Defaults to `false`
+     */
+    deprecatedArgumentOrder?: boolean;
+
+    /**
+     * 5 and 6 respectively indicate that an object from the corresponding
+     * version of Protobuf.js is provided in the value argument. If the option
+     * is 'detect', gRPC wll guess what the version is based on the structure
+     * of the value.
+     */
+    protobufjsVersion?: 5 | 6 | "detect";
+  }
+
+  /**
+   * Map from `.proto` file.
+   * - Namespaces become maps from the names of their direct members to those member objects
+   * - Service definitions become client constructors for clients for that service. They also
+   *   have a service member that can be used for constructing servers.
+   * - Message definitions become Message constructors like those that ProtoBuf.js would create
+   * - Enum definitions become Enum objects like those that ProtoBuf.js would create
+   * - Anything else becomes the relevant reflection object that ProtoBuf.js would create
+   */
+  export interface GrpcObject {
+    [namespace: string]: {
+      [service: string]: typeof Client,
+    };
+  }
+
+  /**
+   * Load a gRPC object from a .proto file.
+   * @param filename The file to load
+   * @param format The file format to expect. Defaults to 'proto'
+   * @param options Options to apply to the loaded file
+   * @return The resulting gRPC object
+   */
+  export function load(filename: Filename, format?: "proto" | "json", options?: LoadOptions): GrpcObject;
+
+  /**
+   * A filename
+   */
+  export type Filename = string | { root: string, file: string };
+
+  /**
+   * Options for loading proto file as gRPC object
+   */
+  export interface LoadOptions {
+    /**
+     * Load this file with field names in camel case instead of their original case.
+     * Defaults to `false`.
+     */
+    convertFieldsToCamelCase?: boolean;
+
+    /**
+     * Deserialize bytes values as base64 strings instead of Buffers.
+     * Defaults to `false`.
+     */
+    binaryAsBase64?: boolean;
+
+    /**
+     * Deserialize long values as strings instead of objects.
+     * Defaults to `true`.
+     */
+    longsAsStrings?: boolean;
+
+    /**
+     * Use the beta method argument order for client methods, with optional
+     * arguments after the callback. This option is only a temporary stopgap
+     * measure to smooth an API breakage. It is deprecated, and new code
+     * should not use it.
+     * Defaults to `false`
+     */
+    deprecatedArgumentOrder?: boolean;
+  }
+
+  /**
+   * Sets the logger function for the gRPC module. For debugging purposes, the C
+   * core will log synchronously directly to stdout unless this function is
+   * called. Note: the output format here is intended to be informational, and
+   * is not guaranteed to stay the same in the future.
+   * Logs will be directed to logger.error.
+   * @param logger A Console-like object.
+   */
+  export function setLogger(logger: Console): void;
+
+  /**
+   * Sets the logger verbosity for gRPC module logging. The options are members
+   * of the grpc.logVerbosity map.
+   * @param verbosity The minimum severity to log
+   */
+  export function setLogVerbosity(verbosity: number): void;
+
+  /**
+   * Server object that stores request handlers and delegates incoming requests to those handlers
+   */
+  export class Server {
+    /**
+     * Constructs a server object that stores request handlers and delegates
+     * incoming requests to those handlers
+     * @param options Options that should be passed to the internal server
+     *     implementation
+     * ```
+     * var server = new grpc.Server();
+     * server.addProtoService(protobuf_service_descriptor, service_implementation);
+     * server.bind('address:port', server_credential);
+     * server.start();
+     * ```
+     */
+    constructor(options?: object);
+
+    /**
+     * Start the server and begin handling requests
+     */
+    start(): void;
+
+    /**
+     * Registers a handler to handle the named method. Fails if there already is
+     * a handler for the given method. Returns true on success
+     * @param name The name of the method that the provided function should
+     *     handle/respond to.
+     * @param handler Function that takes a stream of
+     *     request values and returns a stream of response values
+     * @param serialize Serialization function for responses
+     * @param deserialize Deserialization function for requests
+     * @param type The streaming type of method that this handles
+     * @return True if the handler was set. False if a handler was already
+     *     set for that name.
+     */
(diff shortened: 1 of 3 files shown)


packages/grpc-native-core/index.d.ts
```diff
@@ -0,0 +1,1214 @@
...
+   * @param value The value to serialize
+   * @return The value serialized as a byte sequence
+   */
+  type serialize = (value: any) => Buffer;
+
+  /**
+   * Callback function passed to server handlers that handle methods with
+   * unary responses.
+   */
+  type sendUnaryData = (error: ServiceError | null, value: any, trailer?: Metadata, flags?: number) => void;
```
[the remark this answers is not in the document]
Crevil: Yeah, that sure looks weird... Let me get to that. And I will give the API a look in `grpc-js-core` as well to make sure we are in line.

packages/grpc-native-core/index.d.ts
```diff
@@ -0,0 +1,1214 @@
...
+
+    /**
+     * Binds the server to the given port, with SSL disabled if creds is an
+     * insecure credentials object
+     * @param port The port that the server should bind on, in the format
+     * "address:port"
+     * @param creds Server credential object to be used for SSL. Pass an
+     * insecure credentials object for an insecure port.
+     */
+    bind(port: string, creds: ServerCredentials): void;
```
murgatroid99: This does not actually return `void`; it returns a `number` that is equal to the bound port, or 0 if the operation failed. The lack of documentation of this fact is an error.

packages/grpc-native-core/index.d.ts
```diff
@@ -0,0 +1,1214 @@
...
+     * The request message from the client
+     */
+    request: any;
+
+    /**
+     * An EventEmitter. Used for unary calls.
+     * @param call The call object associated with the request
+     * @param metadata The request metadata from the client
+     */
+    constructor(call: Call, metadata: Metadata);
```
murgatroid99: If this isn't necessary, I would prefer to omit it. This constructor is not explicitly exposed in the API. In addition, the `Call` type is defined elsewhere as the union of the client call object types, but that is not the correct type for this argument. The argument to this constructor is another type that is also not exposed in the API.

packages/grpc-native-core/index.d.ts
```diff
@@ -0,0 +1,1214 @@
...
+    metadata: Metadata;
+
+    /**
+     * A stream that the server can read from. Used for calls that are streaming
+     * from the client side.
+     * @param call The call object to read data with
+     * @param metadata The request metadata from the client
+     * @param deserialize Deserialization function for reads
+     */
+    constructor(call: Call, metadata: Metadata, deserialize: deserialize);
```
murgatroid99: As mentioned above, this constructor should not be in the type definition

packages/grpc-native-core/index.d.ts
```diff
@@ -0,0 +1,1214 @@
...
+    request: any;
+
+    /**
+     * A stream that the server can write to. Used for calls that are streaming
+     * from the server side.
+     * @param call The call object to send data with
+     * @param metadata The request metadata from the client
+     * @param serialize Serialization function for writes
+     */
+    constructor(call: Call, metadata: Metadata, serialize: serialize);
```
murgatroid99: As mentioned above, this constructor should not be in the type definition

packages/grpc-native-core/index.d.ts
```diff
@@ -0,0 +1,1214 @@
...
+  export class ServerDuplexStream {
+    /**
+     * A stream that the server can read from or write to. Used for calls
+     * with duplex streaming.
+     * @param call Call object to proxy
+     * @param metadata The request metadata from the client
+     * @param serialize Serialization function for requests
+     * @param deserialize Deserialization function for requests
+     */
+    constructor(call: any, metadata: Metadata, serialize: serialize, deserialize: deserialize);
```
murgatroid99: As mentioned above, this constructor should not be in the type definition

packages/grpc-native-core/index.d.ts
```diff
@@ -0,0 +1,1214 @@
...
+   * `grpc.status.OK`, then the request has completed successfully.
+   * Otherwise, the request has failed, `details` will contain a description of
+   * the error. Either way, `metadata` contains the trailing response metadata
+   * sent by the server when it finishes processing the call.
+   */
+  export interface StatusObject {
+    /**
+     * The error code, a key of `grpc.status`
+     */
+    code: number;
```
murgatroid99: The type of `code` should be the `status` enum defined below.

packages/grpc-native-core/index.d.ts
```diff
@@ -0,0 +1,1214 @@
...
+     * @return The credentials object
+     */
+    createFromMetadataGenerator(metadataGenerator: metadataGenerator): CallCredentials;
+
+    /**
+     * Create a gRPC credential from a Google credential object.
+     * @param googleCredential The Google credential object to use
+     * @return The resulting credentials object
+     */
+    createFromGoogleCredential(googleCredential: GoogleCredential): CallCredentials;
```
murgatroid99: I don't believe that `GoogleCredential` is defined in this file, which is not surprising because gRPC does not explicitly depend on the library that that type comes from. That library is `google-auth-library`. I think it would be OK to depend just on the type definitions for that library, if such a module exists.

packages/grpc-native-core/index.d.ts
```diff
@@ -0,0 +1,1214 @@
...
+   * Any client call type
+   */
+  type Call = ClientUnaryCall | ClientReadableStream | ClientWritableStream | ClientDuplexStream;
+
+  export class ClientUnaryCall {
+    /**
+     * An EventEmitter. Used for unary calls.
+     * @param call The call object associated with the request
+     */
+    constructor(call: any);
```
murgatroid99: As with the server-side call classes, I think these constructors should be omitted from the type definition.

packages/grpc-native-core/index.d.ts
```diff
@@ -0,0 +1,1214 @@
...
+     */
+    sendMetadata(responseMetadata: Metadata): void;
+  }
+
+  /**
+   * User provided method to handle client streaming methods on the server.
+   */
+  type handleClientStreamingCall = (call: ServerReadableStream, callback: sendUnaryData) => void;
+
+  export class ServerReadableStream {
```
murgatroid99: In the actual API, this class inherits from a Readable stream in object mode. Could the fact that those APIs are not in the type definitions cause problems for users? This applies to all of the streaming call types on both the client and server side, as well as the `UnaryClientCall` type, which inherits from EventEmitter.

packages/grpc-native-core/index.d.ts
```diff
@@ -0,0 +1,1214 @@
...
+   * @param logger A Console-like object.
+   */
+  export function setLogger(logger: Console): void;
+
+  /**
+   * Sets the logger verbosity for gRPC module logging. The options are members
+   * of the grpc.logVerbosity map.
+   * @param verbosity The minimum severity to log
+   */
+  export function setLogVerbosity(verbosity: number): void;
```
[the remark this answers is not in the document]
Crevil: You'r right. This could very well be the enum.

packages/grpc-native-core/index.d.ts
```diff
@@ -0,0 +1,1214 @@
...
+    addService(service: Service, implementation: { [name: string]: handleCall }): void;
+
+    /**
+     * Add a proto service to the server, with a corresponding implementation
+     * @deprecated Use `Server#addService` instead
+     * @param service The proto service descriptor
+     * @param implementation Map of method names to method implementation
+     * for the provided service.
+     */
+    addProtoService(service: Service, implementation: { [name: string]: handleCall }): void;
```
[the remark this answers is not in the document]
Crevil: Got it.

packages/grpc-native-core/index.d.ts
Crevil: I'll remove them.

packages/grpc-native-core/index.d.ts
Crevil: I'm not sure about this one. I think we need to explicitly inherit from those types as well otherwise something like this would cause an error.

```typescript
let c: ServerUnaryCall;
c.on("data", () => {});
```

To make this possible we need the Node.js typings. Should I include them as a dev dependency?

packages/grpc-native-core/index.d.ts
Crevil: I made a copy of the types into this file, see [line 864](https://github.com/grpc/grpc-node/pull/52/files/bdbbd5324956a118dacf21380a5e3d9eac6be62e#diff-84a425dc635cd3214797bf694fad81e8R864).
The lib is written in TypeScript so it ships with it's own typings (https://github.com/google/google-auth-library-nodejs/blob/master/package.json#L10)
How would you prefer this? I dep to the lib or a copy of the type?

murgatroid99: I see that copy now, and I think that's the right way to handle this. But that's not the right type. I think what I called `GoogleCredentials` actually corresponds to what's listed as `OAuth2Client` in that library. The important part of that interface, from the point of view of how this library uses it, is the `getRequestMetadata` method.

Crevil: Oh, that's my mistake then. I'll change it to the right one.

I've put som questions into the other comments you left, but I think they've been hidden by the other changes. Would you look into them as well?

packages/grpc-native-core/index.d.ts
```diff
@@ -0,0 +1,1176 @@
...
+   * Create a client with the given methods
+   */
+  export class Client {
+    /**
+     * A generic gRPC client. Primarily useful as a base class for generated clients
+     * @param address Server address to connect to
+     * @param credentials Credentials to use to connect to the server
+     * @param {Object} options Options to apply to channel creation
+     */
+    constructor(address: string, credentials: ChannelCredentials, options?: object)
```
connor4312: By the way, I creaed typings for the options [here](https://github.com/mixer/etcd3/blob/master/src/types/grpc.d.ts#L258)

packages/grpc-native-core/index.d.ts
murgatroid99: I'm OK with depending on the Node.js typings. Wouldn't we have to include them as a regular dependency, if this `index.d.ts` we're distributing in the release references it?

Also, it's not as simple as just inheriting from those types. Those types are explicitly written with the assumption that the stream is a byte stream, and it takes some intermediate type redefinitions to make them valid object stream types. You can see what I had to do for this in the other package here: https://github.com/grpc/grpc-node/blob/master/packages/grpc-js-core/src/object-stream.ts

murgatroid99: You shouldn't have to do as much as that, because you don't necessarily need strongly typed object streams that restrict inputs and outputs to the specific request and response types. But you at least need equivalents of the intermediate types to allow `any` in the inputs to `write` operations and the outputs from `read` operations.

packages/grpc-native-core/index.d.ts
Crevil: Great, I'll merge them into this file. Thanks!

packages/grpc-native-core/index.d.ts
Crevil: I'll look into the definitions and see what I can do with it.

packages/grpc-native-core/index.d.ts
murgatroid99: That's good to know; I would prefer not to include those in this PR. There's some complexity with handling the case where new arguments are added, and I think we'll be able to handle that better in the future after some other changes we're planning to make.

Crevil: Roger. We leave them out for now then.

packages/grpc-native-core/index.d.ts
Crevil: All right, I've looked into this. The tests allow for your requested API, so I think we are good for now.

Crevil:
I've identified the following differences from `grpc-js-core` package.

* The method `Client#getChannel(): Channel` [L773](https://github.com/grpc/grpc-node/blob/master/packages/grpc-native-core/src/client.js#L773) is not exposed  in `js-core`.
* Functions not found in `js-core`
  * `load` [src](https://github.com/grpc/grpc-node/blob/master/packages/grpc-native-core/index.js#L121)
  * `loadObject` [src](https://github.com/grpc/grpc-node/blob/master/packages/grpc-native-core/index.js#L73)
  * `setLogger` [src](https://github.com/grpc/grpc-node/blob/master/packages/grpc-native-core/index.js#L163)
  * `setLogVerbosity` [src](https://github.com/grpc/grpc-node/blob/master/packages/grpc-native-core/index.js#L184)

packages/grpc-native-core/index.d.ts
murgatroid99: I don't think the changes you made here quite work. As I mentioned, the builtin stream types are written for byte streams. They handle buffers and strings, not arbitrary objects. You need to override the method definitions, like I did in that object stream TypeScript file.

Also, it looks like you didn't change all of the client stream types.

Crevil: When looking into the Node.js typings it seems to allow the `any` type for both `read` results ([stream.Readable](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/index.d.ts#L5030)) and `write` messages ([stream.Writeable](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/index.d.ts#L5116)). 

See this test file where it's possible to write an object: https://github.com/lunarway/grpc-ts-test/blob/master/index.ts#L42

Let me know if I've misunderstood something here?

murgatroid99: On the readable side, you can see that the `data` event restricts the values to `Buffer | string`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/index.d.ts#L5067. The `Writable` type is actually fine.

Crevil: The [first overload](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/index.d.ts#L5065) of `on` allows for the `any` type when reading data events. 

```typescript
readableStream.on("data", (chunk: object) => {});
```

I've made a test case in this commit: https://github.com/lunarway/grpc-ts-test/commit/23a74d585e2701a993f0b874ad017152cbc41083

murgatroid99:
Thank you for your contribution. The typings file will be in grpc@1.7, which will be published soon.

ksaldana1:
Thanks a bunch @Crevil for taking the initiative on this and getting it done! I know I've been eagerly waiting for this, so I'm sure others are too. Thanks a bunch also to @murgatroid99 for the code reviews and @connor4312 for providing his typings file to build from.

Crevil:
Thank you for the great review @murgatroid99 .
Let me know if there is anything I can help you with in the future.

chrisalexander55:
Is this available in the latest build? I am unable to find a type definition file in node_modules/grpc directory. Thanks.

murgatroid99:
Today (hopefully) we will be publishing a prerelease package for the next version, which includes these changes.

PR #59, opened 2017-10-06
Implement client interceptors for grpc-native-core

drobertduke:
A NodeJS implementation of client-side interceptors, as described in
the proposal: https://github.com/grpc/proposal/pull/14

### DIFF
--- a/packages/grpc-native-core/src/client_interceptors.js
+++ b/packages/grpc-native-core/src/client_interceptors.js
@@ -1,0 +1,1343 @@
+/**
+ * @license
+ * Copyright 2018 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+/**
+ * Client Interceptors
+ *
+ * This module describes the interceptor framework for clients.
+ * An interceptor is a function which takes an options object and a nextCall
+ * function and returns an InterceptingCall:
+ *
+ * ```
+ * var interceptor = function(options, nextCall) {
+ *   return new InterceptingCall(nextCall(options));
+ * }
+ * ```
+ *
+ * The interceptor function must return an InterceptingCall object. Returning
+ * `new InterceptingCall(nextCall(options))` will satisfy the contract (but
+ * provide no interceptor functionality). `nextCall` is a function which will
+ * generate the next interceptor in the chain.
+ *
+ * To implement interceptor functionality, create a requester and pass it to
+ * the InterceptingCall constructor:
+ *
+ * `return new InterceptingCall(nextCall(options), requester);`
+ *
+ * A requester is a POJO with zero or more of the following methods:
+ *
+ * `start(metadata, listener, next)`
+ * * To continue, call next(metadata, listener). Listeners are described
+ * * below.
+ *
+ * `sendMessage(message, next)`
+ * * To continue, call next(message).
+ *
+ * `halfClose(next)`
+ * * To continue, call next().
+ *
+ * `cancel(message, next)`
+ * * To continue, call next().
+ *
+ * A listener is a POJO with one or more of the following methods:
+ *
+ * `onReceiveMetadata(metadata, next)`
+ * * To continue, call next(metadata)
+ *
+ * `onReceiveMessage(message, next)`
+ * * To continue, call next(message)
+ *
+ * `onReceiveStatus(status, next)`
+ * * To continue, call next(status)
+ *
+ * A listener is provided by the requester's `start` method. The provided
+ * listener implements all the inbound interceptor methods, which can be called
+ * to short-circuit the gRPC call.
+ *
+ * Three usage patterns are supported for listeners:
+ * 1) Pass the listener along without modification: `next(metadata, listener)`.
+ *   In this case the interceptor declines to intercept any inbound operations.
+ * 2) Create a new listener with one or more inbound interceptor methods and
+ *   pass it to `next`. In this case the interceptor will fire on the inbound
+ *   operations implemented in the new listener.
+ * 3) Make direct inbound calls to the provided listener's methods. This
+ *   short-circuits the interceptor stack.
+ *
+ * Do not modify the listener passed in. Either pass it along unmodified,
+ * ignore it, or call methods on it to short-circuit the call.
+ *
+ * To intercept errors, implement the `onReceiveStatus` method and test for
+ * `status.code !== grpc.status.OK`.
+ *
+ * To intercept trailers, examine `status.metadata` in the `onReceiveStatus`
+ * method.
+ *
+ * This is a trivial implementation of all interceptor methods:
+ * var interceptor = function(options, nextCall) {
+ *   return new InterceptingCall(nextCall(options), {
+ *     start: function(metadata, listener, next) {
+ *       next(metadata, {
+ *         onReceiveMetadata: function (metadata, next) {
+ *           next(metadata);
+ *         },
+ *         onReceiveMessage: function (message, next) {
+ *           next(message);
+ *         },
+ *         onReceiveStatus: function (status, next) {
+ *           next(status);
+ *         },
+ *       });
+ *     },
+ *     sendMessage: function(message, next) {
+ *       next(message);
+ *     },
+ *     halfClose: function(next) {
+ *       next();
+ *     },
+ *     cancel: function(message, next) {
+ *       next();
+ *     }
+ *   });
+ * };
+ *
+ * This is an interceptor with a single method:
+ * var interceptor = function(options, nextCall) {
+ *   return new InterceptingCall(nextCall(options), {
+ *     sendMessage: function(message, next) {
+ *       next(message);
+ *     }
+ *   });
+ * };
+ *
+ * Builders are provided for convenience: StatusBuilder, ListenerBuilder,
+ * and RequesterBuilder
+ *
+ * gRPC client operations use this mapping to interceptor methods:
+ *
+ * grpc.opType.SEND_INITIAL_METADATA -> start
+ * grpc.opType.SEND_MESSAGE -> sendMessage
+ * grpc.opType.SEND_CLOSE_FROM_CLIENT -> halfClose
+ * grpc.opType.RECV_INITIAL_METADATA -> onReceiveMetadata
+ * grpc.opType.RECV_MESSAGE -> onReceiveMessage
+ * grpc.opType.RECV_STATUS_ON_CLIENT -> onReceiveStatus
+ *
+ * @module
+ */
+
+'use strict';
+
+var _ = require('lodash');
+var grpc = require('./grpc_extension');
+var Metadata = require('./metadata');
+var constants = require('./constants');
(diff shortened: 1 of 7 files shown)


murgatroid99:
Can you give a high-level explanation of this implementation? How do the different classes interact? What is the common code path for a call using one or more interceptors?

drobertduke:
At a high level, this PR inserts an InterceptingCall between the JS client and the C-core.

Before:
`client.js => call.cc startBatch`

After:
`client.js => client_interceptors.js => call.cc startBatch`

client_interceptors.js presents the API interceptors must be written to, and the mechanism for executing a chain of interceptors. The interceptors are chained together dynamically, allowing them to be provided either during client construction or during any RPC invocation.

All interception work is completed before submitting a batch to `startBatch`. A unary call with a single FooInterceptor would be broken out like this like this:
Outbound
```
client.js => FooInterceptor.start
          => FooInterceptor.sendMessage
          => FooInterceptor.halfClose    => call.cc startBatch
```
Inbound
```
call.cc startBatch => FooInterceptor.onReceiveMetadata
                   => FooInterceptor.onReceiveMessage
                   => FooInterceptor.onReceiveStatus   => consumer's callback
```



Logic previously in client.js for creating the client_batch and the callback provided to `startBatch` had to be moved to execute after all interceptor method chains have fired. This logic (and the `Client*Stream` logic) has been consolidated/de-duplicated and moved to client_batches.js.

A typical code path with two interceptors (InterceptorFoo and InterceptorBar) would look like this. This is the outbound path for a unary call performing the `start` operation:

**Client.makeUnaryRequest** client.js
**InterceptingCall.start** client_interceptors.js
**InterceptingCall._callNext** client_interceptors.js
**InterceptorFoo.start** interceptor_foo.js
**next** (a closure in InterceptingCall.start) client_interceptors.js
**InterceptingCall.start** client_interceptors.js
**InterceptingCall._callNext** client_interceptors.js
**InterceptorBar.start** interceptor_bar.js
**next** (a closure in InterceptingCall.start) client_interceptors.js
**InterceptingCall.start** client_interceptors.js
**InterceptingCall._callNext** client_interceptors.js
**InterceptingCall.start** (from _getOutboundBatchingInterceptor) client_interceptors.js
**_joinOutboundOperations** client_interceptors.js
**_joinOpsGroup** client_interceptors.js
**_joinOpsCustom** client_interceptors.js

At this point the `start` path has been fully completed on the JS side and the result recorded in the BatchRegistry. We unwind the stack back to **Client.makeUnaryRequest** and repeat the process for `sendMessage`, and a third time for `halfClose`. Once all required outbound operations have been recorded in the BatchRegistry, **_joinOpsCustom** will proceed to deliver the full outbound batch to call.cc's **startBatch** via this path:

**_joinOpsCustom** client_interceptors.js
**_joinOperations** client_interceptors.js
**joiner** (a closure in _getOutboundBatchingInterceptor) client_interceptors.js
**join_outbound** (from _registerUnaryBatch) client_batches.js
**startBatch** call.cc

Inbound operations take a similar path but the interceptor order is reversed, we call the `inbound` join functions, and it ends with delivering RPC errors/results to the client's consumer.

Crevil:
How is progress going on this? Anything I can do to help get this merged?

thelinuxfoundation:

* User @drobertduke isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

drobertduke:
@Crevil Re-writing this to eliminate the batching of operations is going to take some time, and I'll also need to assess the performance impact for our use cases. I hope to have the requested changes addressed in the next two weeks.

Crevil:
Sound great @drobertduke . We have been looking forward to this change and I would just check up if we could assist in getting it forward.

drobertduke:
Update: I hope to have a new version of this PR posted next week

murgatroid99:
In the meantime, it looks like you actually created this PR before we switched to the CNCF CLA, so you'll have to sign that before this can be merged.

thelinuxfoundation:

* User @drobertduke isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

drobertduke:
I've re-written the PR to send operations as they become available instead of trying to match the same batches the client was making originally. We've deployed this new code to production in a limited capacity and do not see a performance penalty from this change vs earlier versions of this PR.

The new interaction pattern with `call.startBatch` leaves some questions about how a singular callback should be used when multiple batches are fired. I'll comment on this in the code.

I've signed the CLA now, thanks for the reminder.

packages/grpc-native-core/src/client_interceptors.js
```diff
@@ -0,0 +1,1343 @@
...
+            metadata._getCoreRepresentation(),
+        };
+        first_listener = listener;
+        call.startBatch(batch, function () {});
+      },
+      sendMessage: function(message) {
+        var batch = {
+          [grpc.opType.SEND_MESSAGE]: serialize(message),
+        };
+        call.startBatch(batch, function () {});
```
drobertduke: How should I reconcile multiple batches (each with their own errors potentially) vs a single callback provided by the consumer? I've skipped the callback in early batches in favor of using it when receiving a status as seen in the [client streaming RPC](https://github.com/grpc/grpc-node/blob/master/packages/grpc-native-core/src/client.js#L645) but will this obscure useful errors from the consumer?

murgatroid99: I believe that what you are doing is correct. I think that every instance of multiple operations in a batch where the single callback actually matters includes a `status` operation. And I'm pretty sure that the status operation is always guaranteed to complete eventually. And I'm also pretty sure that if there is an error with another operation, it is always because of some circumstance that would cause a non-OK status. So in the end, you will want to surface the status.

One thing you need to be careful of here to avoid spurious errors is that you have to start the `SEND_MESSAGE` batch before you start the `SEND_CLOSE_FROM_CLIENT` batch, because the latter indicates that no more messages will be sent and the write will fail if you do them in the wrong order. I haven't seen yet if you are ensuring that ordering elsewhere.

drobertduke: The non-streaming operations are ordered in client.js, here's the unary case: https://github.com/grpc/grpc-node/pull/59/files#diff-86af358fb1304514580c24212de4f182R486

This doesn't guarantee against a badly-writtten interceptor forcing an incorrect order but I assume we don't need to guard against that.

murgatroid99: Do you consider an interceptor that does any asynchronous operations to be badly written? Because I'm pretty sure that an asychronous write interceptor is sufficient to cause those operations to be performed out of order if there is no guard.

murgatroid99: This actually brings up another question for me: if a `sendMessage` interceptor on a streaming request takes an asnychronous action before passing the message to the next layer, is there any protection against messages getting passed down out of order?

murgatroid99: OK, the `sendMessageWithContext` actually answers that question, and I think it also provides a solution to the other problem. I am still worried about `start`, and not just because of unary calls. For any call, sending a message before sending the metadata will not do the right thing, and may crash.

drobertduke: I don't think it's desirable to prevent an interceptor from changing the order of streaming messages. Interceptors can modify messages, swap a message out for a new message, or short circuit the operation entirely to skip a message. Building an ordered list of messages as they are passed in by the client's consumer and then enforcing that order when starting batches would make certain use cases impossible, like circuit breaking on streaming messages.

re: enforcing the order of operations in non-streaming requests, I was expecting interceptor authors to do that work if they want to make a continuation asynchronous. Adding a mechanism to the framework to delay an operation until a required operation is ready could reduce the burden on interceptor authors with async continuations. But it could also lead to difficult-to-debug scenarios with silent failures:

1) An interceptor author writes a bad interceptor which unexpectedly fails to call the SEND_MESSAGE continuation n% of the time.
2) An operation-order-enforcement mechanism in the framework allows the SEND_METADATA operation to be sent but doesn't send CLOSE n% of the time because it is waiting on a SEND_MESSAGE that never comes.
3) n% of all calls pile up without error until they time out or memory runs out.

The same bad interceptor would cause immediate failures with useful errors if the framework is not waiting on an order-enforcement mechanism.

murgatroid99: The use of `sendMessageWithContext` already enforces the order of messages by preventing the top-level stream from sending down each message until the previous message has been fully processed.

And from my point of view, it feels very surprising to users that this interceptor API is clearly written to support asynchronous operations, but if they try to actually use an asynchronous operation, it will screw up some requests unless they do significant extra work to explicitly track their own pending operations. On top of that, in the current library it is not possible for the user to call the API in a way that results in out-of-order calls to the C core. I think that is a valuable guarantee and I would be very hesitant to change it.

drobertduke:
I've added a mechanism to define 'dependencies' for individual operations, check whether those dependencies are met and defer the batch if needed. The use case only applies to RPCs with synchronous requests (unary and serverStreaming).

I'm guessing this implementation is fairly expensive. Some of the lodash calls can be optimized away but I'm leaving them in for now for clarity. I'll work on producing some benchmarks.

drobertduke:
I ran benchmarks on an m4 EC2 instance for three versions of grpc-native:
1) The "base" version (the 044102d commit on master this PR is currently based on)
2) This PR without protections against out-of-order operations (f644c06)
3) This PR with those protections (d83d354)

Each benchmark is the time it takes to issue and get responses to 50k simple unary RPC requests (with no interceptors) to a local server running in another process. I ran the benchmark three times for each version.

**BASE - 044102d**
```
real	0m58.352s
user	0m13.152s
sys	0m0.916s

real	0m58.275s
user	0m13.300s
sys	0m0.784s

real	0m57.998s
user	0m13.360s
sys	0m0.820s
```
**PR without order protection - f644c06**
```
real	0m52.190s
user	0m17.204s
sys	0m1.480s

real	0m51.821s
user	0m16.808s
sys	0m1.596s

real	0m51.974s
user	0m17.164s
sys	0m1.528s
```
**PR with order protection - d83d354**
```
real	1m6.632s
user	0m19.756s
sys	0m3.140s

real	1m6.389s
user	0m20.468s
sys	0m2.768s

real	1m6.010s
user	0m19.656s
sys	0m3.180s
```
Looking at the median timing for each version we get these results (lower is better):

**`real` median benchmarks**
PR without order protection: 89.2% of base
PR with order protection: 113.9% of base

I was surprised by the performance improvement of the PR without order protection over the base. Looking at the `user` timings we can see that more work is being done by the PR-versions of the client as I would expect. It looks like breaking up the request operations is significantly increasing efficiency, at least in this benchmark scenario.

**`user` median benchmarks**
PR without order protection: 132.6% of base
PR with order protection: 148.5% of base

I won't have time to work on performance improvements (reducing `sort()` and lodash calls) for about a week. If you want to merge this one as-is I can post a follow up PR later with those improvements.

packages/grpc-native-core/src/common.js
```diff
@@ -75,6 +77,73 @@ exports.defaultGrpcOptions = {
...
+  }
+};
+
+/**
+ * Get a call object built with the provided options.
+ * @param {grpc.Channel} channel
+ * @param {string} path
+ * @param {grpc.Client~CallOptions=} options Options object.
+ */
+exports.getCall = function(channel, path, options) {
```
murgatroid99: It looks like this function is only used in `client_interceptors.js`, so it should be defined there.

murgatroid99:
OK, I can't say I've totally wrapped my head around all of the code here, but at this point I can't find any other problems, so I'm going to merge it. Thank you for your contribution, and your patience in working through the contribution process. Assuming nothing goes wrong, this will be published in the 1.11 release.

ebenoist:
Is there a timeline for when the 1.11 release is expected? Happy to test some of this out in our organization as well if that helps.

PR #62, opened 2017-10-12
Add requiring a fixture to specify implementation to test

kjin:
This change makes it so that running an api test requires one of four possible fixture files to be pre-required (either through the command line, or programmatically using gulp). Fixture files are named with a `fixtures/${server}_${client}.js` naming convention. Currently, `fixtures/js_*.js` files won't be used in gulp because the pure JS server implementation isn't there.

I moved the `api/interop_sanity_test.js` to `interop/` to make glob-matching easier -- since the set of required files is fundamentally different between it and other API tests.

Also, I enabled the full set of linking steps in the top level `gulpfile` -- I didn't observe the behavior seen on npm/npm#18835. Hopefully this is reproducible in other environments...

### DIFF
--- a/test/api/surface_test.js
+++ b/test/api/surface_test.js
@@ -485,7 +485,7 @@ describe('Echo metadata', function() {
     call.end();
   });
   it('shows the correct user-agent string', function(done) {
-    var version = require('../any_grpc')['$implementationInfo'].server.corePjson.version;
+    var version = require('../any_grpc')['$implementationInfo'].client.corePjson.version;
     var call = client.unary({}, metadata,
                             function(err, data) { assert.ifError(err); });
     call.on('metadata', function(metadata) {
--- a/test/api/surface_test.js
+++ b/test/api/surface_test.js
@@ -21,7 +21,7 @@
 var assert = require('assert');
 var _ = require('lodash');
 
-var grpc = require('grpc');
+var grpc = require('../any_grpc').all;
 
 var MathClient = grpc.load(
     __dirname + '/../../packages/grpc-native-core/deps/grpc/src/proto/math/math.proto').math.Math;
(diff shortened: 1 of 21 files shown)


murgatroid99:
I don't think you should move `interop_sanity_test.js` to the `interop` directory. Instead, the other test files should also support the `js_native` and `native_js` fixtures. I know that this will require possibly significant modifications to the test files themselves, but it's probably the best way to get significant API-level test coverage of the pure JS client in the near future, before we have a pure JS server too.

test/gulpfile.js
```diff
@@ -29,25 +29,36 @@ const gulp = help(_gulp);
...
+        reporter: 'mocha-jenkins-reporter',
+        require: `${testDir}/fixtures/${fixture}.js`
+      }))
+      .resume() // put the stream in flowing mode
+      .on('end', resolve)
+      .on('error', reject);
+  });
+  const apiTestGlob = `${apiTestDir}/*.js`;
+  const runTestsArgPairs = [
+    [apiTestGlob, 'native_native'],
```
murgatroid99: This scheme for specifying fixtures seems a bit backwards to me. Instead of writing the pair as one string and splitting it up, why not just write it like `{client: 'native', server: 'native'}` and join them if you have to?

test/api/surface_test.js
```diff
@@ -485,7 +485,7 @@ describe('Echo metadata', function() {
     call.end();
   });
   it('shows the correct user-agent string', function(done) {
-    var version = require('grpc/package.json').version;
+    var version = require('../any_grpc')['$implementationInfo'].server.corePjson.version;
```
murgatroid99: I'm pretty sure the user agent should match the client version, not the server version.

PR #67, opened 2017-10-20
Add artifact build scripts for native core

murgatroid99:
This copies the Node artifact build scripts pretty much verbatim to this repo, except that it adds 1.7.0 to the list of electron versions to build for.


murgatroid99:
Yes, it's copied from there.

PR #72, opened 2017-10-31
Include index.d.ts in the published npm package

ThomWright:
I tried using `grpc@1.7.0-pre1` expecting to use the types written for #52, but `index.d.ts` isn't published.

I also did another couple of minor corrections in `package.json`.

### DIFF
--- a/packages/grpc-native-core/package.json
+++ b/packages/grpc-native-core/package.json
@@ -16,7 +16,7 @@
     }
   ],
   "directories": {
-    "lib": "src/node/src"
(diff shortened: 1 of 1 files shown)


ThomWright:
I signed it!

murgatroid99:
Can you please also make the corresponding changes to `packages/grpc-native-core/templates/package.json.template`?

murgatroid99:
Also, this change should be made starting with the versions of those files in the v1.7.x branch, and the PR should be based on that branch.

PR #75, opened 2017-11-01
Fix channel state code and add backoff code

murgatroid99:
This streamlines some of the channel state transition code and adds proper channel reconnection support with exponential backoff in the pure JS implementation, in accordance with the [Connectivity Semantics and API](https://github.com/grpc/grpc/blob/master/doc/connectivity-semantics-and-api.md) and [Connection Backoff Protocol](https://github.com/grpc/grpc/blob/master/doc/connection-backoff.md) documents.

I removed the idle timeout code because it wasn't correct, and according to grpc/grpc#13212, the C core doesn't have that either.

### DIFF
--- a/packages/grpc-js-core/src/channel.ts
+++ b/packages/grpc-js-core/src/channel.ts
@@ -15,6 +15,12 @@ import {Metadata, MetadataObject} from './metadata';
 
 const IDLE_TIMEOUT_MS = 300000;
 
+const MIN_CONNECT_TIMEOUT_MS = 20000;
+const INITIAL_BACKOFF_MS = 1000;
+const BACKOFF_MULTIPLIER = 1.6;
+const MAX_BACKOFF_MS = 120000;
+const BACKOFF_JITTER = 0.2;
+
 const {
   HTTP2_HEADER_AUTHORITY,
   HTTP2_HEADER_CONTENT_TYPE,
@@ -35,6 +41,10 @@ export enum ConnectivityState {
   TRANSIENT_FAILURE,
   IDLE,
   SHUTDOWN
+}
+
+function uniformRandom(min:number, max: number) {
+  return Math.random() * (max - min) + min;
 }
 
 // todo: maybe we want an interface for load balancing, but no implementation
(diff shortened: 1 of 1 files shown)


packages/grpc-js-core/src/channel.ts
```diff
@@ -62,49 +72,97 @@ export interface Channel extends EventEmitter {
...
+        this.currentBackoffDeadline = new Date(now.getTime() + this.currentBackoff + uniformRandom(-jitterMagnitude, jitterMagnitude));
+      }
+      this.startConnecting();
+      break;
+    case ConnectivityState.READY:
+      this.emit('connect');
+      break;
+    case ConnectivityState.TRANSIENT_FAILURE:
+      this.subChannel = null;
+      this.backoffTimerId = setTimeout(() => {
```
kjin: `this.backoffTimerId` is never used. Should it be used to cancel the timeout on shutdown?

packages/grpc-js-core/src/channel.ts
```diff
@@ -62,49 +72,97 @@ export interface Channel extends EventEmitter {
...
+        this.transitionToState([ConnectivityState.TRANSIENT_FAILURE], ConnectivityState.CONNECTING);
+      }, this.currentBackoffDeadline.getTime() - now.getTime());
+      break;
+    case ConnectivityState.IDLE:
+    case ConnectivityState.SHUTDOWN:
+      if (this.subChannel !== null) {
+        this.subChannel.shutdown({graceful: true});
```
kjin: Looks like this is duplicated with https://github.com/grpc/grpc-node/pull/75/files#diff-6270a066e64c77ba494d58b714011deeR256 -- should the shutdown be removed from there?

packages/grpc-js-core/src/channel.ts
murgatroid99: yes, I've added that.

packages/grpc-js-core/src/channel.ts
murgatroid99: That's right. I removed the other instance.

PR #83, opened 2017-11-08
Fixed name of ServerCredentials.createSsl in native docs & types.

jSherz:
The docs and typings had a duplicate method on ServerCredentials, both called `createInsecure`. I've updated them to have the correct name (createSsl).


jSherz:
Hi @murgatroid99,

Thank you for getting to this so quickly!

Is there anything else I need to do for this to be merged?

Thanks,
James

PR #84, opened 2017-11-08
Generic Ts definitions

andrew8er:
I added some generic type parameters to the index.d.ts, in particular:
```ts
loadObject<T = GrpcObject>(…): T;
load<T = GrpcObject>(…): T;
```
can now be parametrised to an autogenerated or handwritten specific type. Also in
```ts
class Server {
   // ...
   addService<T>(service: Service<T>, implementation: T): void;
   // ...
}
```
the implementation can now be typechecked against a specific type.

This is most probably a breaking change for Ts projects. When supplying default values for all generic parameters, it would be non breaking, but that would defeat the point somehow.


andrew8er:
I signed it!

packages/grpc-native-core/index.d.ts
```diff
@@ -9,7 +9,7 @@ declare module "grpc" {
    * @param options Options to apply to the loaded file
    * @return The resulting gRPC object.
    */
-  export function loadObject(value: object, options?: LoadObjectOptions): GrpcObject;
+  export function loadObject<T = GrpcObject>(value: object, options?: LoadObjectOptions): T;
```
murgatroid99: I don't understand this change. The output of this function can only be `GrpcObject` as defined below. What does parameterizing it accomplish?

packages/grpc-native-core/index.d.ts
```diff
@@ -76,7 +76,7 @@ declare module "grpc" {
    * @param options Options to apply to the loaded file
    * @return The resulting gRPC object
    */
-  export function load(filename: Filename, format?: "proto" | "json", options?: LoadOptions): GrpcObject;
+  export function load<T = GrpcObject>(filename: Filename, format?: "proto" | "json", options?: LoadOptions): T;
```
murgatroid99: Similar to the issue with `loadObject`, I don't understand what this accomplishes.

packages/grpc-native-core/index.d.ts
```diff
@@ -132,6 +132,8 @@ declare module "grpc" {
    */
   export function setLogVerbosity(verbosity: logVerbosity): void;
 
+  export type Service<T> = ProtobufService;
```
murgatroid99: What is the purpose of a generic type that doesn't use the type parameter in its definition?

packages/grpc-native-core/index.d.ts
```diff
@@ -193,7 +195,7 @@ declare module "grpc" {
      * @param implementation Map of method names to method implementation
      * for the provided service.
      */
-    addService(service: Service, implementation: { [name: string]: handleCall }): void;
+    addService<T>(service: Service<T>, implementation: T): void;
```
murgatroid99: This seems to make the function signature less useful by removing the specific definition of the type of `implementation`, and I don't see any clear benefit. Because the `Service<T>` definition doesn't even use the type parameter `T`, there is no useful information about how the types of the two arguments should be related.

Also, the original file is actually incorrect here. The type of the `service` argument should be `ServiceDefinition`.

packages/grpc-native-core/index.d.ts
```diff
@@ -202,7 +204,7 @@ declare module "grpc" {
      * @param implementation Map of method names to method implementation
      * for the provided service.
      */
-    addProtoService(service: Service | ServiceDefinition, implementation: { [name: string]: handleCall }): void;
+    addProtoService<T>(service: Service<T> | ServiceDefinition, implementation: T): void;
```
murgatroid99: I have the same concerns here as with `addService` regarding the usefulness of the type parameterization.

packages/grpc-native-core/index.d.ts
```diff
@@ -258,17 +260,21 @@ declare module "grpc" {
...
+    handleUnaryCall<any> |
+    handleClientStreamingCall<any> |
+    handleServerStreamingCall<any> |
+    handleBidiStreamingCall<any, any>;
 
   /**
    * User-provided method to handle unary requests on a server
    */
-  type handleUnaryCall = (call: ServerUnaryCall, callback: sendUnaryData) => void;
+  type handleUnaryCall<T> = (call: ServerUnaryCall<T>, callback: sendUnaryData<T>) => void;
```
murgatroid99: Later in the file, `ServerUnaryCall` is defined in terms of the request type, and `sendUnaryData` is defined in terms of the response type, but you are using just one type parameter here for both. This is incorrect.

packages/grpc-native-core/index.d.ts
```diff
@@ -302,13 +308,13 @@ declare module "grpc" {
   /**
    * User provided method to handle client streaming methods on the server.
    */
-  type handleClientStreamingCall = (call: ServerReadableStream, callback: sendUnaryData) => void;
+  type handleClientStreamingCall<T> = (call: ServerReadableStream<T>, callback: sendUnaryData<T>) => void;
```
murgatroid99: As with `handleUnaryCall`, this is conflating the request and response types.

packages/grpc-native-core/index.d.ts
```diff
@@ -302,13 +308,13 @@ declare module "grpc" {
...
 
   /**
    * A stream that the server can read from. Used for calls that are streaming
    * from the client side.
    */
-  export class ServerReadableStream extends Readable {
+  export class ServerReadableStream<T> extends Readable {
```
murgatroid99: I assume that `T` is supposed to be the request type here, but because it's not used, it's not entirely clear.

packages/grpc-native-core/index.d.ts
```diff
@@ -337,13 +343,13 @@ declare module "grpc" {
...
    */
-  type handleServerStreamingCall = (call: ServerWriteableStream) => void;
+  type handleServerStreamingCall<T> = (call: ServerWriteableStream<T>) => void;
 
   /**
    * A stream that the server can write to. Used for calls that are streaming
    * from the server side.
    */
-  export class ServerWriteableStream extends Writable {
+  export class ServerWriteableStream<T> extends Writable {
```
murgatroid99: The class `ServerWritableStream` is also responsible for sending responses, so I think it's a bit misleading to include only the request type in the type definition.

packages/grpc-native-core/index.d.ts
```diff
@@ -941,15 +947,15 @@ declare module "grpc" {
      * @param callback The callback to for when the response is received
      * @return An event emitter for stream related events
      */
-    makeUnaryRequest(
+    makeUnaryRequest<T>(
```
murgatroid99: This again conflates the request and response types. The `argument` parameter should have the request type, and the `callback` parameter should have the response type. And it's not clear which of those is supposed to be the type parameter for `ClientUnaryCall`.

Also, if you're going to add type parameters to this function, it would make sense to do the same with `serialize` and `deserialize`.

packages/grpc-native-core/index.d.ts
```diff
@@ -962,14 +968,14 @@ declare module "grpc" {
      * @param callback The callback to for when the response is received
      * @return An event emitter for stream related events
      */
-    makeClientStreamRequest(
+    makeClientStreamRequest<W>(
```
murgatroid99: The same concerns about conflating the request and response types apply here too.

packages/grpc-native-core/index.d.ts
```diff
@@ -983,14 +989,14 @@ declare module "grpc" {
      * @param options Options map
      * @return An event emitter for stream related events
      */
-    makeServerStreamRequest(
+    makeServerStreamRequest<R>(
```
murgatroid99: The same concerns about conflating the request and response types apply here too.

packages/grpc-native-core/index.d.ts
```diff
@@ -417,7 +423,7 @@ declare module "grpc" {
    * Callback function passed to server handlers that handle methods with
    * unary responses.
    */
-  type sendUnaryData = (error: ServiceError | null, value: any, trailer?: Metadata, flags?: number) => void;
+  type sendUnaryData<T> = (error: ServiceError | null, value: T, trailer?: Metadata, flags?: number) => void;
```
murgatroid99: The `value` argument can be `null` or `undefined` if the `error` argument is non-null.

packages/grpc-native-core/index.d.ts
```diff
@@ -1189,7 +1199,7 @@ declare module "grpc" {
    * @param error The error, if the call failed
    * @param value The response value, if the call succeeded
    */
-  export type requestCallback = (error: ServiceError | null, value: any) => void;
+  export type requestCallback<T> = (error: ServiceError | null, value: T) => void;
```
murgatroid99: The `value` argument can be `null` or `undefined` if the `error` argument is non-null.

packages/grpc-native-core/index.d.ts
andrew8er: This way, you can give the return object structure a type. For the "routeguide" example that would be something like this:

```ts
import { Message } from 'protobufjs';
// message types in package "routeguide" defined here:
import * as routeguide from './routeguide';


// a modified copy of Message
interface MessageBase<T> {
	new(values?: Partial<T>, var_args?: string[]): T & Message;
	decode(buffer: ArrayBuffer | ByteBuffer | Buffer | string, length?: number | string, enc?: string): T & Message;
	decodeDelimited(buffer: ByteBuffer | ArrayBuffer | Buffer | string, enc?: string): T & Message;
	decode64(str: string): T & Message;
	decodeHex(str: string): T & Message;
	decodeJSON(str: string): T & Message;
}

// use this as T
export default interface Description {
	routeguide: {
		// types
		Point: MessageBase<routeguide.Point>;
		Rectangle: MessageBase<routeguide.Rectangle>;
		Feature: MessageBase<routeguide.Feature>;
		RouteNote: MessageBase<routeguide.RouteNote>;
		RouteSummary: MessageBase<routeguide.RouteSummary>;
		// services
		RouteGuide: typeof routeguide.RouteGuide.Client;
	}
}
```

This will give you a type checked object you can use later on.

packages/grpc-native-core/index.d.ts
andrew8er: This gives the Ts compiler the ability to infer the type in `Server.addService()`:

```ts
addService<T>(service: Service<T>, implementation: T): void;
```
Since you give a `Service<T>` object as the first argument, it follows that the implementation in the second argument must satisfy `T`. Thus you usually do not need to add the parameter at all. For example:

```ts
// definitions
type RouteGuide = typeof routeguide.Client;

namespace routeguide {
	interface Service {
		getFeature(call: ServerUnaryCall<Point>, callback: sendUnaryData<Feature>): void;
		listFeatures(call: ServerWriteableStream<Rectangle>): void;
		recordRoute(call: ServerReadableStream<Point>, callback: sendUnaryData<RouteSummary>): void;
		routeChat(call: ServerDuplexStream<RouteNote, RouteNote>): void;
	}

	declare class Client extends GrpcClient {
		static service: GrpcService<Service>;

		getFeature(arg: Point, callback: sendUnaryData<Feature>): void;
		listFeatures(arg: Rectangle): ClientReadableStream<Feature>;
		recordRoute(callback: sendUnaryData<RouteSummary>): ClientWritableStream<Point>;
		routeChat(): ClientDuplexStream<RouteNote, RouteNote>;
	}
}

// usage

server.addService(RouteGuide.service, {
	getFeature(call: ServerUnaryCall<Point>, callback: sendUnaryData<Feature>): void {
		// omitted
	},
	listFeatures(call: ServerWriteableStream<Rectangle>): void {
		// omitted
	},
	recordRoute(call: ServerReadableStream<Point>, callback: sendUnaryData<RouteSummary>): void {
		// omitted
	},
	routeChat(call: ServerDuplexStream<RouteNote, RouteNote>): void {
		// omitted
	}
});
```

packages/grpc-native-core/index.d.ts
andrew8er: What do you mean by the "specific definition"? The purpose of this change was to make the required type of implementation as specific as possible.

`{ [name: string]: handleCall }` basically means an object with any handlers, named in any way, so to me it is quite unspecific.

packages/grpc-native-core/index.d.ts
andrew8er: See above

packages/grpc-native-core/index.d.ts
andrew8er: Ok, I did not dig deep enough into the API, yes, so this should be revised.

packages/grpc-native-core/index.d.ts
andrew8er: The hope was that one could later introduce type specific stream class definitions that refine the existing ones from node, without having another API breakage.

packages/grpc-native-core/index.d.ts
andrew8er: So that should be extended to `T | null`.

packages/grpc-native-core/index.d.ts
andrew8er: See below.

packages/grpc-native-core/index.d.ts
murgatroid99: It looks like what you're saying here is that the user has to both define `T` and explicitly state that another object has the type `Service<T>`. I don't see how that improves type safety or provides any benefit whatsoever.

packages/grpc-native-core/index.d.ts
murgatroid99: The old definition says that the `implementation` type has to be an object that maps string names to one of four call handler function types. The new definition just says that the argument is `T`, which provides no information at all.

packages/grpc-native-core/index.d.ts
andrew8er: If you have a `T` defined (preferably autogenerated from the proto files) like in the example above, then it is an improvement, since the compiler will warn you about any missing/wrong function declarations. Also you do not need to create a `Service<T>` type, you only need 
to define the server side API (`T`) like in `routeguide.Service` and `GrpcService<Service>` will give you the correct type.

We could give `T` a default value, so users do not have to provide a type definition like so:
```
type ServiceImplementation = { [name: string]: handleCall };

export class Server {
    // …
    addService<T = ServiceImplementation>(service: Service<T>, implementation: T): void;
}
```
But having a generic implementation definition actually means that you cannot benefit from Typescripts type checking.

LouiseBC:
Thanks for this, is there any chance this can be merged soon?

thelinuxfoundation:

* User @andrew8er isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

andrew8er:
(Please ignore this comment, it is only for the CLA GitHub bot)

I signed the CLA…

packages/grpc-native-core/index.d.ts
```diff
@@ -216,18 +228,23 @@ declare module "grpc" {
...
+  export type UntypedServiceImplementation = { [name: string]: handleCall<any, any> };
+
   /**
    * An object that completely defines a service.
    * @typedef {Object.<string, grpc~MethodDefinition>} grpc~ServiceDefinition
    */
-  export interface ServiceDefinition {
-    [s: string]: MethodDefinition;
+  export interface ServiceDefinition<ImplementationType> {
+    [s: string]: MethodDefinition<any, any>;
```
murgatroid99: Is it possible to use the type information in `ImplementationType` to further specify these `MethodDefinition` types?

packages/grpc-native-core/index.d.ts
murgatroid99: For consistency with the other call types, this should probably be parameterized on `RequestType`, even if you don't use the type in the definition.

packages/grpc-native-core/index.d.ts
andrew8er: I tightened the definition of `ServiceDefinition`: the type can only contain keys from the `ImplementationType`. But I do not see how to apply any further typing to the `MethodDefinition`. But I think it is not too bad, `MethodDefinition` is used for run time reflection, and user code is checked against `ServiceDefinition`.

LouiseBC:
Why are the checks not completing? I'd love if this could be a part of the next release :)

PR #86, opened 2017-11-09
Update submodule to head, and update templates to match

murgatroid99:


murgatroid99:
That Mac failure is grpc/grpc#12932. Rerunning

murgatroid99:
@nicolasnoble I think it might make sense to force merge this. The tests that grpc/grpc runs against this repo are currently always failing, so this should help with that.

PR #87, opened 2017-11-10
Only run watchConnectivityState tests, with core API tracing

murgatroid99:
DO NOT MERGE

This PR exists only to trigger tests


murgatroid99:
This test has now been fixed due to grpc/grpc#13433 and #95.

PR #88, opened 2017-11-13
Failing metadata plugin results in UNAVAILABLE

jtattermusch:
Context: https://github.com/grpc/grpc/pull/13322

Needs to be merged before https://github.com/grpc/grpc/pull/13363. The change is a little awkward, but since grpc-node now lives in a different repository, we don't have merge atomicity anymore and we can't really come up anything better without breaking the test temporarily.

### DIFF
--- a/test/api/credentials_test.js
+++ b/test/api/credentials_test.js
@@ -306,7 +306,7 @@ describe('client credentials', function() {
       done();
     });
   });
-  it('should propagate errors that the updater emits', function(done) {
+  it('should fail the call if the updater emits', function(done) {
     var metadataUpdater = function(service_url, callback) {
       var error = new Error('Authentication error');
       error.code = grpc.status.UNAUTHENTICATED;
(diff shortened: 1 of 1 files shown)


jtattermusch:
CC @jboeuf

murgatroid99:
We're just going to have to accept that that test is going to break sometimes. I don't think we need to make this kind of transitional change. We should just make the actual change with the next submodule update.

jtattermusch:
I'm fine with breaking the tests for duration of ~minutes (basically having all the PRs ready to merge), but knowingly introducing a breakage for longer time is wrong. What is the right ordering of PRs to minimize the impact?
Is this correct:
1. merge this PR (after modifying it to only accept UNAVAILABLE status) - that will break the grpc/grpc tests temporarily
2. merge grpc/grpc#13363 shortly after 1) which will fix the tests in both grpc/grpc and grpc-node.

murgatroid99:
I have opened #90, which is my preferred solution. You should merge your PR on grpc/grpc, then I will update that one and merge it. The grpc/grpc tests will be broken temporarily.

nicolasnoble:
I would side with @murgatroid99 here that we shouldn't try making commits that handles two things at once. Unless the change is too big, preparing the change such as #90 and committing grpc/grpc#13363 in almost lockstep is the better way to handle this.

jtattermusch:
Closing in favor of #90.

PR #90, opened 2017-11-14
Fix a credentials plugin test to match changes in core

murgatroid99:
This currently points to the PR that makes the core change. It will need to be modified to point to grpc/grpc master instead before it is merged.


jtattermusch:
It actually looks like that if we merge this PR first, not tests will be broken (as this accepts both UNAVAILABLE and UNAUTHENTICATED) - so we can merge this first and then merge https://github.com/grpc/grpc/pull/13363?

jtattermusch:
@murgatroid99, I checked the test results for https://github.com/grpc/grpc/pull/13363 and the UNAUTHENTICATED -> UNAVAILABLE change is the only breakage there. So it seems it's ready to merge after this PR is merged.

murgatroid99:
As I said, this submodule update points to your PR instead of to master. I don't want to merge in that state. This shows that the tests will pass, so as soon as you merge your PR, I can update the submodule reference in this one and merge it.

jtattermusch:
Ah, ok, get it now. Feel free to proceed. Thanks!

jtattermusch:
@murgatroid99  any progress here? we've had this PR and https://github.com/grpc/grpc/pull/13363 around for a while now.

murgatroid99:
As I said before, I am waiting for you to merge your PR before I merge this one.

jtattermusch:
Thanks!

PR #105, opened 2017-11-29
Add Alpine build script and Dockerfile, improve existing build scripts

murgatroid99:
This removes the need to run individual scripts multiple times with pre-determined arguments. Now a single script run per operating system is sufficient to build all of the binaries.


packages/grpc-native-core/tools/docker/alpine_artifact/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM node:8-alpine
+RUN apk add --no-cache python curl bash build-base
```
nicolasnoble: My OCD is a bit piqued that you're missing a new line here ;-)

packages/grpc-native-core/tools/run_tests/artifacts/build_all_linux_artifacts.sh
```diff
@@ -0,0 +1,38 @@
...
+
+export ARTIFACTS_OUT=$base_dir/artifacts
+
+rm -rf build || true
+
+mkdir -p "${ARTIFACTS_OUT}"
+
+docker build -t alpine_node_artifact $base_dir/tools/docker/alpine_artifact
+
+$tool_dir/build_artifact_node.sh false
```
nicolasnoble: Suggest something a bit more explicit here than just "false": maybe `alpine`/`no_alpine`, which then gets translated into the proper command line arguments.

Maybe something a tiny bit fancier would be `--with_alpine`, and then processing the command line using the typical paradigm:

```
NODE_ALPINE_BUILD=false

while true ; do
  case $1 in
  --with-alpine)
    NODE_ALPINE_BUILD=true
    ;;
  *)
    echo "Unknown parameter: $1"
    exit 1
    ;;
  esac
  shift || break
done
```

PR #106, opened 2017-11-30
Add more helpful error message for failure to load extension

murgatroid99:
We've gotten multiple reports where the extension failed to load, and this will help us diagnose the problem in more cases.

### DIFF
--- a/packages/grpc-native-core/src/grpc_extension.js
+++ b/packages/grpc-native-core/src/grpc_extension.js
@@ -27,6 +27,26 @@ var binary = require('node-pre-gyp/lib/pre-binding');
 var path = require('path');
 var binding_path =
     binary.find(path.resolve(path.join(__dirname, '../package.json')));
-var binding = require(binding_path);
+var binding;
+try {
+  binding = require(binding_path);
+} catch (e) {
+  var fs = require('fs');
+  var searchPath = path.dirname(path.dirname(binding_path));
+  var searchName = path.basename(path.dirname(binding_path));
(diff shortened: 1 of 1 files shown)


packages/grpc-native-core/src/grpc_extension.js
```diff
@@ -27,6 +27,26 @@ var binary = require('node-pre-gyp/lib/pre-binding');
...
+try {
+  binding = require(binding_path);
+} catch (e) {
+  var fs = require('fs');
+  var searchPath = path.dirname(path.dirname(binding_path));
+  var searchName = path.basename(path.dirname(binding_path));
+  var foundNames = fs.readdirSync(searchPath);
+  if (foundNames.indexOf(searchName) === -1) {
+    var message = `Failed to load gRPC binary module because it was not installed for the current system
+Expected: ${searchName}
```
nicolasnoble: Didn't you mean `binding_path` here instead of `searchName` ?

murgatroid99: No. `searchName` is the name of the directory that contains `grpc_node.node`. And that directory name contains the relevant information for diagnosing this problem. Plus, it can be directly (visually) compared with the contents of `foundNames`.

nicolasnoble: Okay, then maybe rephrase it a tiny bit, because the words "gRPC binary module" to me means `grpc_node.node`. Maybe you need to add the word "location" next to "Expected" ?

PR #118, opened 2017-12-07
Revert "Copy API tests back to grpc-native-core"

murgatroid99:
Reverts grpc/grpc-node#116. It broke the tests when run from the grpc/grpc repo.


kjin:
How do you run them from the `grpc/grpc` repo? Some of the paths did rely on the `grpc/grpc` submodule being in `../deps/grpc`.

murgatroid99:
The script for doing so is https://github.com/grpc/grpc/blob/master/tools/run_tests/helper_scripts/run_grpc-node.sh. The purpose is for testing on Kokoro, and it actually just clones this repo, then checks out the relevant commit of grpc/grpc for the submodule.

PR #126, opened 2017-12-15
Add error code name to status error messages

murgatroid99:
We frequently get questions that reference status errors, but they almost never realize that the `code` is important or know what it means. This adds the status code number and name to the error message.

### DIFF
--- a/packages/grpc-native-core/src/client.js
+++ b/packages/grpc-native-core/src/client.js
@@ -61,10 +61,11 @@ var version = require('../package.json').version;
  */
 function createStatusError(status) {
   let statusName = _.invert(constants.status)[status.code];
-  let message = `${status.code} ${status.name}: ${status.details}`;
+  let message = `${status.code} ${statusName}: ${status.details}`;
   let error = new Error(message);
   error.code = status.code;
   error.metadata = status.metadata;
+  error.details = status.details;
   return error;
 }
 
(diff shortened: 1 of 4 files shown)


packages/grpc-native-core/src/client.js
```diff
@@ -53,6 +53,21 @@ var Duplex = stream.Duplex;
...
+ * @private
+ * @param {grpc~StatusObject} status The status object
+ * @return {Error} The resulting Error
+ */
+function createStatusError(status) {
+  let statusName = _.invert(constants.status)[status.code];
+  let message = `${status.code} ${status.name}: ${status.details}`;
+  let error = new Error(message);
+  error.code = status.code;
+  error.details = status.details;
```
kjin: Should we include `error.metadata` as well?

murgatroid99: Yes. That is actually what I meant to write there. The details string is in the message. It probably doesn't need to be in a separate field too.

kjin:
LGTM once tests are amended to reflect new message changes

murgatroid99:
I was lazy and didn't run the tests, so I didn't realized that I had so many tests that depended on the specific text of the error messages. I'll have to think about how to handle that.

murgatroid99:
My solution is to add a details string to the error objects that exactly matches the original details string, and check that in the tests.

PR #132, opened 2017-12-20
grpc-js-core: prevent callback before 'end' event and handle eos headers as trailers

kjin:
#### `grpc-js-core: in unary calls, don't invoke callback before end event`

For unary response calls, `'status'` might be emitted too soon, and the user won't see any data in their callback. This change ensures that if `'status` comes before `'end'`, that `'end'` will invoke the user callback instead.

#### `grpc-js-core: handle eos headers as trailers`

We should treat headers with `END_STREAM` flags as if they were trailers. In addition, I've added code to prevent race conditions for emitted StatusObject instances: if a header or trailer (in that order of priority) is still being put through the filter stack, wait to give those execution threads an opportunity to end the call first, before allowing the call to be ended synchronously in a `streamClosed` event.

### DIFF
--- a/packages/grpc-js-core/src/call-stream.ts
+++ b/packages/grpc-js-core/src/call-stream.ts
@@ -105,6 +105,13 @@ export class Http2CallStream extends Duplex implements CallStream {
   // Status code mapped from :status. To be used if grpc-status is not received
   private mappedStatusCode: Status = Status.UNKNOWN;
 
+  // Promise objects that are re-assigned to resolving promises when headers
+  // or trailers received. Processing headers/trailers is asynchronous, so we
+  // can use these objects to await their completion. This helps us establish
+  // order of precedence when obtaining the status of the call.
+  private handlingHeaders = Promise.resolve();
+  private handlingTrailers = Promise.resolve();
+
   // This is populated (non-null) if and only if the call has ended
   private finalStatus: StatusObject|null = null;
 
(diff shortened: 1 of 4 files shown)


kjin:
Just to summarize what we talked about earlier in person -- there are three new `await *` calls:

1. Just before possibly emitting status from trailers, wait for headers to finish processing (if they are being processed)
2. Just before possibly emitting status from `'close'` event, wait for trailers to finish processing (if they are being processed)
3. Just before possibly emitting status from `cancelWithStatus` call, wait for trailers to finish processing (if they are being processed)

We decided that 1 & 2 are necessary. For 3 I believe this is also necessary because we don't want trailing metadata to be ignored in a race condition where `cancelWithStatus` could be called before the async processing of trailing metadata is finished. In practice this manifests in a test like this: 
https://github.com/grpc/grpc-node/blob/master/test/api/surface_test.js#L905-L913
and also other tests that currently check the exact error type.

FYI, I made two more changes:
- dropped the change (that looked like [this](https://github.com/kjin/grpc-node/commit/25880254135df64084cec6e3c7280ad65786c2e7)) to call the unary response callback in a call's `'end'` handler if it occurs after `'status'`, per our earlier discussion. It turns out it's not even needed to pass tests.
- changed `'streamClosed'` to `'close'`, as the event name [changed in Node 8.9.3/9.3.0](https://github.com/nodejs/node/pull/17328)

PR #134, opened 2017-12-20
grpc-js-core: merge all eventemitter overloads

kjin:
This is a non-logical change (there are slight re-orderings of statements that should have no visible effect) to reduce the boilerplate `EventEmitter` type overloads.

### DIFF
--- a/packages/grpc-js-core/src/client.ts
+++ b/packages/grpc-js-core/src/client.ts
@@ -152,9 +152,7 @@ export class Client {
     call.write(writeObj);
     call.end();
     this.handleUnaryResponse<ResponseType>(call, deserialize, callback);
-
-    const emitter: ClientUnaryCall = new ClientUnaryCallImpl(call);
-    return emitter;
+    return new ClientUnaryCallImpl(call);
   }
 
   makeClientStreamRequest<RequestType, ResponseType>(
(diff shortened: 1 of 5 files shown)


packages/grpc-js-core/src/client.ts
```diff
@@ -147,13 +146,14 @@ export class Client {
...
         this.channel.createStream(method, metadata, options);
-    const emitter: ClientUnaryCall = new ClientUnaryCallImpl(call);
     const message: Buffer = serialize(argument);
     const writeObj: WriteObject = {message: message};
     writeObj.flags = options.flags;
     call.write(writeObj);
     call.end();
     this.handleUnaryResponse<ResponseType>(call, deserialize, callback);
+
+    const emitter: ClientUnaryCall = new ClientUnaryCallImpl(call);
```
murgatroid99: Why move this (and the others in the file) to later in the function?

kjin: The `ClientUnaryCall` is made only to be passed to the client, and doesn't have to be a local variable at all (we can just do `return new ClientUnaryCallImpl(call);` instead). I guess this was a conservative step towards that approach. Do you think it's more readable with all of the object allocations together, or avoiding creating the `emitter` local variable entirely?

murgatroid99: For some reason I was thinking that you would want the `emitter` to exist before taking actions on the call, but everything is asynchronous anyway so that doesn't really matter. I think it makes sense to just remove the `emitter` variable.

thelinuxfoundation:

* User @kjin isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

kjin:
I've signed the CLA.

Issue #142, opened 2018-01-05
Bug when calling server.tryShutdown


franciscocpg:
**OS Version**
Linux 4.14.11-1-ARCH #1 SMP PREEMPT Wed Jan 3 07:02:42 UTC 2018 x86_64 GNU/Linux

**Node**
v9.3.0

**grpc-node**
1.8.0

When calling `server.tryShutdown()` without a callback I'm getting this error:
```
#
# Fatal error in ../deps/v8/src/objects.cc, line 3574
# Check failed: receiver->IsJSFunction().
#
```

Isn't this (no callback) supposed to be handled by the library?

murgatroid99:
The callback is not optional. But we should be checking and throwing a more helpful error instead of crashing.

murgatroid99:
In 1.8.4 this now throws an exception indicating that a callback argument is required.

franciscocpg:
:+1:

PR #143, opened 2018-01-05
Add JS error for missing callback to sever#tryShutdown

murgatroid99:
This improves the situation reported in #142.


murgatroid99:
Travis is having problems, and I am confident that the failures we are seeing are caused by network errors, not our code.

PR #147, opened 2018-01-09
grpc-js-core: fixes and update to use node 9.4 http2 api

kjin:
#### `docs for call.ts`
* Adds a few docs for classes
* Removes `ServiceErrorImpl`, as we can accomplish the same with the `Error` constructor
#### ~~`support propagation cancellation`~~
* ~~Adds support for cancellation propagation~~
#### `split incoming headers on comma for metadata`
* If an incoming metadata value is a single string, split it on the comma delimiter to get multiple values. This appears to be the behavior of the native implementation, as not doing this will fail tests.
#### `add user-agent`
* Self-explanatory
#### `keep up-to-date with node-master`
* Replaces `shutdown` with `close` (this was changed in Node 9.4, and I believe will be backported to Node 8 as well if not already)
#### `clone local metadata before applying filters`
* Self-explanatory
#### `clear deadline timers`
* Timers set to expire actions are now cleared when the actions are completed first
#### `change rstStream to close`
* Accounts for another change that was made in Node 9.4

### DIFF
--- a/packages/grpc-js-core/src/call-stream.ts
+++ b/packages/grpc-js-core/src/call-stream.ts
@@ -156,7 +156,7 @@ export class Http2CallStream extends Duplex implements CallStream {
 
   attachHttp2Stream(stream: http2.ClientHttp2Stream): void {
     if (this.finalStatus !== null) {
-      (stream as any).close(NGHTTP2_CANCEL);
+      stream.close(NGHTTP2_CANCEL);
     } else {
       this.http2Stream = stream;
       stream.on('response', (headers, flags) => {
@@ -328,7 +328,7 @@ export class Http2CallStream extends Duplex implements CallStream {
     if (this.http2Stream !== null && !this.http2Stream.destroyed) {
       /* TODO(murgatroid99): Determine if we want to send different RST_STREAM
        * codes based on the status code */
-      (this.http2Stream as any).close(NGHTTP2_CANCEL);
+      this.http2Stream.close(NGHTTP2_CANCEL);
     }
   }
 
--- a/packages/grpc-js-core/src/call-stream.ts
+++ b/packages/grpc-js-core/src/call-stream.ts
@@ -9,7 +9,7 @@ import {FilterStackFactory} from './filter-stack';
 import {Metadata} from './metadata';
 import {ObjectDuplex} from './object-stream';
 
-const {HTTP2_HEADER_STATUS, HTTP2_HEADER_CONTENT_TYPE} = http2.constants;
+const {HTTP2_HEADER_STATUS, HTTP2_HEADER_CONTENT_TYPE, NGHTTP2_CANCEL} = http2.constants;
 
 export type Deadline = Date | number;
 
@@ -156,7 +156,7 @@ export class Http2CallStream extends Duplex implements CallStream {
 
   attachHttp2Stream(stream: http2.ClientHttp2Stream): void {
     if (this.finalStatus !== null) {
-      stream.rstWithCancel();
+      (stream as any).close(NGHTTP2_CANCEL);
     } else {
       this.http2Stream = stream;
       stream.on('response', (headers, flags) => {
@@ -328,7 +328,7 @@ export class Http2CallStream extends Duplex implements CallStream {
     if (this.http2Stream !== null && !this.http2Stream.destroyed) {
       /* TODO(murgatroid99): Determine if we want to send different RST_STREAM
        * codes based on the status code */
-      this.http2Stream.rstWithCancel();
+      (this.http2Stream as any).close(NGHTTP2_CANCEL);
     }
   }
 
--- a/packages/grpc-js-core/src/call-stream.ts
+++ b/packages/grpc-js-core/src/call-stream.ts
@@ -19,6 +19,8 @@ export interface CallStreamOptions {
   flags: number;
 }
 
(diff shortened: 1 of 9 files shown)


packages/grpc-js-core/src/channel.ts
```diff
@@ -3,16 +3,26 @@ import * as http2 from 'http2';
...
 import {DeadlineFilterFactory} from './deadline-filter';
 import {FilterStackFactory} from './filter-stack';
 import {Metadata, MetadataObject} from './metadata';
 import { MetadataStatusFilterFactory } from './metadata-status-filter';
+import { PropagateFlags } from './index';
+
+export type CallOptions = {
+  // Represents a parent server call.
+  // For our purposes we only need to know of the 'cancelled' event.
+  parent?: EmitterAugmentation0<'cancelled'> & EventEmitter;
```
murgatroid99: We have to be careful doing something like this, and I think it will have to change before the actual 1.0 release. The problem is that if we further specify this type, for propagating other things, we can break users who were obeying the previous type constraint.

kjin: I wasn't planning on making any additional assumptions about what the server-side call object looks like. Given that this is how cancellation is propagated in `grpc` currently (with a 'cancelled' event), I thought this would be a safe and minimal assumption to make.

I can drop this commit entirely and skip tests that rely on it for now -- but I don't think it's necessary assuming that we will eventually have a document detailing the exact interactions between client and server before the 1.0 release anyway. What do you think?

kjin:
@murgatroid99 I've reverted the propagation change and added some other small changes (essentially you can treat this as a new PR). PTAL

packages/grpc-js-core/src/channel.ts
```diff
@@ -204,11 +207,12 @@ export class Http2Channel extends EventEmitter implements Channel {
   private startHttp2Stream(
       methodName: string, stream: Http2CallStream, metadata: Metadata) {
     let finalMetadata: Promise<Metadata> =
-        stream.filterStack.sendMetadata(Promise.resolve(metadata));
+        stream.filterStack.sendMetadata(Promise.resolve(metadata.clone()));
     Promise.all([finalMetadata, this.connect()])
       .then(([metadataValue]) => {
         let headers = metadataValue.toHttp2Headers();
         headers[HTTP2_HEADER_AUTHORITY] = this.authority.hostname;
+        headers[HTTP2_HEADER_USER_AGENT] = `grpc-node/${clientVersion}`;
```
murgatroid99: I think this should be `grpc-node-js` or something like that, so that it's clearly a variant. Also, I think it would be good to pipe through the user agent channel arguments. There are actually two of them: `grpc.primary_user_agent` gets put at the beginning, and `grpc.secondary_user_agent` goes at the end.

packages/grpc-js-core/src/channel.ts
```diff
@@ -150,7 +153,7 @@ export class Http2Channel extends EventEmitter implements Channel {
           return checkServerIdentity(sslTargetNameOverride, cert);
         }
       }
-      subChannel = http2.connect(this.authority, connectionOptions);
+      subChannel = http2.connect(this.authority, connectionOptions) as typeof subChannel;
```
murgatroid99: What's up with this change?

packages/grpc-js-core/src/call-stream.ts
```diff
@@ -156,7 +156,7 @@ export class Http2CallStream extends Duplex implements CallStream {
 
   attachHttp2Stream(stream: http2.ClientHttp2Stream): void {
     if (this.finalStatus !== null) {
-      stream.rstWithCancel();
+      (stream as any).close(NGHTTP2_CANCEL);
```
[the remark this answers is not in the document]
kjin: It's not... I think this was a mistake with the type definitions.

packages/grpc-js-core/src/call.ts
```diff
@@ -7,33 +7,46 @@ import {Status} from './constants';
 import {Metadata} from './metadata';
 import {ObjectReadable, ObjectWritable} from './object-stream';
 
-export interface ServiceError extends Error {
+/**
+ * A type extending the built-in Error object with additional fields.
+ */
+export type ServiceError = {
```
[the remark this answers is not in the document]
kjin: Done

packages/grpc-js-core/src/channel.ts
kjin: Artifact of experimenting earlier... removing.

packages/grpc-js-core/src/channel.ts
kjin: I added them and also moved some code so that `options` doesn't get modified. I didn't see code to add `secondary_user_agent` in grpc-native-core though (I might be mistaken).

kjin:
@murgatroid99 Sorry for the delay, PTAL at the latest commit. I'll fix the type definitions in a later commit as it is taking some time to land the HTTP2 typedefs.

kjin:
@murgatroid99 Actually the new types just came out, so I took the liberty of using them: https://github.com/grpc/grpc-node/pull/147/commits/8108cc606e6d86a3f3e46507d0069231f0a593ee

Will merge once compile completes, as they're mostly minor changes. Note that some HTTP2 changes might not have been backported to 8 yet, which is why I'm using v9 definitions.

PR #148, opened 2018-01-09
grpc-js-core: implement getPeer

kjin:


murgatroid99:
Are you sure this is correct? I don't actually remember, but I thought that the current `getPeer` function returned information corresponding to the actual connected backend, like IP address. This is important because the server that a call is made against is not necessarily the same as the server that the channel is currently connected to.

kjin:
The implementation is solely based on the value of `getPeer()` I observed when running tests; I wasn't able to find much documentation on it. That being said, is it currently possible for the JS client to do what you described?

kjin:
I'm going to close this to prevent PR pollution... it is heavily outdated.

PR #159, opened 2018-01-22
V1.8.x upmerge

murgatroid99:


kjin:
Why is the diff so small compared to the number of commits?

murgatroid99:
Three of them are merge commits and one just changes the version number, which doesn't impact master.

Issue #161, opened 2018-01-23
Replace ArgueJS because of unsafe call of eval()?


StabbarN:
ArgueJS uses `eval()`, see https://github.com/zvictor/ArgueJS/blob/master/argue.js#L203

I'm trying to prevent any usage of eval for all dependencies in an app by setting:
```
window.eval = function() {
  throw new Error("Nope, eval not permitted");
};
```

Are you OK with replacing the usage of ArgueJS (last meaningful update was in May 2013) with something more modern and safe? For example Typescript, flow or manual type checking?
I can see that some files of yours are already converted to Typescript. 

ArgueJS is required here:
https://github.com/grpc/grpc-node/blob/master/packages/grpc-native-core/src/client.js

murgatroid99:
gRPC no longer uses ArgueJS.

PR #162, opened 2018-01-23
Replace argue.js usage with inline argument checking

### LINKED ISSUE #133: grpc@1.8.0 -- cannot read property 'name' of null
We are seeing this error with `grpc@1.8.0`. It looks like arguejs may have some difficulty resolving the argument order inside `makeUnaryRequest`. 

```
"TypeError: Cannot read property 'name' of null
    at (/srv/www/link/releases/20171219165514/node_modules/plaid-escalate client/node_modules/grpc/node_modules/arguejs/argue.js:117:46)
    at ServiceClient.Client.makeUnaryRequest (/srv/www/link/releases/20171219165514/node_modules/plaid-escalate-client/node_modules/grpc/src/client.js:507:14)
    at apply (/srv/www/link/releases/20171219165514/node_modules/plaid-escalate-client/node_modules/grpc/node_modules/lodash/lodash.js:499:17)
    at ServiceClient.wrapper (/srv/www/link/releases/20171219165514/node_modules/plaid-escalate-client/node_modules/grpc/node_modules/lodash/lodash.js:5356:16)
```

It seems like this is a known issue with `arguejs`: https://github.com/zvictor/ArgueJS/issues/23.
The "bug" is just that it does an invalid dereference before it can throw a more useful error. If you change the problem line there to print out "foobar" instead of throwing a NPE, we end up with something like this:

```node
"Error: Incompatible type signature. Expecting ( String, Function, Function, foobar, [Metadata], [Object], Function ), given ( String, Function, Function, Object, Object, Object, Function )
```

It seems like this is still an "irrecoverable error" type scenario, though.

here is how we are creating the client (typescript)

```ts
const root = new $protobufjs.Root();
root.resolvePath = (origin, target) =>
  path.resolve(PATH_TO_PROTO) + '/' + target;
root.loadSync(GRPC_SERVICE_DEFINITION', {
  keepCase: true,
});

const proto = grpc.loadObject(root, {
  enumsAsStrings: false,
  protobufjsVersion: 6,
});
```

We implement a custom promisification wrapper around the grpc prototype... at the end of the day it looks something like:

```ts
function grpcUnaryPromisifier(func: grpc.Request): any {
  // tslint:disable-next-line only-arrow-functions
  return async function promisified(
    // The argument to use
    argument: any,

    // The grpc metadata object, if undefined -- grpc will default to creation
    // of new metadata.
    metadata?: grpc.Metadata,

    // Options for this rpc
    options: grpc.CallOptions = ({} as grpc.CallOptions),
  ): P<any> {
    if (metadata == null) {
      metadata = new grpc.Metadata();
    }

    // tslint:disable no-invalid-this
    const self: EscalateClientInternal = this;
    // tslint:enable no-invalid-this

    // Populate default values into the options.
    options = R.merge({
      deadline: Date.now() + DEFAULT_TIMEOUT_MS,
    }, options);

    // do some logging, retry with exponential backoff, etc.
    try {
      const result = await P.fromCallback(cb => func.call(
            self, argument, metadata, options, cb));
    } catch (e) {
      // log and retthrow
    }

   // return the result
}
```

This all works fine and dandy with the client using `1.7.3`...

```js
     const metadata = new grpc.Metadata();

     if (typeof context.requestId === 'string') {
       metadata.set('request_id', context.requestId);
     }

     plaidGrpcService.client()
     .nameOfPromisifiedMethodAsync(requestObject, metadata, {
       deadline: plaidGrpcService.defaultDeadline(),
     })
     .then(res => {
       // business logic.
     })
```

I _initially_ thought that there was some problem because the calling service was using `grpc@1.8.0` despite the client using `grpc@1.7.3`. Since `npm` does not flatten dependencies, this should be a valid use case.

The problem first arose when we were passing a `grpc.Metadata` that was created by `1.8.0` to this client implementation using `1.7.3`. To test whether this is the problem, I tried changing both the calling service and the client to use `grpc@1.8.0` and this issue persists. With both the service and the client using `1.7.3`, there is no issue. I have not tested what happens when the client uses 1.8.0 and the service uses 1.7.3.

Based upon the error I was able to coerce out of `arguejs`, it _looks_ as if it is not recognizing that the `Object` passed in is actually a `Metadata`. Why this would be, I am not sure. Note that the version of arguejs that `grpc` uses is constant between `1.7.3` and `1.8.0`.

I looked at the diff between [v1.7.3 and v1.8.0](https://github.com/grpc/grpc-node/compare/v1.7.3...v1.8.0) and it seems fairly innocuous, so I am rather at a loss here. 

Help would be appreciated.

murgatroid99:
This fixes #133 and #161. The error messages will be a little less helpful right now, but we can improve that later and I think it's good to get the fix out now.

The dependency on argue.js was actually removed a while ago in grpc/grpc#11707, but for some reason that change never made it out of the v1.4.x branch. So this is also correcting that error.

### DIFF
--- a/packages/grpc-native-core/src/client.js
+++ b/packages/grpc-native-core/src/client.js
@@ -33,7 +33,6 @@
 'use strict';
 
 var _ = require('lodash');
-var arguejs = require('arguejs');
 
 var grpc = require('./grpc_extension');
 
@@ -500,21 +499,27 @@ exports.Client = Client;
 Client.prototype.makeUnaryRequest = function(method, serialize, deserialize,
                                              argument, metadata, options,
                                              callback) {
-  /* While the arguments are listed in the function signature, those variables
-   * are not used directly. Instead, ArgueJS processes the arguments
-   * object. This allows for simple handling of optional arguments in the
-   * middle of the argument list, and also provides type checking. */
-  var args = arguejs({method: String, serialize: Function,
-                      deserialize: Function,
-                      argument: null, metadata: [Metadata, new Metadata()],
-                      options: [Object], callback: Function}, arguments);
-  var call = getCall(this.$channel, method, args.options);
+  if (!(metadata instanceof Metadata)) {
+    callback = options;
+    options = metadata;
+    metadata = new Metadata();
+  }
+  if (options instanceof Function) {
+    callback = options;
+    options = {};
+  }
+  if (!((metadata instanceof Metadata) &&
+	(options instanceof Object) &&
+	(callback instanceof Function))) {
+    throw new Error("Argument mismatch in makeUnaryRequest");
+  }
+  var call = getCall(this.$channel, method, options);
   var emitter = new ClientUnaryCall(call);
-  metadata = args.metadata.clone();
+  metadata = metadata.clone();
   var client_batch = {};
-  var message = serialize(args.argument);
-  if (args.options) {
-    message.grpcWriteFlags = args.options.flags;
+  var message = serialize(argument);
+  if (options) {
+    message.grpcWriteFlags = options.flags;
   }
 
   client_batch[grpc.opType.SEND_INITIAL_METADATA] =
@@ -535,7 +540,7 @@ Client.prototype.makeUnaryRequest = function(method, serialize, deserialize,
     if (status.code === constants.status.OK) {
       if (err) {
         // Got a batch error, but OK status. Something went wrong
-        args.callback(err);
+        callback(err);
         return;
       } else {
         try {
@@ -554,9 +559,9 @@ Client.prototype.makeUnaryRequest = function(method, serialize, deserialize,
       error = new Error(status.details);
       error.code = status.code;
       error.metadata = status.metadata;
-      args.callback(error);
+      callback(error);
     } else {
-      args.callback(null, deserialized);
+      callback(null, deserialized);
     }
     emitter.emit('status', status);
   });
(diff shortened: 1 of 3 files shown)


ntindall:
thanks for the fix @murgatroid99 !

packages/grpc-native-core/src/client.js
```diff
@@ -500,21 +499,27 @@ exports.Client = Client;
...
+    callback = options;
+    options = metadata;
+    metadata = new Metadata();
+  }
+  if (options instanceof Function) {
+    callback = options;
+    options = {};
+  }
+  if (!((metadata instanceof Metadata) &&
+	(options instanceof Object) &&
```
allenhartwig: This causes requests to throw when `options === undefined`. If this is new, intended behavior it is a breaking change from argue.js's behavior. This caused quite the headache for me and my team, as we didn't think to look to the grpc lib originally, because we felt such a breaking change would not be in a minor version bump.

For reference, we were calling the client request as such:
```
function makeCall(rpc, message, meta, options) {
  return new Bluebird((resolve, reject) => {
    client[rpc](message, meta, options, (err, res) => {
        if (err) {
          return reject(err);
        }
        return resolve(res);
    });
  });
}
```

Giving the `options` argument a default of `= {}` resolved the issue, but I don't feel like this should be necessary.

murgatroid99: Thank you for reporting this issue. It was also reported in #178 and a fix is in #179. We will try to publish a patch release soon after that is merged.

PR #169, opened 2018-02-02
Add async_hooks integration for gRPC Server

kjin:
This change adds `async_hooks` integration when it's available (Node 8+), and aims to make almost no impact otherwise. A short explanation of which `AsyncResource` objects are created is at the top of `server.js`.

As the `async_hooks` docs are presently a little hard to digest, I'll try to summarize it here (please correct me if I am mistaken):

- Synchronous continuations in JavaScript are separated from each other by asynchronous tasks
- We sometimes want a way to establish links between these continuations, as typically a continuation "kicks off" a task that leads to another continuation:
  ```js
  /*continuation a*/
  fs.readFile('my-file', (err, contents) => { /*continuation b, caused by a*/ })
  ```
- For example, to support tracing we want to know the ancestry of each continuation to determine the amount of time spent on a single request (group by top-level parents)
- `async_hooks` gives us that information by allowing us to add hooks for:
  1. when an asynchronous task is started (`init`)
  2. when we enter the continuation that follows it (`before`)
  3. when we leave that continuation (`after`)
  4. when the asynchronous task is finished (`destroy`)
- It automatically does this for (almost) all asynchronous tasks in Node core, but modules relying on native components for asynchronous work (such as gRPC) need to implement this on their own
- It allows us to do so with the `AsyncResource` API (an `AsyncResource` is an object associated with an asynchronous task) which:
  - when constructed, emits an `init` event
  - exposes `emitBefore`, `emitAfter`, and `emitDestroy` which corresponds to the ordered list above

### DIFF
--- a/packages/grpc-native-core/src/async-hooks-integration.js
+++ b/packages/grpc-native-core/src/async-hooks-integration.js
@@ -31,16 +31,23 @@ var useAsyncHooks = semver.satisfies(process.version, '>=8');
  */
 
 if (useAsyncHooks) {
-  var asyncHooks = require('async_hooks');
-  module.exports = function createAsyncResourceWrapper(name) {
-    var resource = new asyncHooks.AsyncResource(name);
+  const asyncHooks = require('async_hooks');
+  class GrpcAsyncResource extends asyncHooks.AsyncResource {
+    constructor(name, handle) {
+      super(name);
+      this.handle = handle;
+    }
+  }
+  module.exports = function createAsyncResourceWrapper(name, handle) {
+    let resource = new GrpcAsyncResource(name, handle);
     return {
-      wrap: function(fn) {
+      wrap: (fn) => {
         return function() {
           if (resource) {
             resource.emitBefore();
+            let result;
             try {
-              var result = fn.apply(this, arguments);
+              result = fn.apply(this, arguments);
             } finally {
               resource.emitAfter();
             }
@@ -50,18 +57,20 @@ if (useAsyncHooks) {
           }
         }
       },
-      destroy: function() {
-        if (resource) {
-          resource.emitDestroy();
-          resource = null;
-        }
+      destroy: () => {
+        setImmediate(() => {
+            if (resource) {
+            resource.emitDestroy();
+            resource = null;
+          }
+        });
       }
     };
   }
 } else {
-  var noImpl = {
-    wrap: function(fn) { return fn; },
-    destroy: function() {}
+  const noImpl = {
+    wrap: fn => fn,
+    destroy: () => {}
   };
   module.exports = function createAsyncResourceWrapper() { return noImpl; }
 }
--- a/packages/grpc-native-core/src/async-hooks-integration.js
+++ b/packages/grpc-native-core/src/async-hooks-integration.js
@@ -1,0 +1,67 @@
+/**
+ * @license
+ * Copyright 2018 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+var semver = require('semver');
+var useAsyncHooks = semver.satisfies(process.version, '>=8');
+
+/**
+ * Assuming that async_hooks is available, this file exposes a function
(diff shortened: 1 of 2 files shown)


packages/grpc-native-core/src/async-hooks-integration.js
```diff
@@ -0,0 +1,67 @@
...
+
+if (useAsyncHooks) {
+  var asyncHooks = require('async_hooks');
+  module.exports = function createAsyncResourceWrapper(name) {
+    var resource = new asyncHooks.AsyncResource(name);
+    return {
+      wrap: function(fn) {
+        return function() {
+          if (resource) {
+            resource.emitBefore();
```
ofrobots: Note that I am changing this to be safer here: https://github.com/nodejs/node/pull/18513. But this is ought to be fine for now.

packages/grpc-native-core/src/async-hooks-integration.js
```diff
@@ -0,0 +1,67 @@
...
+            try {
+              var result = fn.apply(this, arguments);
```
ofrobots: nit: declare var at the top so that it doesn't seem to escape from the scope (even though that's how var is actually implemented).

packages/grpc-native-core/src/server.js
```diff
@@ -567,14 +576,15 @@ ServerDuplexStream.prototype.waitForCancel = waitForCancel;
  * @param {grpc.Metadata} metadata Metadata from the client
  */
 function handleUnary(call, handler, metadata) {
+  var asyncResource = createAsyncResourceWrapper('GrpcServerRequest');
```
ofrobots: As suggested by the [official docs](https://nodejs.org/dist/latest-v9.x/docs/api/async_hooks.html#async_hooks_type), it would be best to use the module name as a prefix. Something like 'grpc.ServerRequest'.

packages/grpc-native-core/src/server.js
```diff
@@ -664,22 +677,26 @@ function handleServerStreaming(call, handler, metadata) {
...
     handleError(call, error);
   });
   stream.waitForCancel();
-  handler.func(stream, function(err, value, trailer, flags) {
-    stream.terminate();
-    if (err) {
-      if (trailer) {
-        err.metadata = trailer;
+  asyncResource.wrap(function() {
+    handler.func(stream, function(err, value, trailer, flags) {
```
ofrobots: This use isn't making sense to me. `handler.func` seems to be getting executed in the same continuation as `handleClientStreaming` – it is called synchronously.

It doesn't seem logical to create an `AsyncResource` to wrap a synchronously executed function.

packages/grpc-native-core/src/server.js
```diff
@@ -702,10 +719,15 @@ function handleClientStreaming(call, handler, metadata) {
...
  */
 function handleBidiStreaming(call, handler, metadata) {
+  var asyncResource = createAsyncResourceWrapper('GrpcServerRequest');
   var stream = new ServerDuplexStream(call, metadata, handler.serialize,
                                       handler.deserialize);
+  stream.on('error', asyncResource.destroy);
+  stream.on('finish', asyncResource.destroy);
   stream.waitForCancel();
-  handler.func(stream);
+  asyncResource.wrap(function() {
```
ofrobots: likewise.

packages/grpc-native-core/src/server.js
```diff
@@ -598,8 +608,9 @@ function handleUnary(call, handler, metadata) {
       } else {
         sendUnaryResponse(call, value, handler.serialize, trailer, flags);
       }
+      asyncResource.destroy();
```
ofrobots: Can you elaborate on why it is okay that the destroy doesn't get called on all paths?

packages/grpc-native-core/src/server.js
kjin: This is to mirror what the other handler types do. I don't think there's harm in having nesting here (from the perspective of an agent listening for events with only one type of async resource, there is no concept of nesting) -- and I can't think of a good alternative without changing the behavior of the code to move the invocation `handler.func` to the next turn of the event loop. At the minimum we will definitely want a per-request `grpc.ServerRequest`-type async resource. The `grpc.Server` async resource doesn't seem strictly required at the moment but it follows the precedent set by `http`.

packages/grpc-native-core/src/server.js
kjin: It's not okay 🙃 I've added additional calls to `destroy` if the user function is never called.

packages/grpc-native-core/src/server.js
kjin: Done

packages/grpc-native-core/src/async-hooks-integration.js
kjin: Actually, changing this to `let`/`const` -- I think language-level parity with the rest of the code shouldn't be too important right now.

kjin:
Closing this for now since #234 was landed and released. I've made a note to self to check what needs to be done in the JS layer, if any.

PR #175, opened 2018-02-08
Improve module load error message when the directory does not exist

murgatroid99:
A couple of times, e.g. in grpc/grpc#6435, people have reported errors indicating that the binary module directory doesn't exist at all, which isn't communicated well. This fixes that.

### DIFF
--- a/packages/grpc-native-core/src/grpc_extension.js
+++ b/packages/grpc-native-core/src/grpc_extension.js
@@ -31,17 +31,27 @@ var binding;
 try {
   binding = require(binding_path);
 } catch (e) {
-  var fs = require('fs');
-  var searchPath = path.dirname(path.dirname(binding_path));
-  var searchName = path.basename(path.dirname(binding_path));
-  var foundNames = fs.readdirSync(searchPath);
+  let fs = require('fs');
+  let searchPath = path.dirname(path.dirname(binding_path));
+  let searchName = path.basename(path.dirname(binding_path));
+  let foundNames;
+  try {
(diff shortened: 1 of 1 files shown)


packages/grpc-native-core/src/grpc_extension.js
```diff
@@ -31,17 +31,27 @@ var binding;
...
+  let fs = require('fs');
+  let searchPath = path.dirname(path.dirname(binding_path));
+  let searchName = path.basename(path.dirname(binding_path));
+  let foundNames;
+  try {
+    foundNames = fs.readdirSync(searchPath);
+  } catch (readDirError) {
+    let message = `The gRPC binary module was not installed
+This may be fixed by running "npm rebuild"
+Original error: ${e.message}`;
```
kjin: Should this be `readDirError.message`? (And same with `e.code` below)

kjin: Also, it seems like the first two lines can fit together on a single line.

murgatroid99: No, it's supposed to be `e.message` and `e.code`, because `e` is the original error that we are trying to communicate.

Also, I just made the structure of the error message match the other error. I could merge those lines, though.

murgatroid99: OK, I merged the lines.

kjin: Right, didn't see that 🙃

PR #177, opened 2018-02-09
Fix usage of Protobuf.js Message type in TS file

### LINKED ISSUE #176: Typings issue in version 1.9.0
If I run `tsc -d` on my Typescript source I get the following message:

```bash
../../../../../grpc/index.d.ts(69,50): error TS2314: Generic type 'Message<T>' requires 1 type argument(s).                                                  
```
grpc-node version: 1.9.0
Typescript version: 2.7.1
`tsconfig.json` compiler options:

```js
"compilerOptions": {
    "target": "es6",
    "module": "commonjs",
    "moduleResolution": "node",
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "allowSyntheticDefaultImports": true
  }
```

murgatroid99:
This should fix #176.

### DIFF
--- a/packages/grpc-native-core/index.d.ts
+++ b/packages/grpc-native-core/index.d.ts
@@ -66,7 +66,7 @@ declare module "grpc" {
    * - Anything else becomes the relevant reflection object that ProtoBuf.js would create
    */
   export interface GrpcObject {
-    [name: string]: GrpcObject | typeof Client | Message;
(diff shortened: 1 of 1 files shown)


murgatroid99:
The `Message` type here is a Protobuf.js type, not a general type for any kind of message that might be transmitted by gRPC. And honestly, I put `any` there because I have no idea what that type parameter means, even after reading Protobuf.js's `.d.ts` file.

PR #196, opened 2018-02-27
Add package definition loader function to native package

murgatroid99:
@kjin This is the implementation in the existing native library of the function to load the "package definition" from the protobuf library. As you can see, it calls `makeClientConstructor`, which is the other function that is not currently implemented in the pure JavaScript implementation.

The result type for this looks something like

```typescript
type GrpcPackage = {
  [name: string]: GrpcPackage | typeof Client
}
```

### DIFF
--- a/packages/grpc-native-core/index.js
+++ b/packages/grpc-native-core/index.js
@@ -156,13 +156,13 @@ exports.loadPackageDefinition = function loadPackageDefintion(packageDef) {
   for (const serviceFqn in packageDef) {
     const service = packageDef[serviceFqn];
     const nameComponents = serviceFqn.split('.');
-    const serviceName = nameComponents[-1];
+    const serviceName = nameComponents[nameComponents.length-1];
     let current = result;
-    for (const package in nameComponents.slice(0, -1)) {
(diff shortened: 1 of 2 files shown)


packages/grpc-native-core/index.js
```diff
@@ -146,6 +146,29 @@ exports.load = function load(filename, format, options) {
...
+ * Load a gRPC package definition as a gRPC object hierarchy
+ * @param packageDef grpc~PackageDefinition The package definition object
+ * @return {Object<string, *>} The resulting gRPC object
+ */
+exports.loadPackageDefinition = function loadPackageDefintion(packageDef) {
+  const result = {};
+  for (const serviceFqn in packageDef) {
+    const service = packageDef[serviceFqn];
+    const nameComponents = serviceFqn.split('.');
+    const serviceName = nameComponents[-1];
```
kjin: Did you mean `nameComponents.slice(-1)[0]`?

murgatroid99: Yes, something like that. I forgot that wasn't a thing in JavaScript.

PR #198, opened 2018-02-28
grpc-protobufjs: add d.ts to package

kjin:
Just adding the d.ts file to the package exports so we can access types from outside (and lowering the version number)

@murgatroid99 Can you publish a placeholder for this package?


murgatroid99:
Why do you want a placeholder package?

kjin:
@murgatroid99 I want to link to @grpc/protobufjs as a dependency so I can get `*Definition` types in @grpc/js-core.

PR #199, opened 2018-03-01
Move API tests back to grpc-native-core

kjin:
This PR moves API tests back to the `grpc-native-core` package directory. Hopefully the commit breakup makes this easy to digest.

This time I ran https://github.com/grpc/grpc/blob/master/tools/run_tests/helper_scripts/run_grpc-node.sh to make sure #118 doesn't need to happen again.


murgatroid99:
This goes a little too far; we should only delete the contents of `test/api`. In particular, the interop client and server are referenced from a different repo, so we don't want to get rid of those. And that also uses the fixtures. On top of that, you're deleting a few tests without moving them. If there is no equivalent replacement in the target directory, I'd rather just leave them there instead of deleting them.

kjin:
The deleted tests (`credentials_api_test.js` and `file_load_test.js`) are actually completely duplicated from existing tests (`credentials_test.js` and `surface_test.js` respectively) at a time when I was trying to break them down to make it easier to support both gRPC versions. Besides these, all other tests should be preserved exactly (except for path changes).

I moved the interop tests back to the repo test directory. Currently running the helper script again to make sure there are no bad things happening.

murgatroid99:
I guess I wasn't clear. The `native_native` fixture is referenced from the other repo, so we need to keep that too. It's OK if it doesn't actually do anything for now, though. And because the interop tests are there, you shouldn't delete `test/package.json` either.

PR #206, opened 2018-03-06
grpc-js: add makeClientConstructor and loadPackageDefinition

kjin:
This change adds `makeClientConstructor` & `loadPackageDefinition` to grpc-js-core, and makes its interface a subset of `grpc`.

### DIFF
--- a/packages/grpc-js-core/src/client.ts
+++ b/packages/grpc-js-core/src/client.ts
@@ -8,20 +8,24 @@ import {ChannelCredentials} from './channel-credentials';
 import {Status} from './constants';
 import {Metadata} from './metadata';
 
+// This symbol must be exported (for now).
+// See: https://github.com/Microsoft/TypeScript/issues/20080
+export const kChannel = Symbol();
+
 export interface UnaryCallback<ResponseType> {
   (err: ServiceError|null, value?: ResponseType): void;
 }
 
 export class Client {
-  private readonly channel: Channel;
+  private readonly [kChannel]: Channel;
   constructor(
       address: string, credentials: ChannelCredentials,
       options: Partial<ChannelOptions> = {}) {
-    this.channel = new Http2Channel(address, credentials, options);
+    this[kChannel] = new Http2Channel(address, credentials, options);
   }
 
   close(): void {
-    this.channel.close();
+    this[kChannel].close();
   }
 
   waitForReady(deadline: Date|number, callback: (error: Error|null) => void):
@@ -29,7 +33,7 @@ export class Client {
     let cb: (error: Error|null) => void = once(callback);
     let callbackCalled = false;
     let timer: NodeJS.Timer | null = null;
-    this.channel.connect().then(() => {
+    this[kChannel].connect().then(() => {
       if (timer) {
         clearTimeout(timer);
       }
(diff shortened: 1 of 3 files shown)


packages/grpc-js-core/src/client.ts
```diff
@@ -8,28 +8,32 @@ import {ChannelCredentials} from './channel-credentials';
...
+// See: https://github.com/Microsoft/TypeScript/issues/20080
+export const kChannel = Symbol();
+
 export interface UnaryCallback<ResponseType> {
   (err: ServiceError|null, value?: ResponseType): void;
 }
 
 export class Client {
-  private readonly channel: Channel;
+  private readonly [kChannel]: Channel;
```
kjin: Currently investigating whether this is needed (I changed this while experimenting with the ServiceClient class)

kjin: Ok, now I remember -- this is necessary because of this line.

https://github.com/grpc/grpc-node/pull/206/files#diff-bef77206fce0848d6ef29377963c0a10R115

packages/grpc-js-core/src/make-client.ts
```diff
@@ -0,0 +1,178 @@
...
+        method_type = 'server_stream';
+      } else {
+        method_type = 'unary';
+      }
+    }
+    const serialize = attrs.requestSerialize;
+    const deserialize = attrs.responseDeserialize;
+    const method_func = _.partial(requester_funcs[method_type], attrs.path,
+                                serialize, deserialize);
+    if (classOptions.deprecatedArgumentOrder) {
```
murgatroid99: At least for now, I would prefer not to accept this option, and not to have these wrapping functions at all.

packages/grpc-js-core/src/make-client.ts
```diff
@@ -0,0 +1,178 @@
...
+ * @param methods An object mapping method names to
+ *     method attributes
+ * @param serviceName The fully qualified name of the service
+ * @param classOptions An options object.
+ * @return New client constructor, which is a subclass of
+ *     {@link grpc.Client}, and has the same arguments as that constructor.
+ */
+export function makeClientConstructor(
+    methods: ServiceDefinition, serviceName: string,
+    classOptions: ClientClassOptions): ServiceClientConstructor {
```
murgatroid99: I don't even know if this makes sense, but is it possible to specify that the `methods` argument to this function is the same object as the `service` property attached to the return value?

kjin: Hmm... I don't think TS is expressive enough for that.

packages/grpc-js-core/src/make-client.ts
kjin: Done

packages/grpc-js-core/src/make-client.ts
```diff
@@ -0,0 +1,141 @@
...
+
+export interface MethodDefinition<RequestType, ResponseType> {
+  path: string;
+  requestStream: boolean;
+  responseStream: boolean;
+  requestSerialize: Serialize<RequestType>;
+  responseSerialize: Serialize<ResponseType>;
+  requestDeserialize: Deserialize<RequestType>;
+  responseDeserialize: Deserialize<ResponseType>;
+  originalName: string;
```
murgatroid99: This should be optional

kjin:
@murgatroid99 I changed it because `[methodName: string]: Function` currently isn't visible on exported objects... should be now

PR #209, opened 2018-03-09
Make interop tests use new proto loader, run them with pure js client

murgatroid99:
This also includes some fixes to the relevant packages. With these changes, the interop sanity test suite passes with both client implementations. I haven't set these tests to run in the CI yet because the tests with the pure JS client implementation will only work on Node 9, so we have to make some changes to how we run the tests to facilitate that.


packages/grpc-js-core/src/channel.ts
```diff
@@ -224,7 +225,14 @@ export class Http2Channel extends EventEmitter implements Channel {
     Promise.all([finalMetadata, this.connect()])
       .then(([metadataValue]) => {
         let headers = metadataValue.toHttp2Headers();
-        headers[HTTP2_HEADER_AUTHORITY] = this.authority.hostname;
+        let host: string;
+        // TODO(murgatroid99): Add more centralized handling of channel options
+        if (this.options['grpc.default_authority']) {
+          host = this.options['grpc.default_authority'] as string;
```
kjin: Can you add this as an entry in the type def? https://github.com/grpc/grpc-node/pull/209/files#diff-6270a066e64c77ba494d58b714011deeL40

PR #215, opened 2018-03-13
Improve performance of batch operation serializer

drobertduke:
This is a follow-up to #59, addressing the performance penalties of the code I wrote to enforce a certain order in batch operations, as discussed here: https://github.com/grpc/grpc-node/pull/59#issuecomment-369114921

This replaces the worst-performing lodash and standard library calls. In micro-benchmarks, _startBatchIfReady runs in about half the time after these changes are applied.


murgatroid99:
Can you rebase this change against the v1.10.x branch and either update this PR or send a new one?

drobertduke:
@murgatroid99 The code this is based on is not in v1.10.x, #59 was merged to master after 1.10.x was cut.

murgatroid99:
You're right. Never mind.

PR #219, opened 2018-03-14
grpc-js: add google credentials implementation

kjin:
This PR:

1. Fixes a unit test for CallStream to reflect event name changes in a previous PR
2. Makes Call/ChannelCredentials generic on options type (defaults to {})
3. Implements `createFromGoogleCredential`
4. Documents public API

### DIFF
--- a/packages/grpc-js-core/src/call-credentials.ts
+++ b/packages/grpc-js-core/src/call-credentials.ts
@@ -2,33 +2,33 @@ import {map, reduce} from 'lodash';
 
 import {Metadata} from './metadata';
 
-export type CallMetadataGenerator =
-    (options: {}, cb: (err: Error|null, metadata?: Metadata) => void) =>
+export type CallMetadataGenerator<T extends {} = {}> =
+    (options: T, cb: (err: Error|null, metadata?: Metadata) => void) =>
         void;
 
 /**
  * A class that represents a generic method of adding authentication-related
  * metadata on a per-request basis.
  */
-export interface CallCredentials {
+export interface CallCredentials<T extends {} = {}> {
   /**
    * Asynchronously generates a new Metadata object.
    * @param options Options used in generating the Metadata object.
    */
-  generateMetadata(options: {}): Promise<Metadata>;
+  generateMetadata(options: T): Promise<Metadata>;
   /**
    * Creates a new CallCredentials object from properties of both this and
    * another CallCredentials object. This object's metadata generator will be
    * called first.
    * @param callCredentials The other CallCredentials object.
    */
-  compose(callCredentials: CallCredentials): CallCredentials;
+  compose<S>(callCredentials: CallCredentials<S>): CallCredentials<S&T>;
 }
 
-class ComposedCallCredentials implements CallCredentials {
-  constructor(private creds: CallCredentials[]) {}
+class ComposedCallCredentials<T> implements CallCredentials<T> {
+  constructor(private creds: CallCredentials<T>[]) {}
 
-  async generateMetadata(options: {}): Promise<Metadata> {
+  async generateMetadata(options: T): Promise<Metadata> {
     let base: Metadata = new Metadata();
     let generated: Metadata[] = await Promise.all(
         map(this.creds, (cred) => cred.generateMetadata(options)));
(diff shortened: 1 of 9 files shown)


packages/grpc-js-core/src/index.ts
```diff
@@ -5,36 +5,86 @@ import { Client } from './client';
 import { Status} from './constants';
 import { makeClientConstructor, loadPackageDefinition } from './make-client';
 import { Metadata } from './metadata';
+import { IncomingHttpHeaders } from 'http';
 
-const notImplementedFn = () => { throw new Error('Not implemented'); };
+export interface OAuth2Client {
+  getRequestMetadata: (url: string) => Promise<{ Authorization: string }>;
```
murgatroid99: Unfortunately, in order to match the API of the existing library, it has to accept the type of `OAuth2Client` that would be created by `google-auth-library` version 0.9.x. Maybe in the future it could be more flexible and accept both, but for now we should focus on matching the existing API.

packages/grpc-js-core/src/call-credentials.ts
```diff
@@ -2,33 +2,33 @@ import {map, reduce} from 'lodash';
...
+export type CallMetadataGenerator<T extends {} = {}> =
+    (options: T, cb: (err: Error|null, metadata?: Metadata) => void) =>
         void;
 
 /**
  * A class that represents a generic method of adding authentication-related
  * metadata on a per-request basis.
  */
-export interface CallCredentials {
+export interface CallCredentials<T extends {} = {}> {
```
murgatroid99: The generic parameters here and on the `ChannelCredentials` aren't really necessary. They don't even seem to be used for the `ChannelCredentials`, and the `options` parameter for `generateMetadata` and `CallMetadataGenerator` should just be `{ service_url: string }`.

packages/grpc-js-core/test/test-call-stream.ts
```diff
@@ -103,16 +103,16 @@ describe('CallStream', () => {
     assert2.afterMustCallsSatisfied(done);
   });
 
-  it('should end a call with an error if a stream was closed', (done) => {
+  describe('should end a call with an error if a stream was closed', () => {
     const c = http2.constants;
     const s = Status;
     const errorCodeMapping = {
-      [c.NGHTTP2_NO_ERROR]: s.INTERNAL,
+      [c.NGHTTP2_NO_ERROR]: s.OK,
```
murgatroid99: This change isn't correct. A `NO_ERROR` HTTP/2 `RST_STREAM` code maps to an `INTERNAL` gRPC status code because closing a stream with a `RST_STREAM` frame instead of a trailers frame with the proper information is considered a gRPC protocol error.

packages/grpc-js-core/test/test-call-stream.ts
```diff
@@ -103,16 +103,16 @@ describe('CallStream', () => {
...
       [c.NGHTTP2_PROTOCOL_ERROR]: s.INTERNAL,
       [c.NGHTTP2_INTERNAL_ERROR]: s.INTERNAL,
       [c.NGHTTP2_FLOW_CONTROL_ERROR]: s.INTERNAL,
       [c.NGHTTP2_SETTINGS_TIMEOUT]: s.INTERNAL,
-      [c.NGHTTP2_STREAM_CLOSED]: null,
+      [c.NGHTTP2_STREAM_CLOSED]: s.INTERNAL,
```
murgatroid99: The gRPC protocol spec doesn't specify a mapping for `STREAM_CLOSED`, so we shouldn't assert that the value is anything in particular in the tests. I think we should just skip this, at least for now.

packages/grpc-js-core/src/call-credentials.ts
kjin: Is it guaranteed that the only option accepted by `CallCredentials` is `service_url`? I thought that since `createFromMetadataGenerator` is exposed to users that it can take arbitrary options.

murgatroid99: That option is provided by `CallCredentials` to user callbacks, so we decide what fields are there.

packages/grpc-js-core/test/test-call-stream.ts
kjin: Done -- skipping this now.

packages/grpc-js-core/src/call-stream.ts
```diff
@@ -271,9 +271,6 @@ export class Http2CallStream extends Duplex implements CallStream {
         let code: Status;
         let details = '';
         switch (errorCode) {
-          case http2.constants.NGHTTP2_NO_ERROR:
-            code = Status.OK;
```
murgatroid99: This shouldn't be removed. It should be `Status.INTERNAL`.

packages/grpc-js-core/test/test-call-stream.ts
kjin: Ok -- this change was made in response to [this `case` addition](https://github.com/grpc/grpc-node/blame/master/packages/grpc-js-core/src/call-stream.ts#L274-L276) in a previous PR, which was done to support the JS implementation passing tests. I've gone back to see exactly which test case failed because NO_ERROR was mapped to INTERNAL, but couldn't reproduce it. So I'll change it back as you mentioned and remove that `case`.

packages/grpc-js-core/src/index.ts
kjin: Done

murgatroid99:
I don't know if it pushes the scope of this change too far, but we will also have to change the call credentials filter to actually pass the `service_url` to the `generateMetadata` method. Then `service_url` won't have to be listed as optional.

kjin:
I set `service_url` to `''` and amended the TODO.

murgatroid99:
OK, that's probably good for now. I think we still want the real implementation to get the relevant interop tests to pass, so I'll try to figure that out after this PR is merged.

PR #220, opened 2018-03-14
test: test grpc-js in ci

kjin:


murgatroid99:
I'm rerunning the tests because some stuff has changed in the intervening time, but I would like to get this up and running.

PR #226, opened 2018-03-16
Added a 100MB maxBuffer since the default 200kb is too small

OlivierBoucher:
Hi guys,

I ran into this issue where stdout maxBuffer is too small on a large proto codebase. This fixes the issue, 1MB is fair but I can adjust it per your liking.

Also, was not sure if master is the good branch, let me know and I'll re-submit

### DIFF
--- a/packages/grpc-tools/bin/protoc_plugin.js
+++ b/packages/grpc-tools/bin/protoc_plugin.js
@@ -32,7 +32,7 @@ var exe_ext = process.platform === 'win32' ? '.exe' : '';
 
 var plugin = path.resolve(__dirname, 'grpc_node_plugin' + exe_ext);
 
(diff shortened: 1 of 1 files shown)


murgatroid99:
What is the impact of having a maxBuffer size that is too small?

OlivierBoucher:
The process crashes

murgatroid99:
I think that if the impact is that bad, we should probably have a wider safety margin. This script should only be run on development machines, and it only spawns one process, so I think 100MB or even 1GB would be reasonable.

OlivierBoucher:
@murgatroid99 friendly ping -- submitted the proper changes

olivierboucher:
@murgatroid99 friendly ping. This issue is still on going since grpc-tools has to be released

PR #229, opened 2018-03-16
Checking node version during tests to see if nvm worked properly.

nicolasnoble:


murgatroid99:
What about the Windows script?

nicolasnoble:
One platform at a time :)

PR #231, opened 2018-03-20
Pure JS: Pass real service_url to call credentials

murgatroid99:
I also added the `host` call option, because that impacts the same code paths.

### DIFF
--- a/packages/grpc-js-core/src/call-credentials-filter.ts
+++ b/packages/grpc-js-core/src/call-credentials-filter.ts
@@ -9,7 +9,7 @@ import {Metadata} from './metadata';
 export class CallCredentialsFilter extends BaseFilter implements Filter {
   private serviceUrl: string;
   constructor(
-      private readonly channel: Http2Channel,
+      private readonly credentials: CallCredentials,
       private readonly host: string,
       private readonly path: string) {
     super();
@@ -27,7 +27,7 @@ export class CallCredentialsFilter extends BaseFilter implements Filter {
   }
 
   async sendMetadata(metadata: Promise<Metadata>): Promise<Metadata> {
-    let credsMetadata = this.credentials.generateMetadata({ service_url: serviceUrl });
+    let credsMetadata = this.credentials.generateMetadata({ service_url: this.serviceUrl });
     let resultMetadata = await metadata;
     resultMetadata.merge(await credsMetadata);
     return resultMetadata;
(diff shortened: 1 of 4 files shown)


packages/grpc-js-core/src/channel.ts
```diff
@@ -267,11 +270,12 @@ export class Http2Channel extends EventEmitter implements Channel {
     let finalOptions: CallStreamOptions = {
       deadline: options.deadline === undefined ? Infinity : options.deadline,
       credentials: options.credentials || CallCredentials.createEmpty(),
-      flags: options.flags || 0
+      flags: options.flags || 0,
+      host: options.host || defaultAuthority
```
kjin: Does this need to be `this.defaultAuthority`?

packages/grpc-js-core/src/call-credentials-filter.ts
```diff
@@ -7,13 +7,27 @@ import {BaseFilter, Filter, FilterFactory} from './filter';
 import {Metadata} from './metadata';
 
 export class CallCredentialsFilter extends BaseFilter implements Filter {
-  constructor(private readonly credentials: CallCredentials) {
+  private serviceUrl: string;
+  constructor(
```
kjin: Does `CallCredentials` not contain this information? If it does maybe we should prefer using that.

In any case, the call site for this constructor probably needs to be updated.

packages/grpc-js-core/src/call-credentials-filter.ts
```diff
@@ -7,13 +7,27 @@ import {BaseFilter, Filter, FilterFactory} from './filter';
...
+      private readonly channel: Http2Channel,
+      private readonly host: string,
+      private readonly path: string) {
     super();
+    let splitPath: string[] = path.split('/');
+    let serviceName: string = '';
+    /* The standard path format is "/{serviceName}/{methodName}", so if we split
+     * by '/', the first item should be empty and the second should be the
+     * service name */
+    if (splitPath.length >= 2) {
```
kjin: Is `serviceName` meant to be empty if this is not the case?

murgatroid99: yes

packages/grpc-js-core/src/call-credentials-filter.ts
murgatroid99: No, the service URL is a property of the call, not the credentials.

And I fixed the constructor and its call site.

PR #234, opened 2018-03-20
fix: avoid using deprecated Nan APIs

ofrobots:
WIP, but opening early to get early feedback, and to kick off the CI. I intend to revert the `-Wno-deprecated-declarations` removal before this lands.

Starting with Nan 2.9.x certain Nan::Callback::Call APIs are
deprecated. Instead there are mechanisms in place that allow native
modules to preserve async context across async calls.

### DIFF
--- a/packages/grpc-native-core/ext/call.h
+++ b/packages/grpc-native-core/ext/call.h
@@ -103,6 +103,7 @@ struct tag {
       v8::Local<v8::Value> call_value);
   ~tag();
   Nan::Callback *callback;
+  Nan::AsyncResource *async_resource;
   OpVec *ops;
   Call *call;
   Nan::Persistent<v8::Value, Nan::CopyablePersistentTraits<v8::Value>>
--- a/packages/grpc-native-core/ext/call_credentials.cc
+++ b/packages/grpc-native-core/ext/call_credentials.cc
@@ -233,7 +233,7 @@ NAUV_WORK_CB(SendPluginCallback) {
         // Get Local<Function> from Nan::Callback*
         **plugin_callback};
     Nan::Callback *callback = state->callback;
-    callback->Call(argc, argv);
+    callback->Call(argc, argv, data->async_resource);
     delete data;
   }
 }
@@ -245,11 +245,10 @@ int plugin_get_metadata(
     grpc_metadata creds_md[GRPC_METADATA_CREDENTIALS_PLUGIN_SYNC_MAX],
     size_t *num_creds_md, grpc_status_code *status,
     const char **error_details) {
+  HandleScope scope;
   plugin_state *p_state = reinterpret_cast<plugin_state *>(state);
-  plugin_callback_data *data = new plugin_callback_data;
-  data->service_url = context.service_url;
-  data->cb = cb;
-  data->user_data = user_data;
+  plugin_callback_data *data =
+      new plugin_callback_data(context.service_url, cb, user_data);
 
   uv_mutex_lock(&p_state->plugin_mutex);
   p_state->pending_callbacks->push(data);
--- a/packages/grpc-native-core/ext/node_grpc.cc
+++ b/packages/grpc-native-core/ext/node_grpc.cc
@@ -57,6 +57,7 @@ typedef struct log_args {
 
 typedef struct logger_state {
   Nan::Callback *callback;
+  Nan::AsyncResource *async_resource;
   std::queue<log_args *> *pending_args;
   uv_mutex_t mutex;
   uv_async_t async;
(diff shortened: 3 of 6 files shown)


thelinuxfoundation:

* User @ofrobots isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

packages/grpc-native-core/ext/node_grpc.cc
```diff
@@ -252,6 +253,7 @@ NAN_METHOD(SetDefaultLoggerCallback) {
     grpc_logger_state.logger_set = true;
   }
   grpc_logger_state.callback = new Nan::Callback(info[0].As<v8::Function>());
+  grpc_logger_state.async_resource = new Nan::AsyncResource("grpc:logger");
```
ofrobots: I didn't find a delete of `grpc_logger_state.callback`, but I didn't look too hard either. Wherever we free one, we should free the other.

packages/grpc-native-core/ext/server.cc
```diff
@@ -318,8 +318,9 @@ NAN_METHOD(Server::TryShutdown) {
   Server *server = ObjectWrap::Unwrap<Server>(info.This());
   if (server->wrapped_server == NULL) {
     // Server is already shut down. Call callback immediately.
```
ofrobots: Note that this looks like an instance of Zalgo. Its an edge case though.

ofrobots:
Hello @thelinuxfoundation bot. Please re-verify.

packages/grpc-native-core/ext/call_credentials.cc
```diff
@@ -245,11 +245,10 @@ int plugin_get_metadata(
     grpc_metadata creds_md[GRPC_METADATA_CREDENTIALS_PLUGIN_SYNC_MAX],
     size_t *num_creds_md, grpc_status_code *status,
     const char **error_details) {
+  HandleScope scope;
```
murgatroid99: I don't think we need a `HandleScope` here because this function doesn't call any Node APIs.

packages/grpc-native-core/ext/call.h
```diff
@@ -98,7 +98,7 @@ class Op {
 };
 
 typedef std::vector<unique_ptr<Op>> OpVec;
-struct tag {
+struct tag : public Nan::AsyncResource {
```
murgatroid99: Why have `tag` inherit from `Nan::AsyncResource`, instead of just having a `Nan::AsyncResource` member?

packages/grpc-native-core/ext/call_credentials.cc
ofrobots: other reason: `tag` represents an async resource. So it makes sense to be a subclass of `AsyncResource`.

packages/grpc-native-core/ext/call.h
ofrobots: Other reason: `tag` represents an async operation, so it makes sense for it to be a subclass of `AsyncResource`.

murgatroid99: We don't explicitly delete any of the members. They get deleted by default when the tag object is deleted. And subclassing merely for the purpose of categorization isn't really compelling to me, so I would prefer to use composition here.

ofrobots:
> In general, are the AsyncResource names supposed to have any semantic meaning at the JS level?

Documented [here](https://github.com/nodejs/node/blob/master/doc/api/async_hooks.md#type). These represent the type of asynchronous operation that caused an async hop. This module introduces async hops of its own via direct usage of the uv event loops. The specified type string will be provided to any debug or diagnostic tool that is monitoring async execution via async_hooks. Ultimately this would be provided to the user where they would be able to distinguish async hops due to timers, net, etc. vs. gRPC.

packages/grpc-native-core/ext/call.h
ofrobots: Composition it is.

packages/grpc-native-core/ext/call_credentials.cc
ofrobots: Fix in Nan: https://github.com/nodejs/nan/pull/757 which will take some time to review / land / release. For now let's use a HandleScope ourselves.

ofrobots:
@murgatroid99 I am unable to reproduce the test failures locally. Any ideas?

murgatroid99:
I made some changes in #248 that touch the same code that this PR touches, and I think that PR also likely fixed the crashes you are seeing here. So if you could merge from master, I think you can get this change to work.

PR #241, opened 2018-03-28
Hide symbols.

### LINKED ISSUE #240: The native module exports too many symbols.
We need to compile the module while hiding symbols as much as possible, to avoid polluting node's runtime global namespace.

nicolasnoble:
Fixes #240.

### DIFF
--- a/packages/grpc-native-core/templates/binding.gyp.template
+++ b/packages/grpc-native-core/templates/binding.gyp.template
@@ -58,7 +58,7 @@
       % for arg, prop in [('CPPFLAGS', 'cflags'), ('LDFLAGS', 'ldflags')]:
       %  if defaults['global'].get(arg, None) is not None:
       '${prop}': [
-        % for item in defaults['global'].get(arg).split():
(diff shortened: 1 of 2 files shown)


packages/grpc-native-core/templates/binding.gyp.template
```diff
@@ -45,7 +45,7 @@
           % for arg, prop in [('CPPFLAGS', 'cflags'), ('DEFINES', 'defines')]:
           %  if args.get(arg, None) is not None:
           '${prop}': [
-            % for item in args.get(arg).split():
```
murgatroid99: This is redundant with the one on line 61.

packages/grpc-native-core/templates/binding.gyp.template
```diff
@@ -88,7 +88,7 @@
           % for arg, prop in [('CPPFLAGS', 'cflags'), ('DEFINES', 'defines'), ('LDFLAGS', 'ldflags')]:
           %  if configs['gcov'].get(arg, None) is not None:
           '${prop}': [
-            % for item in configs['gcov'].get(arg).split():
```
murgatroid99: This is also redundant with line 61.

nicolasnoble:
Done.

PR #242, opened 2018-03-28
Add README files for new packages

murgatroid99:
This just adds some basic documentation to the two newer packages that we want to publish soon.

### DIFF
--- a/packages/grpc-protobufjs/README.md
+++ b/packages/grpc-protobufjs/README.md
@@ -1,0 +1,16 @@
+# gRPC Protobuf Loader
+
+## Installation
+
+```sh
+npm install @grpc/proto-loader
+```
+
+## Usage
+
+```js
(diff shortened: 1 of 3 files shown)


packages/grpc-protobufjs/README.md
```diff
@@ -0,0 +1,19 @@
...
+npm install @grpc/proto-loader
+```
+
+## Usage
+
+```js
+const protoLoader = require('@grpc/proto-loader');
+const grpcLibrary = require('grpc');
+// OR
+const grpcLibrary = require('@grpc/grpc-js
```
nicolasnoble: missing the `');` at the end of that line

nicolasnoble: Hah, apparently I can fix it myself! Neat!

murgatroid99:
If I comment with the CLA checker check again?

PR #248, opened 2018-04-02
Fix object lifetime and memory management problems in server code

murgatroid99:
This includes a few fixes:

 - Ensure that client and server credentials construction arguments are null-terminated.
 - Ensure that the callback for `Server#tryShutdown` is always called asynchronously.
 - Ensure that the server doesn't get garbage collected while it is running, and remove server destructor code the server down.

### DIFF
--- a/packages/grpc-native-core/ext/server_credentials.cc
+++ b/packages/grpc-native-core/ext/server_credentials.cc
@@ -16,6 +16,8 @@
  *
  */
 
+#include <vector>
+
 #include <nan.h>
 #include <node.h>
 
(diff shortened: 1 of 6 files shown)


packages/grpc-native-core/ext/server_credentials.cc
```diff
@@ -145,36 +147,34 @@ NAN_METHOD(ServerCredentials::CreateSsl) {
   }
   Local<Array> pair_list = Local<Array>::Cast(info[1]);
   uint32_t key_cert_pair_count = pair_list->Length();
-  grpc_ssl_pem_key_cert_pair *key_cert_pairs =
-      new grpc_ssl_pem_key_cert_pair[key_cert_pair_count];
-
+  grpc_ssl_pem_key_cert_pair key_cert_pairs[key_cert_pair_count];
+  StringOrNull key_strings[key_cert_pair_count];
```
nicolasnoble:
```
#include <memory>
[...]
std::auto_ptr<StringOrNull> key_strings = new StringOrNull[key_cert_pair_count];
std::auto_ptr<StringOrNull> cert_strings = new StringOrNull[key_cert_pair_count];
```

And then later use `cert_strings[i]->assign()`.

nicolasnoble:
Also, from tests:

```
E0402 15:50:47.195617365   31663 server.cc:1247]             assertion failed: gpr_atm_acq_load(&server->shutdown_flag) || !server->listeners
```

Issue #136, opened 2017-12-21
cannot build on FreeBSD


drook:
(in case you will think "some guy just keeps posting about my app not building on abandonware" - I really have no live/decent Linuxes there, so please don't think it's some kind of trolling - it's not)

gcc:
````GCC
Using built-in specs.
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/usr/local/libexec/gcc5/gcc/x86_64-portbld-freebsd11.0/5.4.0/lto-wrapper
Target: x86_64-portbld-freebsd11.0
Configured with: /wrkdirs/usr/ports/lang/gcc/work/gcc-5.4.0/configure --disable-bootstrap --disable-nls --enable-gnu-indirect-function --libdir=/usr/local/lib/gcc5 --libexecdir=/usr/local/libexec/gcc5 --program-suffix=5 --with-as=/usr/local/bin/as --with-gmp=/usr/local --with-gxx-include-dir=/usr/local/lib/gcc5/include/c++/ --with-ld=/usr/local/bin/ld --with-pkgversion='FreeBSD Ports Collection' --with-system-zlib --with-ecj-jar=/usr/local/share/java/ecj-4.5.jar --enable-languages=c,c++,objc,fortran,java --prefix=/usr/local --localstatedir=/var --mandir=/usr/local/man --infodir=/usr/local/info/gcc5 --build=x86_64-portbld-freebsd11.0
Thread model: posix
gcc version 5.4.0 (FreeBSD Ports Collection)
````

````build log
> npm install grpc

> grpc@1.8.0 install /usr/home/emz/node_modules/grpc
> node-pre-gyp install --fallback-to-build --library=static_library

node-pre-gyp ERR! Tried to download(403): https://storage.googleapis.com/grpc-precompiled-binaries/node/grpc/v1.8.0/node-v51-freebsd-x64-unknown.tar.gz
node-pre-gyp ERR! Pre-built binaries not found for grpc@1.8.0 and node@7.8.0 (node-v51 ABI, unknown) (falling back to source compile with node-gyp)
gmake: вход в каталог «/usr/home/emz/node_modules/grpc/build»
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/surface/init.o
  CXX(target) Release/obj.target/grpc/deps/grpc/src/core/lib/backoff/backoff.o
[...]
  CXX(target) Release/obj.target/gpr/deps/grpc/src/core/lib/support/tmpfile_windows.o
[...]

PR #255, opened 2018-04-04
Fixing zdefs.

### LINKED ISSUE #254: zdefs is a gcc-only flag.
When compiling the native extension using clang, the `zdefs` flag is unrecognized, and the compilation will fail.

nicolasnoble:
Fixes #254.


murgatroid99:
Also probably fixes #136.

PR #263, opened 2018-04-06
client lib integration: monkeypatches for google-gax & @google-cloud/common-grpc

kjin:
This PR changes `use-grpc-js.js` to monkeypatch google-gax and @google-cloud/common-grpc to meaningfully use functions exported by `@grpc-js` and `@grpc/proto-loader`. The eventual goal is that the contents of these monkeypatches are directly merged into the codebases themselves.

### DIFF
--- a/test/client-libraries-integration/init.sh
+++ b/test/client-libraries-integration/init.sh
@@ -3,11 +3,13 @@
 npm install
 
 for dir in $(node -p "require('./repositories.json').join('\n')"); do
+  pushd libs
   if [ ! -d $dir ]; then
     git clone https://github.com/googleapis/$dir
+    pushd $dir
+    npm install
+    popd
   fi
-  pushd $dir
-  npm install
   popd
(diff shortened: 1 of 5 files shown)


test/client-libraries-integration/init.sh
```diff
@@ -3,11 +3,13 @@
...
   if [ ! -d $dir ]; then
     git clone https://github.com/googleapis/$dir
+    pushd $dir
+    npm install
+    popd
   fi
-  pushd $dir
-  npm install
   popd
-  node --require ./use-grpc-js.js $(npm bin)/_mocha --timeout 60000 $dir/system-test/*.js
```
murgatroid99: Does this depend on any specific previously set values of `GCLOUD_PROJECT`?

kjin: Some tests seem to already depend on the previously set value of `GCLOUD_PROJECT` itself, not exclusively through the smoke test variable. This just makes it so that we don't have to keep setting multiple environmental variables all the time.

In some cases the project needs to be set up in a certain way as a pre-requisite. I'm still trying to figure out the exact conditions.

PR #271, opened 2018-04-11
client library integration: adjust options

kjin:
This PR contains changes that help client libraries pass when using methods on the new gRPC protobuf package.

- ~~The move of `_.camelCase` seemed appropriate as it matches what I see here: https://github.com/grpc/grpc-node/blob/master/packages/grpc-native-core/src/protobuf_js_6_common.js#L106-L121~~
- I've updated the monkeypatch implementations of methods accessing `grpc.load[Object]` to have more thorough conversion options. `oneofs: true` in particular seemed to allow Datastore tests to pass.

### DIFF
--- a/packages/grpc-protobufjs/src/index.ts
+++ b/packages/grpc-protobufjs/src/index.ts
@@ -97,14 +97,14 @@ function createMethodDefinition(method: Protobuf.Method, serviceName: string, op
     responseSerialize: createSerializer(method.resolvedResponseType as Protobuf.Type),
     responseDeserialize: createDeserializer(method.resolvedResponseType as Protobuf.Type, options),
     // TODO(murgatroid99): Find a better way to handle this
-    originalName: _.camelCase(method.name)
+    originalName: method.name
   };
 }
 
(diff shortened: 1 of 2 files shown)


packages/grpc-protobufjs/src/index.ts
```diff
@@ -97,14 +97,14 @@ function createMethodDefinition(method: Protobuf.Method, serviceName: string, op
...
-    originalName: _.camelCase(method.name)
+    originalName: method.name
   };
 }
 
 function createServiceDefinition(service: Protobuf.Service, name: string, options: Options): ServiceDefinition {
   const def: ServiceDefinition = {};
   for (const method of service.methodsArray) {
-    def[method.name] = createMethodDefinition(method, name, options);
+    def[_.camelCase(method.name)] = createMethodDefinition(method, name, options);
```
murgatroid99: The changes in this file shouldn't make a difference. The whole point of the `originalName` property is that both variants are usable as method names in clients and servers. So I would prefer not to change this.

PR #286, opened 2018-04-19
Update benchmark code to handle both implementations

murgatroid99:


murgatroid99:
I don't want to merge this yet because the pure JS client currently throws an error after the server is shut down. I think it may be incorrectly reconnecting after it should be closed, and it also does not seem to be handling the connection failure properly.

Issue #189, opened 2018-02-21
Unused message properties are being defaulted


respinha-ribeiro:
Hello.

If I send a message via gRPC which doesn't have all of its properties set, they are being defaulted to a null/empty value. I've modified the officially provided [example](https://github.com/grpc/grpc/tree/master/examples/node) to simulate this issue. 
I included some more properties in the `HelloReply` message

```
message HelloReply {
  string message = 1;
  double number_property = 2;
  string string_property = 3;
  repeated string array_property = 4;
  message NestedMsg {
  	string property = 1;
  }
  NestedMsg nested_property = 5;
}
```

and I just added a small print statement to `dynamic_codegen/greeter_client.js`

```
console.log('Total message:', JSON.stringify(response));
```

and it prints:

```
Total message: {"message":"Hello world","number_property":0,"string_property":"","array_property":[],"nested_property":null}
```

So, all properties are being defaulted to 0/""/[]/null. Is this the expected behaviour? I thought it could also be a protobuf.js-specific issue.
In my case, this behaviour is not desirable as in some cases I override existing data in my system with data from protobuf messages and if those properties are defaulted I might even override valid data with a null or empty value. 

If this is the expected workflow for gRPC messages, can it be turned off?

murgatroid99:
This is the expected behavior of protobuf, as documented [here](https://developers.google.com/protocol-buffers/docs/proto3#default). Currently, gRPC does not provide a way to change that, but I am working on a new package (proposed [here](https://github.com/grpc/proposal/blob/master/L23-node-protobufjs-library.md)) that will provide that option among other things.

respinha-ribeiro:
I didn't notice it in the documentation, thanks, @murgatroid99. I'll wait for the new package then.

nicolasnoble:
The new `@grpc/proto-loader` package has been published. Could you tell us if this helps with your problem ?

respinha-ribeiro:
Hey, @nicolasnoble, thanks for the notice.

However, when I try to apply the package definition as in the [example](https://www.npmjs.com/package/@grpc/proto-loader) within Typescript code, I get a type definition error: 
```
error TS2339: Property 'loadPackageDefinition' does not exist on type 'typeof "grpc"'.
```

murgatroid99:
We've published version 1.11.1 which includes those definitions in the TypeScript types file.

respinha-ribeiro:
[...]

PR #306, opened 2018-05-02
Add loadPackageDefinition and interceptor APIs to .d.ts file

murgatroid99:
The issue with `loadPackageDefinition` was pointed out in #189.


nicolasnoble:
I guess we don't really have tests for this yet, due to #220 not being merged in ?

murgatroid99:
We don't have tests for this anyway, because this is type definitions for the old library, not the new one.

PR #311, opened 2018-05-03
V1.11.x merge

murgatroid99:


packages/grpc-js-core/README.md
```diff
@@ -1,5 +1,7 @@
 # Pure JavaScript gRPC Client
 
+**Note: This is an alpha-level release. Some APIs may not yet be present and there may be bugs. Please report any that you encounter**
```
nicolasnoble: That's an interesting change...

packages/grpc-js-core/src/call-stream.ts
```diff
@@ -9,9 +9,10 @@ import {FilterStackFactory} from './filter-stack';
 import {Metadata} from './metadata';
 import {ObjectDuplex} from './object-stream';
 
-const {HTTP2_HEADER_STATUS, HTTP2_HEADER_CONTENT_TYPE, NGHTTP2_CANCEL} = http2.constants;
+const {HTTP2_HEADER_STATUS, HTTP2_HEADER_CONTENT_TYPE, NGHTTP2_CANCEL} =
+    http2.constants;
 
-export type Deadline = Date | number;
+export type Deadline = Date|number;
```
nicolasnoble: We seem to have troubles deciding if we want spaces or not around these :D We really need a linter...

packages/grpc-js-core/README.md
murgatroid99: I added that on v1.11.x before we published. It's just getting into master here.

packages/grpc-js-core/src/call-stream.ts
murgatroid99: It's possible that this is a change by the formatter. Or me accidentally reversing a change by the formatter.

PR #316, opened 2018-05-06
upgrade protobufjs from v5 to v6 for TS Typings.

zixia:
I got the following error when run tslint:

```shell
node_modules/grpc/index.d.ts(2,55): error TS7016: Could not find a declaration file for module 'protobufjs'. '/home/zixia/chatie/grpc/node_modules/protobufjs/dist/protobuf.js' implicitly has an 'any' type.
```

It is because:
1. `protobufjs` v5 has no typing
1. `@types/protobufjs` had been deprecated because `protobufjs` v6 supported typing.

So please upgrade our dependencies if possible. Thanks.

### DIFF
--- a/packages/grpc-native-core/package.json
+++ b/packages/grpc-native-core/package.json
@@ -31,7 +31,7 @@
     "lodash": "^4.15.0",
     "nan": "^2.0.0",
     "node-pre-gyp": "^0.10.0",
-    "protobufjs": "^5.0.0"
+    "protobufjs": "^6.8.6"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


murgatroid99:
We can't just upgrade the dependency. There are a lot of breaking changes in Protobuf.js 6. I would recommend using the `@grpc/proto-loader` package instead of `grpc.load` to get loading via Protobuf.js 6 a different way.

zixia:
Thanks for the reply and explanation. I'll try `@grpc/proto-loader` later.

PR #317, opened 2018-05-06
add dev dependencies

zixia:
all of them is required when runing `npm test`


murgatroid99:
The way we have this set up is that those dependencies are in the `package.json` in the root of the repository. So, if you run `npm install` in the repository root, the test should work.

zixia:
Okey, thanks for telling me that.

PR #322, opened 2018-05-08
session.unref(v9.4.0+ ) replace with socket.unref(v0.3.4+)

leonlau:
session.unref(v9.4.0+ ) replace with socket.unref(v0.3.4+)

see: [https://github.com/nodejs/node/blob/master/lib/internal/http2/core.js#L1263](https://github.com/nodejs/node/blob/master/lib/internal/http2/core.js#L1263)


leonlau:
CLA

murgatroid99:
We understand that we only support Node 9.4+. When `session.unref` is backported to Node 8.x, that is when we will also support Node 8.x.

PR #332, opened 2018-05-10
Add coverage reporting for JavaScript and TypeScript files

murgatroid99:
This also has the test script run the tests in `test/api`, including the interop sanity test.

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -39,7 +39,6 @@ npm install --unsafe-perm
 
 mkdir -p reports
 export JOBS=8
-export JUNIT_REPORT_PATH="reports/node$version/"
 export JUNIT_REPORT_STACK=1
 
 OS=$(uname)
@@ -57,6 +56,8 @@ fi
 
 # TODO(mlumish): Add electron tests
 
+FAILED="false"
+
 for version in ${node_versions}
 do
   # Install and setup node for the version we want.
@@ -65,6 +66,8 @@ do
   nvm install $version
   nvm use $version
   set -ex
+
+  export JUNIT_REPORT_PATH="reports/node$version/"
 
   # https://github.com/mapbox/node-pre-gyp/issues/362
   npm install -g node-gyp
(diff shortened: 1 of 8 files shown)


run-tests.sh
```diff
@@ -39,6 +39,16 @@ npm install --unsafe-perm
 
 mkdir -p reports
 export JOBS=8
+export JUNIT_REPORT_PATH="reports/node$version/"
+export JUNIT_REPORT_STACK=1
+
+gsutil cp gs://grpc-testing-secrets/coveralls_credentials/grpc-node.rc .
```
nicolasnoble: This can actually only work from Linux...

run-tests.sh
```diff
@@ -39,6 +39,16 @@ npm install --unsafe-perm
...
+set +x
+. grpc-node.rc
```
nicolasnoble: Something seems wrong sourcing from the CWD. Maybe let's try /tmp ?

run-tests.sh
murgatroid99: How would you feel about separating the Linux and Mac test scripts?

nicolasnoble: Not sure - it seems to be "pick your poison" to me. I don't like to burden the script with various platform-specific checks, but I also don't like the idea of duplicating code.

Note that we're already doing SOME platform-specific work in some of these scripts (https://github.com/grpc/grpc-node/blob/master/tools/release/kokoro.sh#L39) so maybe adding OS-specific code is a slightly less bitter pill to swallow ?

But I don't have any strong opinion one way or another.

nicolasnoble:
Did TypeScript break us again ?

jtattermusch:
Looks like this has broken   grpc/grpc tests on master.

https://source.cloud.google.com/results/invocations/707c0572-1a9c-495e-aabe-81a92504fd21/targets/github%2Fgrpc/tests
```
+ gsutil cp gs://grpc-testing-secrets/coveralls_credentials/grpc-node.rc /tmp
./run-tests.sh: 48: ./run-tests.sh: gsutil: not found
```

Issue #277, opened 2018-04-15
ReDoS Vulnerability due to protobufjs 


SpiralOutDotEu:
[ReDoS Vulnerability](https://snyk.io/test/npm/protobufjs/5.0.2?severity=high&severity=medium&severity=low) to protobufjs@6.8.5. 
Update to protobufjs@6.8.6 is required. See [relative commit](https://github.com/dcodeIO/protobuf.js/commit/2ee1028d631a328e152d7e09f2a0e0c5c83dc2aa) and [protobufjs 6.8.6 release notes](https://github.com/dcodeIO/protobuf.js/releases/tag/6.8.6)

PatSmuk360:
[`grpc-native-core` still seems to be 5.x](https://github.com/grpc/grpc-node/blob/master/packages/grpc-native-core/package.json#L34)

nicolasnoble:
We can't update to protobufjs6 without doing major damage to our dependents. However, we made a backport of the fix to protobufjs5 (see https://github.com/dcodeIO/protobuf.js/pull/1030), and we are now waiting for @dcodeio to make a protobufjs5 release with this fix in.

murgatroid99:
Reopening until that fix is published.

dcodeIO:
FYI: Thanks for reminding me, just published 5.0.3. Btw:

> We can't update to protobufjs6 without doing major damage to our dependents.

This is also true the other way around I'm afraid. On my end I saw significant friction encountered by users trying to use grpc with latest protobuf.js.

nicolasnoble:
The gRPC 1.12.1 release finally got out, and it is now pinning on 5.0.3 that no longer has that vulnerability - not that it was very important anyway, unless you are doing something really wonky like parsing user-provided .proto files instead of using the json representation, but in all cases, it's all good now.

At the time of writing this, both nodesecurity and skyx still have outdated information on the status of this problem, and we'll keep trying to get them to amend their databases.

nicolasnoble:
As a quick update: both nodesecurity and skyx now have proper information about the security fix that happened in protobufjs 5.0.3.

PR #338, opened 2018-05-15
Update protobufjs => 6.8.6

wyardley:
https://nodesecurity.io/advisories/605

I think this is a bigger issue than the lodash one, since this is a major version difference, and even after removing node_modules / package-lock, still persists. Let me know if this isn't the right one to update or if tests fail.


nicolasnoble:
This is an API breaking change - tests WILL fail. We're waiting on a new version of protobufjs 5.x to be released, since the fix has been backported.

nicolasnoble:
Feel free to follow https://github.com/grpc/grpc-node/issues/277 about this.

PR #340, opened 2018-05-16
Revert "Add coverage reporting for JavaScript and TypeScript files"

jtattermusch:
Reverts grpc/grpc-node#332

Looks like the PR has broken grpc/grpc tests on master.

https://source.cloud.google.com/results/invocations/707c0572-1a9c-495e-aabe-81a92504fd21/targets/github%2Fgrpc/tests

```
gsutil cp gs://grpc-testing-secrets/coveralls_credentials/grpc-node.rc /tmp
./run-tests.sh: 48: ./run-tests.sh: gsutil: not found
```

### DIFF
--- a/gulpfile.ts
+++ b/gulpfile.ts
@@ -102,7 +102,7 @@ gulp.task('test.only', 'Run tests without rebuilding anything',
           ['js.core.test', 'native.test.only']);
 
 gulp.task('test', 'Run all tests', (callback) => {
-  runSequence('build', 'test.only', 'internal.test.test', callback);
+  runSequence('build', 'test.only', callback);
 });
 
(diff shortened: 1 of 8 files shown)


jtattermusch:
CC @matt-kwong

nicolasnoble:
Now that was unexpected.

matt-kwong:
It looks like the issue stems from grpc-node tests being run in a Dockerfile in grpc/grpc and not having `gsutil` installed. I don't think we should have coverage reports when running grpc/grpc-node tests on grpc/grpc, so we should find a way to only the coverage setup in `run-tests.sh` when running on grpc/grpc-node.

PR #343, opened 2018-05-16
Revert "Revert "Add coverage reporting for JavaScript and TypeScript files""

murgatroid99:
Reverts grpc/grpc-node#340. I will fix the condition for running coverage code in the test script as discussed.

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -39,8 +39,24 @@ npm install --unsafe-perm
 
 mkdir -p reports
 export JOBS=8
+export JUNIT_REPORT_STACK=1
+
+OS=$(uname)
+
+if [ "$OS" = "Linux" ]
+then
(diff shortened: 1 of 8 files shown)


run-tests.sh
```diff
@@ -72,7 +79,20 @@ set -ex
 
 node merge_kokoro_logs.js
 
-if [ "$FAILED" != "" ]
+if [ "$FAILED" = "true" ]
 then
   exit 1
+else
+  if [ "$OS" = "Linux" ]
```
nicolasnoble: I realized we might even just remove this condition here, since gsutil was exactly the reason it wouldn't work on mac.

murgatroid99: I will leave it for now because I think it is semantically correct to publish coverage results only once per PR, even if the command were to start working on Mac in the future.

PR #348, opened 2018-05-17
Trying to downgrade the build distribution.

nicolasnoble:


murgatroid99:
I'm forcing a test rerun because I think it's possible that the problem was with the published version of grpc 1.12.0.

Issue #355, opened 2018-05-19
grpc-node tests broken on master in grpc/grpc


jtattermusch:
See https://github.com/grpc/grpc/issues/15470

nicolasnoble:
I think these are all okay now.

PR #356, opened 2018-05-21
Sticking on a more reproducible node version.

nicolasnoble:


nicolasnoble:
Tentatively fixes #355.

jtattermusch:
I tried to test this PR locally and it seems that it's not helping with #355.  

The revert https://github.com/grpc/grpc/pull/15471 does seem to fix the problem though, so let's approve and merge it to get green tests and investigate then.

nicolasnoble:
How are you reproducing locally exactly ?

jtattermusch:
I added locally update `https://github.com/grpc/grpc/blob/master/tools/run_tests/helper_scripts/run_grpc-node.sh`

with 

```
git clone --recursive https://github.com/nicolasnoble/grpc-node ./../grpc-node
cd ./../grpc-node
git checkout node4
```

Issue #365, opened 2018-05-23
grpc 1.12.1 fails to install with `npm ci`


mdouglass:
### Problem description
grpc 1.12.1 fails to install with `npm ci`

### Reproduction steps
I've attached two empty projects that differ only in their dependency on grpc -- one uses 1.11.3, the other 1.12.1.
[repro-grpc.zip](https://github.com/grpc/grpc-node/files/2033207/repro-grpc.zip)

If you run `npm ci` with repro-grpc-1.11.3 it succeeds.
If you run `npm ci` with repro-grpc-1.12.1 it fails with `sh: ./node_modules/.bin/node-pre-gyp: No such file or directory`
 
### Environment
 - macOS 10.13.4
 - node.js v10.1.0 installed via brew
 - npm v6.0.1

#### Successful install
```
bigbird:repro-grpc-1.11.3 mdouglass$ npm ci
npm WARN prepare removing existing node_modules/ before installation

> grpc@1.11.3 install /Users/mdouglass/kixeye/repros/repro-grpc-1.11.3/node_modules/grpc
> ./node_modules/.bin/node-pre-gyp install --fallback-to-build --library=static_library

[grpc] Success: "/Users/mdouglass/kixeye/repros/repro-grpc-1.11.3/node_modules/grpc/src/node/extension_binary/node-v64-darwin-x64-unknown/grpc_node.node" is installed via remote
added 36 packages in 4.435s
bigbird:repro-grpc-1.11.3 mdouglass$ 
```

#### Failed install:
```
bigbird:repro-grpc-1.12.1 mdouglass$ npm ci
npm WARN prepare removing existing node_modules/ before installation

> grpc@1.12.1 install /Users/mdouglass/kixeye/repros/repro-grpc-1.12.1/node_modules/grpc
> ./node_modules/.bin/node-pre-gyp install --fallback-to-build --library=static_library

sh: ./node_modules/.bin/node-pre-gyp: No such file or directory
npm ERR! file sh
npm ERR! code ELIFECYCLE
npm ERR! errno ENOENT
npm ERR! syscall spawn
npm ERR! grpc@1.12.1 install: `./node_modules/.bin/node-pre-gyp install --fallback-to-build --library=static_library`
npm ERR! spawn ENOENT
npm ERR! 
npm ERR! Failed at the grpc@1.12.1 install script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
[...]

PR #367, opened 2018-05-24
Ensuring the presence of node-pre-gyp before publishing.

nicolasnoble:


nicolasnoble:
This should alleviate issues like #365 in the future.

PR #381, opened 2018-05-30
Add compression handling to grpc-js

murgatroid99:
This passes all of the existing tests, but I haven't added new ones. This mirrors the existing library, which also doesn't have substantial compression test coverage. In a future PR I will add the applicable compression interop tests.

### DIFF
--- a/packages/grpc-js-core/src/call-stream.ts
+++ b/packages/grpc-js-core/src/call-stream.ts
@@ -29,6 +29,12 @@ export interface StatusObject {
   metadata: Metadata;
 }
 
+export const enum WriteFlags {
+  BufferHint = 1,
+  NoCompress = 2,
+  WriteThrough = 4
+}
+
 export interface WriteObject {
   message: Buffer;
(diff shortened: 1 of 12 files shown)


packages/grpc-js-core/src/call-stream.ts
```diff
@@ -219,14 +262,13 @@ export class Http2CallStream extends Duplex implements CallStream {
...
-      stream.on('data', (data) => {
+      stream.on('data', (data: Buffer) => {
         let readHead = 0;
-        let canPush = true;
         let toRead: number;
         while (readHead < data.length) {
           switch (this.readState) {
             case ReadState.NO_DATA:
-              this.readCompressFlag = (data.readUInt8(readHead) !== 0);
+              this.readCompressFlag = data.slice(readHead, readHead+1);
```
nicolasnoble: Very minor nit: not sure this is passing linter properly due to the `+1` with no spaces around it.

PR #385, opened 2018-06-05
Fix generic client interceptor resolution

### LINKED ISSUE #363: TypeError: Cannot read property '/service/method' of undefined
### Problem description
I am using a bare-bone `Client` and not a `ServiceClient`. When executing `makeUnaryRequest` (and or similar `makeXXXRequest`), I get the following error:

```
TypeError: Cannot read property '/<service>/<method>' of undefined
      at Client.makeUnaryRequest (/.../node_modules/grpc/src/client.js:457:39)
```
The offending line number varies with the actual version, however the offending line is:

`var method_name = this.$method_names[path];`

It happens to be that `$method_names` is defined in `makeClientConstructor`:

`ServiceClient.prototype.$method_names = {};`

So, `$method_names` is NOT defined for a bare-bone `Client`, which causes the crash.

### Reproduction

`package.json`:
```
{
  "dependencies": {
    "grpc": "^1.12.2"
  }
}
```

`index.js`:
```
const grpc = require("grpc");
const client = new grpc.Client("1.1.1.1", grpc.credentials.createInsecure());
client.makeUnaryRequest(
  "/service/method",
  null,
  null,
  {},
  null,
  null,
  () => {}
);
```

Run with node `index.js`.

Actual result:
```
TypeError: Cannot read property '/service/method' of undefined
    at Client.makeUnaryRequest (.../grpc-363/node_modules/grpc/src/client.js:457:39)
```

Expected result (error from running with grpc v1.10.1).
```
  var message = serialize(argument);
                ^

TypeError: serialize is not a function
    at Client.makeUnaryRequest (.../grpc-363/node_modules/grpc/src/client.js:544:17)
```

### Environment
- Broken in v1.11.X
- Works in v1.9 and v1.10

murgatroid99:
This moves all of the interceptor resolution logic to the call time to avoid needing to do method name lookups. This should fix #363. I also ended up deleting a couple of helper functions in `client_interceptors.js` that were only used for the previous iteration of this functionality. We could probably also remove `Client.prototype.$method_names`, but it's used in one of the tests and I didn't feel like trying to figure out how to change it.

### DIFF
--- a/packages/grpc-native-core/src/client.js
+++ b/packages/grpc-native-core/src/client.js
@@ -380,19 +380,18 @@ function Client(address, credentials, options) {
   options['grpc.primary_user_agent'] += 'grpc-node/' + version;
 
   // Resolve interceptor options and assign interceptors to each method
-  var interceptor_providers = options.interceptor_providers || [];
-  var interceptors = options.interceptors || [];
-  if (interceptor_providers.length && interceptors.length) {
+  if (_.isArray(options.interceptor_providers) && _.isArray(options.interceptors)) {
     throw new client_interceptors.InterceptorConfigurationError(
       'Both interceptors and interceptor_providers were passed as options ' +
       'to the client constructor. Only one of these is allowed.');
   }
+  self.$interceptors = options.interceptors || [];
+  self.$interceptor_providers = options.interceptor_providers || [];
   _.each(self.$method_definitions, function(method_definition, method_name) {
     self[method_name].interceptors = client_interceptors
-      .resolveInterceptorProviders(interceptor_providers, method_definition)
-      .concat(interceptors);
+      .resolveInterceptorProviders(self.$interceptor_providers, method_definition)
+      .concat(self.$interceptors);
   });
-
   // Exclude interceptor options which have already been consumed
   var channel_options = _.omit(options,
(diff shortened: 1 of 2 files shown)


packages/grpc-native-core/src/client.js
```diff
@@ -380,19 +380,18 @@ function Client(address, credentials, options) {
   options['grpc.primary_user_agent'] += 'grpc-node/' + version;
 
   // Resolve interceptor options and assign interceptors to each method
-  var interceptor_providers = options.interceptor_providers || [];
-  var interceptors = options.interceptors || [];
-  if (interceptor_providers.length && interceptors.length) {
+  if (_.isArray(options.interceptor_providers) && _.isArray(options.interceptors)) {
```
nicolasnoble: Something tells me that the original intend of the code was to basically be `_.isArrayLike`, by checking if it had the `length` property.

nicolasnoble: Also, unrelated: shouldn't we be doing something like:

```
if (!_.isArrayLike(options.interceptor_providers) && _.isSet(options.interceptor_providers)) {
  throw new client_interceptors.InterceptorConfigurationError('interceptor_providers needs to behave like an array')
}
```

PR #392, opened 2018-06-11
Add deprecation notice to grpc.load

murgatroid99:
This doesn't actually change any functionality; it just adds a warning that appears once per process when someone uses `grpc.load`.

### DIFF
--- a/packages/grpc-native-core/index.js
+++ b/packages/grpc-native-core/index.js
@@ -20,6 +20,7 @@
 
 var path = require('path');
 var fs = require('fs');
+var util = require('util');
 
 var SSL_ROOTS_PATH = path.resolve(__dirname, 'deps', 'grpc', 'etc', 'roots.pem');
 
@@ -118,7 +119,7 @@ var loadObject = exports.loadObject;
  *     API breakage. It is deprecated, and new code should not use it.
  * @return {Object<string, *>} The resulting gRPC object
  */
-exports.load = function load(filename, format, options) {
+exports.load = util.deprecate(function load(filename, format, options) {
   options = _.defaults(options, common.defaultGrpcOptions);
   options.protobufjsVersion = 5;
   if (!format) {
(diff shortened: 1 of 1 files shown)


kjin:
Will `loadObject` be deprecated as well?

murgatroid99:
Nico and I were discussing that in person. It seems like `loadPackageDefinition` does not entirely replace the functionality of `loadObject`, so for now I have decided not to deprecate it.

alexeevg:
What is the recommended way to encode and decode Protobuf messages now?
We have some code relying on the deprecated grpc.load function and the recommended replacement isn't really working for us.
Consider the following code:
```protobuf
syntax = "proto3";

package my.package;

option java_multiple_files = true;
option java_outer_classname = "Proto";
option java_package = "my package";

message MyMessage {
    map<string, string> data = 1;
}
```
```javascript
const grpc = require('grpc'),
  protoLoader = require('@grpc/proto-loader'),
  {expect} = require('chai');

describe('GRPC loaders', () => {
  describe('deprecated loader', () => {
    const proto = grpc.load('./proto/message.proto')
    it('should encode and decode', () => {
      const MyMessage = proto.my.package.MyMessage;
      const data = {a: 'a', b: 'b'};
      const encoded = new MyMessage({data}).toBuffer();
      const decoded = MyMessage.decode(encoded).data;
      expect(decoded.get('a')).to.equal('a');
      expect(decoded.get('b')).to.equal('b');
    });
  });

  describe('recommended loader', () => {
    const options = {
      keepCase: true,
      longs: String,
      enums: String,
      defaults: true,
      oneofs: true
    };
    const proto = grpc.loadPackageDefinition(protoLoader.loadSync('./proto/message.proto', options));
    it('should provide to message encoder/decoder', () => {
      const MyMessage = proto.my.package.MyMessage; // fails because proto.my is undefined
      // use MyMessage to encode/decode
    });
  });
});
```

nicolasnoble:
You should use protobufjs directly then. The loader's intention is to provide grpc with the service definition, and the function of grpc (to send and receive RPCs) isn't dependent on the ability of the user to encode or decode protobufs, as this is done by the library.

PR #395, opened 2018-06-13
Changes needed for the nanopb dep refactoring work

dgquintas:
In particular, https://github.com/grpc/grpc/pull/15595


nicolasnoble:
So, yes. The changes aren't mutually exclusive here, so we could do it this way, indeed.

dgquintas:
PTAL!

PR #403, opened 2018-06-19
Add checkServerIdentity callback

JackOfMostTrades:
Core [PR 15274](https://github.com/grpc/grpc/pull/15274) added an option to provide a callback during peer certificate verification in order to perform additional validation. This PR exposes that option in node as a "checkServerIdentity" function that has the same semantics as the option of the same name available in the [node TLS library](https://nodejs.org/api/tls.html#tls_tls_checkserveridentity_host_cert) (except that it does not replace regular hostname; it is in addition to it).

Example usage:

```
grpc.credentials.createSsl(
    fs.readFileSync("truststore.pem"),
    fs.readFileSync("client.pem.key"),
    fs.readFileSync("clent.pem.crt"),
    {
      "checkServerIdentity": function(host, cert) {
        if (fingerprint(cert) !== expectedFingerprint) {
          throw "Invalid server certificate presented";
        }
      }
    });
```

### DIFF
--- a/packages/grpc-native-core/ext/channel_credentials.cc
+++ b/packages/grpc-native-core/ext/channel_credentials.cc
@@ -78,9 +78,15 @@ static int verify_peer_callback_wrapper(const char* servername, const char* cert
     argv[1] = Nan::New<v8::String>(cert).ToLocalChecked();
   }
 
-  callback->Call(argc, argv);
+  Local<Value> result = callback->Call(argc, argv);
 
+  // Catch any exception and return with a distinct status code which indicates this
   if (try_catch.HasCaught()) {
+    return 2;
+  }
+
+  // If the result is an error, return a failure
+  if (result->IsNativeError()) {
     return 1;
   }
 
--- a/packages/grpc-native-core/ext/channel_credentials.cc
+++ b/packages/grpc-native-core/ext/channel_credentials.cc
@@ -186,35 +186,21 @@ NAN_METHOD(ChannelCredentials::CreateSsl) {
   verify_peer_options verify_options = {NULL, NULL, NULL};
   if (!info[3]->IsUndefined()) {
     if (!info[3]->IsObject()) {
-      return Nan::ThrowTypeError("createSsl's fourth argument must be an "
-          "object");
+      return Nan::ThrowTypeError("createSsl's fourth argument must be an object");
     }
-    Local<Context> context = Nan::New<Context>();
     Local<Object> object = info[3]->ToObject();
-    MaybeLocal<Array> maybe_props = object->GetOwnPropertyNames(context);
-    if (!maybe_props.IsEmpty()) {
-      Local<Array> props = maybe_props.ToLocalChecked();
-      for(uint32_t i=0; i < props->Length(); i++) {
-        Local<Value> key = props->Get(i);
-        Local<Value> value = object->Get(key);
 
-        if (key->IsString()) {
-          Nan::Utf8String keyStr(key->ToString());
-
-          if (strcmp("checkServerIdentity", (const char*)(*keyStr)) == 0) {
-
-            if (!value->IsFunction()) {
-                return Nan::ThrowError("Value of checkServerIdentity must be a function.");
-            }
-            Nan::Callback *callback = new Callback(Local<Function>::Cast(value));
-            verify_options.verify_peer_callback = verify_peer_callback_wrapper;
-            verify_options.verify_peer_callback_userdata = (void*)callback;
-            verify_options.verify_peer_destruct = verify_peer_callback_destruct;
-
-          }
-        }
-        // do stuff with key / value
+    Local<Value> checkServerIdentityValue = Nan::Get(object,
+        Nan::New("checkServerIdentity").ToLocalChecked()).ToLocalChecked();
+    if (!checkServerIdentityValue->IsUndefined()) {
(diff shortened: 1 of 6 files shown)


murgatroid99:
Why did you make the new argument an object instead of just a function?

packages/grpc-native-core/ext/channel_credentials.cc
```diff
@@ -58,6 +60,43 @@ ChannelCredentials::~ChannelCredentials() {
   grpc_channel_credentials_release(wrapped_credentials);
 }
 
+struct callback_md {
+    Nan::Callback *callback;
+};
```
murgatroid99: Why did you use a struct here, instead of using the `callback` pointer directly?

packages/grpc-native-core/ext/channel_credentials.cc
```diff
@@ -148,9 +187,42 @@ NAN_METHOD(ChannelCredentials::CreateSsl) {
         "createSsl's second and third arguments must be"
         " provided or omitted together");
   }
+
+  verify_peer_options verify_options = {NULL, NULL, NULL};
+  if (info.Length() >= 4 && info[3]->IsObject()) {
```
murgatroid99: I don't think the length check is necessary here. As far as I can tell, indexing past the end returns undefined. In addition, I would like to also have an error thrown if the argument is provided but not the correct type, like with the other arguments.

nicolasnoble:
Also, you need to regenerate the build files by running the `packages/grpc-native-core/tools/buildgen/generate_projects.sh` script.

JackOfMostTrades:
> Why did you make the new argument an object instead of just a function?

> Why did you use a struct here, instead of using the callback pointer directly?

The answer to both is because in several of my first iterations adding additional verification options, there were several other options (cert fingerprint assertions, skipping hostname verification, etc). All that's been removed now and there's just the one callback option, so it totally makes sense to remove the struct being passed as userdata and I've done that. As for whether the fourth argument should be an object or just the callback function, I can simplify that if you want (but if you think it may be useful to keep the API as it is to allow for additional options, it could be useful left as-is?).

murgatroid99:
I'm fine with leaving the argument as an object. I wish the API overall was more consistent, but having the user pass a function directly might be more confusing.

Also, I am pretty sure you can simplify the logic of handling the object. I think that if the key is not present, `Nan::Get` will just return undefined, so you should be able to just try to get the specific key without iterating over the object.

JackOfMostTrades:
> Also, I am pretty sure you can simplify the logic of handling the object.

Yep you're right. Thanks for the pointer. I also added a few tests to check for the expected type assertions of that argument.

murgatroid99:
OK, one final thing: can you update the documentation in `src/credentials.js` and `index.d.ts` to include these API changes?

JackOfMostTrades:
Sure thing, `src/credentials.js` and `index.d.ts` updated (which I think I've done correctly but you should take a close look at; I'm not fully up to speed on typescript yet).

murgatroid99:
The typescript type looks correct to me. In the jsdoc comment, you can actually do `@param {Function} verify_options.checkServerIdentity` instead of writing in prose that it is an accepted option.

However, looking at that documentation made me realize that I am not really comfortable with a callback that is supposed to throw an error. Nothing else in our API does that, and the equivalent callback in the Node TLS library returns an error on failure, or `undefined` on success. It looks like there isn't a good way to make it asynchronous, so I would prefer to at least make it consistent with the Node TLS library. The `cert` parameter will be different either way, but I don't know if we can do anything about that.

JackOfMostTrades:
> the equivalent callback in the Node TLS library returns an error on failure, or undefined on success.

Good catch! For some reason I read it as throwing an error on failure instead of returning it, so that's what I was designing against. I've corrected the implementation and documentation to expect the same behavior: a returned error on failure or otherwise undefined.

JackOfMostTrades:
Yep, I think you're right that I should update the cert parameter to a parsed version. If we're doing the parsing in JS (e.g. with PKIjs) do you have a hint for the best way to call back into a specific JS function (instead of a passed-in callback) from C++? I feel like it would be cleanest if the C++ `verify_peer_callback_wrapper` could call into a named JS callback wrapper function that parsed the cert and ultimately invoked the user callback.

In lieu of that, my first pass working on this was to change `exports.createSsl` in `credentials.js` to a function that looks for the presence of a `verify_options.checkServerIdentity` argument and wraps it with the aforementioned callback wrapper on the way in.

murgatroid99:
After some further consideration and internal discussion, I suggest that you hold off on further changes to this PR. We are not sure what is the best way to present the certificates to the user and we are planning to create a gRFC (https://github.com/grpc/proposal) to figure out how to proceed here.

JackOfMostTrades:
Cool, I'll keep an eye out for that and jump in on the discussion when it pops up. Thanks for the update!

murgatroid99:
I'm not really sure why the tests are failing now. They seem to be failing to load `node-forge`, even though I can see that it was added as a dev dependency.

murgatroid99:
Actually, I know what the problem is. @JackOfMostTrades please edit `packages/grpc-native-core/templates/package.json.template` with the same changes that were made to `packages/grpc-native-core/package.json`.

We really should add a test for that.

murgatroid99:
Never mind, I made the change myself, and it looks like that did fix the problem.

Issue #407, opened 2018-06-21
Add RequestType and ResponseType Metadata


Jmoore1127:
### Is your feature request related to a problem? Please describe.
I am attempting an upgrade of some tooling that was previously using `grpc.load(...)` to use the new `proto-loader`. The parsed proto the is returned is now much simpler and easier to use in many ways. However, the `RequestType` and `ResponseType` information is no longer available in `proto-loader`. The tooling I am working with uses this information in order to produce sample json objects for the user. 

### Describe the solution you'd like
I would like to add this metadata to the method definition returned from `createMethodDefinition` in `grpc-protobufjs/src/index.ts` so that the tooling has access to this information. 

### Additional context
I have submitted a PR with a basic implementation.

PR #378, opened 2018-05-28
Add request/response type to method definition

kurayama:
Adds the resolved request/response type to `grpc-loader` method definition.

PR #408, opened 2018-06-21
Add RequestType and ResponseType metadata to method definition

Jmoore1127:
PR for issue #407

### DIFF
--- a/packages/grpc-protobufjs/src/index.ts
+++ b/packages/grpc-protobufjs/src/index.ts
@@ -37,6 +37,8 @@ export interface MethodDefinition<RequestType, ResponseType> {
   requestDeserialize: Deserialize<RequestType>;
   responseDeserialize: Deserialize<ResponseType>;
   originalName?: string;
+  requestType: RequestType;
+  responseType: ResponseType;
 }
 
 export interface ServiceDefinition {
@@ -97,7 +99,9 @@ function createMethodDefinition(method: Protobuf.Method, serviceName: string, op
     responseSerialize: createSerializer(method.resolvedResponseType as Protobuf.Type),
     responseDeserialize: createDeserializer(method.resolvedResponseType as Protobuf.Type, options),
     // TODO(murgatroid99): Find a better way to handle this
-    originalName: _.camelCase(method.name)
+    originalName: _.camelCase(method.name),
+    requestType: method.resolvedRequestType,
+    responseType: method.resolvedResponseType,
   };
 }
 


thelinuxfoundation:

* User @Jmoore1127 isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

Jmoore1127:
CLA Signed.

nicolasnoble:
Different take on #378.

packages/grpc-protobufjs/src/index.ts
```diff
@@ -37,6 +37,8 @@ export interface MethodDefinition<RequestType, ResponseType> {
   requestDeserialize: Deserialize<RequestType>;
   responseDeserialize: Deserialize<ResponseType>;
   originalName?: string;
+  requestType: RequestType;
+  responseType: ResponseType;
```
murgatroid99: These type declarations aren't really semantically correct. `RequestType` and `ResponseType` are the types of objects that are passed to and received from the gRPC methods. In the case of this library, those are plain objects with a structure matching the declared proto message types, which is why the `ServiceDefinition` just declares them with `object, object`.

But `requestType` and `responseType` are not that. They are Protobuf.js message descriptor objects that describe the structure of `RequestType` and `ResponseType`. @nicolasnoble linked another PR that tried to do this and they declared those values as `Protobuf.Type`, which I think is the actual correct type.

murgatroid99:
I want to mention that I have the same concerns I had in #378, but I am willing to reconsider.

Jmoore1127:
I am happy to change types when I am back at my desk tomorrow. 

I am not picky about how the information is exposed, I just need some way to access it. I can update our tooling in order to parse the structure. My main problem is that the information is captured within closures for the serializers and deserializers so I have no way of accessing it without parsing the file twice (once with grpc and once with protobufjs).

Jmoore1127:
@murgatroid99 I have updated the typings to match actual types as suggested. I think I better understand your comment about not exposing protobufjs apis now. Do you have any suggestions as to what you would accept for the requestType and responseType objects? At a minimum, I need information such as field names, field types, and structure (for nested types) in order to create sample request and response objects.

murgatroid99:
One option, to make it as general as possible, would be to use the JSON representation of the `DescriptorProto` (from `google/protobuf/descriptor.proto`) objects that correspond to the two message types. Unfortunately, it's pretty hard to actually get that representation out of Protobuf.js.

murgatroid99:
I am sorry for the long delay here. If you are still interested in this problem, I have proposed a solution at https://github.com/grpc/proposal/pull/116. Please feel free to comment on it.

PR #411, opened 2018-06-22
auth_context support.

talrasha007:
This pull request can solve this [issue](https://github.com/grpc/grpc/issues/7834#issuecomment-288785780), Call object can now call getAuthContext to get a AuthContext, then we can read the properties in AuthContext.

## Usage
```js
server.addService(proto.TestService.service, {
  unary: function (call, cb) {
    const authContext = call.getAuthContext();
    if (authContext) {
      const properties = authContext.getProperties();
      console.log(properties);

      const pem = authContext.getProperty('x509_pem_cert');
      pem && console.log(pem.toString());
    }

    call.sendMetadata(call.metadata);
    cb(null, {});
  }
});
```

### DIFF
--- a/packages/grpc-native-core/ext/auth_context.cc
+++ b/packages/grpc-native-core/ext/auth_context.cc
@@ -1,0 +1,141 @@
+/*
+ *
+ * Copyright 2015 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#include "auth_context.h"
+
+namespace grpc {
+namespace node {
+
+using Nan::Callback;
+using Nan::EscapableHandleScope;
+using Nan::HandleScope;
+using Nan::Maybe;
+using Nan::MaybeLocal;
+using Nan::ObjectWrap;
+using Nan::Persistent;
+using Nan::Utf8String;
+
+using v8::Array;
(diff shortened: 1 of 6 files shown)


thelinuxfoundation:

* User @talrasha007 isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

talrasha007:
Well, I signed the Individual Contributor License Agreement. :)

murgatroid99:
I have a few concerns here. First, we have to consider how this API might be ported to the pure JavaScript library, so we should try to make sure that the functionality can be implemented there. In particular for this change, I would want to restrict which parts of the auth context can be returned to the ones that we can guarantee are accessible in the other implementation, with the option to expand it later.

Second, the auth context appears to be just a map, semantically, so I would prefer to return an actual map instead of just mirroring the core api. Alternatively, it could be a class with more human readable getters like `getPeerCertificate`.

This brings up my third point: the Node TLS library has a [`getPeerCertificate`](https://nodejs.org/api/tls.html#tls_class_tls_tlssocket) method and as I also commented in #403 I think it would be valuable to match the API as much as possible, which would require parsing the certificate, possibly using the TSIjs library. In both cases that would probably require an additional bit of wrapping code in JavaScript to perform that parsing.

talrasha007:
@murgatroid99, I agree with you, getPeerCertificate would be nice. I didn't implement it because i want to keep changes simple enough, and i don't know which parser do you prefer.
:)

murgatroid99:
I understand wanting to keep the change simple, but a new API lasts a long time, so I want to get it right the first time. And my suggestion for a parser is [PKIjs](https://www.npmjs.com/package/pkijs) (that I accidentally referred to as TSIjs in my previous comment), which is the only x509 parser library I've found that doesn't rely on its own native extension.

murgatroid99:
After some further consideration and internal discussion, I suggest that you hold off on further changes to this PR. We are not sure what is the best way to present the certificates to the user and we are planning to create a gRFC (https://github.com/grpc/proposal) to figure out how to proceed here.

talrasha007:
ok, understand.

murgatroid99:
With the deprecation of the `grpc` package, we are not going to move forward with this.

PR #416, opened 2018-06-25
grpc-js Add optional private key passphrase for mutual TLS scenarios

### LINKED ISSUE #413: grpc-js support for passphrase-protected private keys on mutual TLS connections
The [`createSSL` method](https://github.com/grpc/grpc-node/blob/f952778ed75b3a000adc6abc787d266af88c9a82/packages/grpc-js-core/src/channel-credentials.ts#L53) assumes that the private key you are providing in a mutual TLS scenario is not protected with a passphrase. 

If you are using a private key protected with a passphrase, you will get an error like this: 
```
Error: error:0906A068:PEM routines:PEM_do_header:bad password read
```

It is interesting to note that while it is possible to address this issue in `grpc-js-core` simply by passing an additional `passphrase` option on the [`createSecureContext` call](https://github.com/grpc/grpc-node/blob/f952778ed75b3a000adc6abc787d266af88c9a82/packages/grpc-js-core/src/channel-credentials.ts#L67), it does not appear to be possible to address this in `grpc-native-core` based on the current lack of support in the underlying library based on [this issue](https://github.com/grpc/grpc/issues/14216).

marzolfb:
Fixes #413 

I didn't see any tests in the api-level/interop tests that addressed a mutual TLS scenario, so I didn't add anything there.

I'm interested to see if this PR is feasible given that there doesn't appear to be support for passphrases in grpc core (see my comment in #413) and thus the grpc-native-core variant wouldn't be able to implement this same optional addition right now.


nicolasnoble:
The fact we wouldn't have a counterpart for the native extension is indeed an issue.

murgatroid99:
This is superseded by #1988.

PR #446, opened 2018-07-19
Add Channel class and Client channel override options to public API

### LINKED ISSUE #386: Add extension point for a channel creation factory
This will require us to do the following:

- [ ] Modify the C++ wrapping code to create `Call` objects using a method on `Channel` instead of a `Call` constructor.
- [ ] Expose the `Channel` class in the library API. *Note that this will also impact the grpc-js library, which needs to expose the same API*
 - [ ] Add a `channelFactoryOverride` option to the `Client` constructor.

murgatroid99:
This is the implementation of [gRFC proposal L32](https://github.com/grpc/proposal/blob/master/L32-node-channel-API.md), and should fully resolve #386. I only have one test so far because I'm not sure what other tests I can usefully add for this API.

### DIFF
--- a/packages/grpc-native-core/ext/channel.cc
+++ b/packages/grpc-native-core/ext/channel.cc
@@ -28,6 +28,7 @@
 #include "completion_queue.h"
 #include "grpc/grpc.h"
 #include "grpc/grpc_security.h"
+#include "slice.h"
 #include "timeval.h"
 
 namespace grpc {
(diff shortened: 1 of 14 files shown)


packages/grpc-native-core/ext/channel.cc
```diff
@@ -88,10 +110,31 @@ bool ParseChannelArgs(Local<Value> args_val,
     } else if (value->IsString()) {
       Utf8String val_str(value);
       channel_args->args[i].type = GRPC_ARG_STRING;
-      channel_args->args[i].value.string =
-          reinterpret_cast<char *>(calloc(val_str.length() + 1, sizeof(char)));
-      memcpy(channel_args->args[i].value.string, *val_str,
-             val_str.length() + 1);
+      /* Append the grpc-node user agent string after the application user agent
```
nicolasnoble: Maybe we want to add some tests for this codepath ? Like, what we should expect as a user agent string when we give no user agent, and when we give one.

murgatroid99: We do have [this test](https://github.com/grpc/grpc-node/blob/master/packages/grpc-native-core/test/surface_test.js#L487). I could add another one to check the case of passing the option.

PR #448, opened 2018-07-20
Restore Protobuf.Types to GrpcObject tree

dvisztempacct:
`@grpc/proto-loader` hides protobufjs' message types from users, making them inaccessible.

This PR suggests a possible approach for making them accessible again.

I'm given to understand that because protobufjs is non-semver and `grpc-node` would like to remain semver, this has not be done.

I wanted to create this PR as a place to document the details of this decision and explore the possibility of finding some way to do this. I am doubtful this PR will be merged, but hopefully it will accomplish its other goals :)

### DIFF
--- a/packages/grpc-protobufjs/package.json
+++ b/packages/grpc-protobufjs/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@TIN/proto-loader",
+  "name": "grpc-alt-proto-loader",
   "version": "0.4.0",
   "author": "Google Inc.",
   "contributors": [
--- a/packages/grpc-protobufjs/package.json
+++ b/packages/grpc-protobufjs/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@grpc/proto-loader",
+  "name": "@TIN/proto-loader",
   "version": "0.3.0",
   "author": "Google Inc.",
   "contributors": [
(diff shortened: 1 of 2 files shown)


thelinuxfoundation:

* User @dvisztempacct isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

nicolasnoble:
I have a semi-crazy proposition, that I'd need to discuss with @murgatroid99 anyway. But the main reason we decided not to do this was that we didn't want to transitively expose protobufjs' API through, because we can't really control protobufjs' API changes.

How about we add an optional field that is protobufjs directly ? More specifically, something along these lines:

```
const protobufjs = require('protobufjs')
const protoLoader = require('@grpc/proto-loader')
const loaderOptions = {
  defaults: true,
  protobufjs: protobufjs
}
const servicesAndMessages = protoLoader.loadSync('some.proto', options)
```

With this proposal, when `protobufjs` is specified, the loader will use it instead of its own protobufjs dependency, ensuring that the user will know exactly which version of protobufjs is used to expose messages. And when it's specified, `proto-loader` will then no longer suppress the message types from the results, because this isn't a transitive API exposition anymore.

dvisztempacct:
@nicolasnoble that sounds like it would work!

murgatroid99:
I am sorry for the long delay here. If you are still interested in this problem, I have proposed a solution at https://github.com/grpc/proposal/pull/116. Please feel free to comment on it.

PR #454, opened 2018-07-23
Stop using lodash's template generator.

### LINKED ISSUE #453: GRPC Prevents Content Security Policy from working without including 'unsafe-eval'
**In GRPC-node 1.11.3:**

```
var log_template = _.template(
    '{severity} {timestamp}\t{file}:{line}]\t{message}',
    {interpolate: /{([\s\S]+?)}/g});
```

The use of `lodash`'s `_.template` function requires `eval` in a production environment. It seems kind of needless that this function is used here for something as mundane as logging, given the security implications. Is it avoidable?

nicolasnoble:
Fixes #453.

### DIFF
--- a/packages/grpc-native-core/index.js
+++ b/packages/grpc-native-core/index.js
@@ -175,9 +175,14 @@ exports.loadPackageDefinition = function loadPackageDefintion(packageDef) {
   return result;
 };
 
-var log_template = _.template(
-    '{severity} {timestamp}\t{file}:{line}]\t{message}',
-    {interpolate: /{([\s\S]+?)}/g});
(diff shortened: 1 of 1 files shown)


murgatroid99:
Add semicolons please, including after the `}`

nicolasnoble:
Ah, right.

Slapbox:
@nicolasnoble that was quick!

nicolasnoble:
Thanks :-) Granted, this was a really easy fix.

Also, for the follow up: we're backporting this fix to our 1.13 branch, and
we are getting firebase to update their grpc dependency.

PR #468, opened 2018-07-31
Pure js: add simple keepalive logic

murgatroid99:
This uses a very simple algorithm: as long as there is at least one active stream, send a ping every `keepaliveTimeMs` milliseconds. If any single ping does not receive a response within `keepaliveTimeoutMs` milliseconds, assume that the subchannel has disconnected and reconnect. This is a bit more resource intensive than the suggested algorithm, but it should work well enough.

This adds support for the channel arguments `grpc.keepalive_time_ms` and `grpc.keepalive_timeout_ms`.


murgatroid99:
Even if that fixes the problem, I think there is a potential race here, where a ping times out, so the subchannel is considered disconnected, then the channel connects with a new subchannel, then the original subchannel's call finishes and it calls `stopKeepalivePings`, which stops pings on the new subchannel.

And I think the best solution might be to do all of this logic per subchannel.

packages/grpc-js-core/src/channel.ts
```diff
@@ -42,6 +43,8 @@ export interface ChannelOptions {
   'grpc.primary_user_agent': string;
   'grpc.secondary_user_agent': string;
   'grpc.default_authority': string;
+  'grpc.keealive_time_ms': number;
```
kjin: Nit: `keepalive_time_ms`

PR #474, opened 2018-08-06
Fix stream.write type incompatibility

murgatroid99:
Backport #473 to 1.14 to make sure the tests stay green.

There's a chance that this actually breaks the tests. It is possible that the original breakage was caused by upgrading `@types/node` on master, in which case this change isn't necessary on the branch.


nicolasnoble:
Tests are indeed failing.

murgatroid99:
OK, I guess we actually don't need this change on this branch.

PR #482, opened 2018-08-07
Package comparison documentation

murgatroid99:
Make sure to look at the rich diff. It looks a lot better than the markdown text does.

I took the feature list directly from our internal document (and omitted a couple of features that neither library has). I also guessed at a couple of `grpc` features that I think are implemented in the C Core, but I'm not entirely sure. The new link in the README links to the new document.

I'm unsure about the wording of the "Binary Addon" line. I think the green and red makes it look like a positive on the `grpc` side, which is not what we want to communicate, but I can't think of a good way to word it to switch them. "Just an npm package without a binary addon" seems convoluted and contrived.

### DIFF
--- a/PACKAGE-COMPARISON.md
+++ b/PACKAGE-COMPARISON.md
@@ -1,0 +1,29 @@
+# Feature comparison of `grpc` and `@grpc/grpc-js` packages
+
+Feature | `grpc` | `@grpc/grpc-js`
+--------|--------|----------
+Client | :heavy_check_mark: | :heavy_check_mark:
+Server | :heavy_check_mark: | :x:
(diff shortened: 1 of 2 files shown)


nicolasnoble:
"pure javascript code" maybe? Then you can switch the colors.

Also maybe add a * next to the supported platforms to have an undertext saying that while the code may work outside this restricted set, we won't prioritize issues reported for them, and it requires a compiler suite.

Also "arm7+", in order to exclude arm6.

murgatroid99:
I made the other changes, but I don't think this is the right place to be describing our platform support policies.

murgatroid99:
I'm just going to merge this as it is now. We can always make more edits later.

PR #484, opened 2018-08-09
fix segmentation fault

theltybc:
fix segmentation fault, if close channel before finely Client.waitForReady, or call any method of $channel.


murgatroid99:
This is not the right way to fix a segfault. The C++ addon should never segfault, so fixing it should involve some changes to the C++ code. To start with, can you file an issue describing the segfault you are experiencing.

thelinuxfoundation:

* User @theltybc isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

theltybc:
Thank you.
Create issue: https://github.com/grpc/grpc-node/issues/490

PR #493, opened 2018-08-13
Add checkServerIdentity callback to pure js library

murgatroid99:
Basically, port #403 to the pure JavaScript library. I had to change a couple of the internal parts of the API.

CC @JackOfMostTrades

### DIFF
--- a/packages/grpc-js-core/src/channel.ts
+++ b/packages/grpc-js-core/src/channel.ts
@@ -213,9 +213,9 @@ export class Http2Channel extends EventEmitter implements Channel {
       private readonly options: Partial<ChannelOptions>) {
     super();
     if (credentials.isSecure()) {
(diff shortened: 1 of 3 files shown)


packages/grpc-js-core/src/channel.ts
```diff
@@ -214,7 +212,7 @@ export class Http2Channel extends EventEmitter implements Channel {
       address: string, readonly credentials: ChannelCredentials,
       private readonly options: Partial<ChannelOptions>) {
     super();
-    if (credentials.getSecureContext() === null) {
+    if (credentials.isSecure()) {
```
JackOfMostTrades: Is this boolean check backwards?

murgatroid99: Yes. I switched the outcomes to avoid a negative if check.

Issue #472, opened 2018-08-06
I am facing gyp error while installing grpc package


garimak29:
### Problem description
While installing grpc package, I am receiving following error:

```
gyp ERR! build error
gyp ERR! stack Error: `C:\Program Files (x86)\MSBuild\14.0\bin\msbuild.exe` failed with exit code: 1
gyp ERR! stack     at ChildProcess.onExit (d:\Profiles\garkumar\AppData\Roaming\npm\node_modules\npm\node_modules\node-gyp\lib\build.js:258:23)
gyp ERR! stack     at emitTwo (events.js:126:13)
gyp ERR! stack     at ChildProcess.emit (events.js:214:7)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:198:12)
gyp ERR! System Windows_NT 10.0.15063
gyp ERR! command "C:\\Users\\garkumar\\Node\\node.exe" "d:\\Profiles\\garkumar\\AppData\\Roaming\\npm\\node_modules\\npm\\node_modules\\node-gyp\\bin\\node-gyp.js" "build" "--fallback-to-build" "--library=static_library" "--module=d:\\Profiles\\garkumar\\AppData\\Roaming\\npm\\node_modules\\grpc\\src\\node\\extension_binary\\node-v57-win32-x64-unknown\\grpc_node.node" "--module_name=grpc_node" "--module_path=d:\\Profiles\\garkumar\\AppData\\Roaming\\npm\\node_modules\\grpc\\src\\node\\extension_binary\\node-v57-win32-x64-unknown" "--napi_version=1" "--node_abi_napi=napi"
gyp ERR! cwd d:\Profiles\garkumar\AppData\Roaming\npm\node_modules\grpc
gyp ERR! node -v v8.9.4
gyp ERR! node-gyp -v v3.6.2
gyp ERR! not ok
node-pre-gyp ERR! build error
[...]

PR #501, opened 2018-08-16
Fix some file includes in package.json

### LINKED ISSUE #471: NPM module includes very large, unnecessary file
### Problem description
The gRPC node module is 36.3MB or so. That appears largely, in part due to a single 20.2MB file from the BoringSSL dep that seems to be a totally unnecessary test data source file.  

### Reproduction steps
Install the node module

### Environment
Mac OS X. NPM 5.5.1. Node 8.9.1

### Additional context
The file in question is "deps/grpc/src/boringssl/crypto_test_data.cc". 

Is this something I should open up against BoringSSL? Since it's the gRPC lib that is packaging it I figured this might be the place, but let me know if I'm barking up the wrong tree.

murgatroid99:
This fixes #471, and also the second error reported in #472.


murgatroid99:
I know that file path is problematic. I already had that conversation with the team and it's getting fixed. See grpc/grpc#16374.

PR #505, opened 2018-08-17
V1.14 upmerge

murgatroid99:
We have a bunch of commits that need to go into master.


packages/grpc-native-core/build.yaml
```diff
@@ -1,2 +1,3 @@
 settings:
   '#': It's possible to have node_version here as a key to override the core's version.
+  node_version: 1.14.1
```
nicolasnoble: Get rid of this one.

PR #508, opened 2018-08-21
Adding getAuthContext call.

nicolasnoble:
Pending proposal write up and approval.


nicolasnoble:
Supersedes #411.

nicolasnoble:
Fixes https://github.com/grpc/grpc/issues/7834.

nicolasnoble:
Proposal up at https://github.com/grpc/proposal/pull/94.

packages/grpc-native-core/index.d.ts
```diff
@@ -294,6 +294,8 @@ declare module "grpc" {
     handleServerStreamingCall<RequestType, ResponseType> |
     handleBidiStreamingCall<RequestType, ResponseType>;
 
+  type authContext = object | null;
```
murgatroid99: I don't think this is a good type for this. I don't see why it should ever be `null`, and we can be a lot more specific about the fields. It should probably be something like

```ts
interface authContext = {
  transportSecurityType: string;
  sslPeerCertificate: Certificate;
}
```

My understanding is that they will always be populated. Otherwise it should be `?:` for optional fields.

packages/grpc-native-core/src/server.js
```diff
@@ -509,6 +509,46 @@ ServerReadableStream.prototype.getPeer = getPeer;
...
+    if (cert instanceof Error) throw cert;
+    return { name: 'sslPeerCertificate', value: { raw: cert } };
+  }
+}
+
+function getAuthContext() {
+  /* jshint validthis: true */
+  var authContext = this.call.getAuthContext();
+  return _.reduce(authContext, function(obj, value, key) {
+    key = _.camelCase(key);
```
murgatroid99: This doesn't really seem necessary. The `authContextMapping` can just use the original names as keys.

packages/grpc-native-core/src/common.js
```diff
@@ -112,6 +112,35 @@ exports.getMethodType = function(method_definition) {
...
+  // https://nodejs.org/api/tls.html#tls_tlssocket_getpeercertificate_detailed
+
+  var pemHeaderIndex = cert.indexOf(PEM_CERT_HEADER);
+  if (pemHeaderIndex === -1) {
+    return new Error("Unable to parse certificate PEM.");
+  }
+  cert = cert.substring(pemHeaderIndex);
+  var pemFooterIndex = cert.indexOf(PEM_CERT_FOOTER);
+  if (pemFooterIndex === -1) {
+    return new Error("Unable to parse certificate PEM.");
```
murgatroid99: Why return an error here, instead of just throwing?

packages/grpc-native-core/index.d.ts
nicolasnoble: That's fair about the "object" type. I didn't necessarily try too hard defining it.

As for the "null" portion, this goes with what the current implementation does in C++. I may have been a bit over protective, but I'm returning `null` if there's a null auth context from the core, but I'm not sure of what that could imply, to be fair.

packages/grpc-native-core/src/common.js
nicolasnoble: This is not new code :-) You're reviewing me moving code from #403, so you're arguing about code written by @JackOfMostTrades here :-D

packages/grpc-native-core/src/server.js
nicolasnoble: Myeah, this is a remnant of what I did previously where THIS was the sole transformation. Still, this kind of has the nice property of letting us have a `getAuthContextMapping` that has camelCase as keys as with the rest of the surrounding code, so it makes my eyes twitch a bit less when reading it, but if you feel strongly about it I'll remove it.

packages/grpc-native-core/src/common.js
murgatroid99: This doesn't even do the same thing as the code you copied, because adding another layer of functions changes what `return` means.

packages/grpc-native-core/index.d.ts
murgatroid99: I don't like having too many layers of "maybe this exists, maybe it doesn't". I would prefer to make both properties optional, and have an empty object if `null` comes up from the core. That being said, does a missing `transportSecurityType` even have meaningful semantics?

packages/grpc-native-core/src/common.js
nicolasnoble: Pff, you're correct. My mistake. Let me fix this.

packages/grpc-native-core/index.d.ts
nicolasnoble: So, yeah, we could be strengthening this a bit more, I don't disagree. My issue is that the core's auth context isn't really... well... defined in terms of what things mean semantically. So on one hand I'm with you on providing a stronger upper layer by reinterpreting what the core means, but on the other, I'm wary of unknown semantic meanings that may even change in the future in a way that might surprise us in an unfortunate manner, so it's kind of a coin toss for me.

murgatroid99: I guess part of the question is what we are communicating to the users. The way I see it, an absent property in the object we return means "we don't have this information", which is true if it's absent from the auth context from core, or if the auth context from core is null. Do we have good reason to believe that our users will ever get any useful information from getting `null` vs `{}`? I am OK with omitting `transportSecurityType` when we don't get it from the core.

murgatroid99:
With the deprecation of the `grpc` package, we are not going to move forward with this.

PR #510, opened 2018-08-23
Pure JS: Implement public Channel API

murgatroid99:
This implements the [public Channel API proposal](https://github.com/grpc/proposal/blob/master/L32-node-channel-API.md) in the pure JS library. I also implemented `Client#getChannel`.

As part of this change I renamed `CallStream` to `Call` and the previously named `Call` to `SurfaceCall`, plus a couple of other changes. The naming scheme is a little confusing now, but I don't think there should be any meaningful user-visible changes.

### DIFF
--- a/packages/grpc-js-core/README.md
+++ b/packages/grpc-js-core/README.md
@@ -18,5 +18,14 @@ npm install @grpc/grpc-js
  - TLS channel credentials
  - Call credentials (for auth)
  - Simple reconnection
+ - Channel API
 
 This library does not directly handle `.proto` files. To use `.proto` files with this library we recommend using the `@grpc/proto-loader` package.
+
+## Some Notes on API Guarantees
+
+The public API of this library follows semantic versioning, with some caveats:
+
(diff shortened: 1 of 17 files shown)


packages/grpc-js-core/README.md
```diff
@@ -18,5 +18,14 @@ npm install @grpc/grpc-js
  - TLS channel credentials
  - Call credentials (for auth)
  - Simple reconnection
+ - Channel API
 
 This library does not directly handle `.proto` files. To use `.proto` files with this library we recommend using the `@grpc/proto-loader` package.
+
+## Some Notes on API Guarantees
+
+The public API of this library follows semantic versioning, with some caveats:
```
nicolasnoble: Maybe reflect some of that in the README of the native implementation ?

nicolasnoble: Yeah, there's no urgency anyway. I was thinking about doing the same comment about underscore and private members that aren't supposed to be part of the public API.

We might also want to start tagging private members using underscores :)

PR #511, opened 2018-08-23
Mark some methods of ChannelCredentials as internal

murgatroid99:
This should make the `ChannelCredentials` classes in the two libraries have the same API.

### DIFF
--- a/packages/grpc-js-core/src/call-credentials-filter.ts
+++ b/packages/grpc-js-core/src/call-credentials-filter.ts
@@ -38,7 +38,7 @@ export class CallCredentialsFilterFactory implements
     FilterFactory<CallCredentialsFilter> {
   private readonly credentials: CallCredentials;
   constructor(channel: Http2Channel) {
-    this.credentials = channel.credentials.getCallCredentials();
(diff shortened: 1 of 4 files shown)


kjin:
Would it be possible to make all hidden methods `protected`, and expose them through an internal subclass? E.g.

```
class ChannelCredentials {
  protected _getConnectionState() {}
}

class InternalChannelCredentials {
  getConnectionState() { return this._getConnectionState(); }
}

// ... SecureChannelCredentialsImpl overrides InternalChannelCredentials
```

Or, we could use symbols for the hidden methods? Though, I'm not sure if TS supports exporting classes with public symbol fields.

murgatroid99:
I don't think I like that option because it ends up requiring some typecasts that I would prefer to avoid. The `Channel` API accepts a generic `ChannelCredentials` object that the user gets from the `ChannelCredentials` API, and it would have to typecast it to whatever internal type to access the internal functions.

nicolasnoble:
Right, as long as we don't have a proper system to declare friend classes, this is the best we might come up with.

PR #515, opened 2018-08-25
grpc-js-core: metadata cleanup

cjihrig:
This PR:
- Simplifies regular expression usage.
- Adds a missing space to an error message.
- Caches `Object.prototype.hasOwnProperty()` to prevent tampering.
- Simplifies the code in `fromHttp2Headers()`.

Another simplification (that would get rid of the `hasOwnProperty()` usage) would be to make `internalRepr` a `Map`, but I wanted to check with the maintainers first.

### DIFF
--- a/packages/grpc-js-core/src/metadata.ts
+++ b/packages/grpc-js-core/src/metadata.ts
@@ -1,5 +1,7 @@
 import * as http2 from 'http2';
 import {forOwn} from 'lodash';
+const LEGAL_KEY_REGEX = /^[0-9a-z_.-]+$/;
+const LEGAL_NON_BINARY_VALUE_REGEX = /^[ -~]*$/;
 
 export type MetadataValue = string|Buffer;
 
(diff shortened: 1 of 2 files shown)


packages/grpc-js-core/src/metadata.ts
```diff
@@ -1,5 +1,8 @@
 import * as http2 from 'http2';
 import {forOwn} from 'lodash';
+const hasOwnProperty = Object.prototype.hasOwnProperty;
```
murgatroid99: I don't understand the value of caching `hasOwnProperty` like this. If it's trying to solve the problem of some other library modifying `Object.prototype`, this only handles the case where that is modified after this library is loaded.

cjihrig: Yea, that's the problem it addresses. It can be useful to prevent tampering by code that is loaded later on. We do it a bit in Node core, but I can rebase it out if you don't want it.

murgatroid99: I guess I don't see the particular value in preventing tampering after the library is loaded if the prototype could just as easily be tampered with before the library is loaded. In addition, part of what I'm unsure about is why you are particularly worried about that method. What about `RegExp.prototype.test`, or `Array.prototype.map`, or any other built in function that we use?

PR #532, opened 2018-09-08
grpc-js-core: add ListenerBuilder support

cjihrig:
This commit ports `ListenerBuilder` to TypeScript.


murgatroid99:
I don't think it makes sense to add this now when we don't have the client interceptors API implemented in this library yet.

cjihrig:
I was planning on porting these things a bit at a time.

murgatroid99:
I understand that, but I think it would be confusing to add the ListenerBuilder class before we have any API that accepts a listener.

cjihrig:
That's fine. No rush to merge this.

PR #534, opened 2018-09-09
grpc-js-core: ignore reserved headers in fromHttp2Headers()

cjihrig:
`Metadata.fromHttp2Headers()` throws if any reserved headers are passed. Instead of deleting headers before calling the function, this commit causes the function to ignore reserved headers.

As a side note, this should make `Metadata.fromHttp2Headers()` easier to use on the server side, where clients may send a wider variety of headers.

### DIFF
--- a/packages/grpc-js-core/src/call-stream.ts
+++ b/packages/grpc-js-core/src/call-stream.ts
@@ -233,8 +233,7 @@ export class Http2CallStream extends Duplex implements Call {
           default:
             this.mappedStatusCode = Status.UNKNOWN;
         }
-        delete headers[HTTP2_HEADER_STATUS];
-        delete headers[HTTP2_HEADER_CONTENT_TYPE];
+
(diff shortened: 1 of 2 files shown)


murgatroid99:
I am not sure about this one. It would definitely be simpler to just ignore those headers like you suggest, but the value of deleting those headers is that we are explicitly "consuming" them, and nothing gets silently ignored.

cjihrig:
So, I know there isn't a server implementation now, but I think the benefit would be more evident once there is. Otherwise, the server would have to perform another loop over all of the headers to filter them before calling `fromHttp2Headers()`. `delete` is also a fairly expensive operation in JavaScript. I think that since the point of this function is to create metadata from HTTP2 headers, it makes sense (and should be documented) because all sets of HTTP2 headers will be expected to have these.

cjihrig:
Thank you 🙏

PR #536, opened 2018-09-09
grpc-js-core: tidy up use of promises

cjihrig:
This commit removes many unnecessary wrapping and unwrapping of values in `Promise`s. This should simplify both the source code and generated code, as well as help performance.

### DIFF
--- a/packages/grpc-js-core/src/call-credentials-filter.ts
+++ b/packages/grpc-js-core/src/call-credentials-filter.ts
@@ -23,12 +23,11 @@ export class CallCredentialsFilter extends BaseFilter implements Filter {
     this.serviceUrl = `https://${host}/${serviceName}`;
   }
 
-  async sendMetadata(metadata: Promise<Metadata>): Promise<Metadata> {
+  async sendMetadata(metadata: Metadata): Promise<Metadata> {
     const credsMetadata =
         this.credentials.generateMetadata({service_url: this.serviceUrl});
-    const resultMetadata = await metadata;
-    resultMetadata.merge(await credsMetadata);
-    return resultMetadata;
+    metadata.merge(await credsMetadata);
+    return metadata;
   }
 }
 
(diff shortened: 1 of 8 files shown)


murgatroid99:
Thank you for your contribution, but I have to decline this one. I had good reasons for making the filter API take promises as arguments. Unless I've made a mistake, this effectively allows all of the filters to start work simultaneously, except for any that need the result of the previous one. This matters for filters that perform asynchronous operations independent of the others, like the deadline filter.

murgatroid99:
That has the same problem that each step doesn't start processing until the previous one finishes.

cjihrig:
Isn't that how the existing code behaves? `_.flow()` calls each function with the output of the previous function. Currently, promises are passed to all of these functions, but then the `await`s stop things from progressing until the previous task completes.

murgatroid99:
I think there's more nuance than that. An `async` function can't just block waiting for an asynchronous promise. My understanding is that it returns a promise that is resolved when the function returns. So each `await` call makes that specific function wait for the previous functions, but it still allows the following ones to run. And if you look at the deadline filter in particular, it doesn't use an `async` function, it explicitly returns a promise that is resolved when both it's own work and the previous promise is resolved.

nicolasnoble:
In other words, even if their outputs look similar, these two pieces of code will behave wildly differently, with drastically different performance impacts:

case 1:
```js
const promiseArray = generateSomePromiseArray()
const promiseResults = await Promise.all(promiseArray)
```

case 2:
```js
const promiseArray = generateSomePromiseArray()
const promiseResults = []
for (let i = 0; i < promiseArray.length; i++) {
  promiseResults[i] = await promiseArray[i]
}
```

cjihrig:
I don't think the loop to `Promise.all()` comparison is really accurate in this case because series behavior is still enforced in almost all of the filters.

That said, I can see the case for allowing two asynchronous filters to execute at once.

PR #538, opened 2018-09-10
grpc-js-core: remove use of return await

cjihrig:
`return await` isn't typically useful in `async` functions. Split out of #536


nicolasnoble:
I can get behind this one.

PR #554, opened 2018-09-25
grpc-js-core: delay composing credentials

cjihrig:
The channel and stream credentials may change between the time a `CallCredentialsFilterFactory` is created and the time that the metadata is sent. This commit delays composing the credentials until the time metadata is sent.

For context, I observed test failures when updating from 0.2.0 to 0.3.0. No other changes were involved. After some debugging, I noticed that incorrect metadata was being sent to the server because the composed credentials were created too early. [These tests](https://github.com/cjihrig/grpc-server-js/blob/master/test/server-credentials.js#L235-L334), ported from this repo, are what began failing (as well as some tests in private repos).


murgatroid99:
I'll do a patch release. I was just waiting until your other PR was merged to avoid doing two patch releases back to back.

Issue #556, opened 2018-09-26
support of the protobuf protoset binary format?


c0b:
from konsumer/grpc-dynamic-gateway#15
and https://groups.google.com/d/msg/grpc-io/dAJz2sPwir8/TlgLny-PBgAJ nobody answered the protoset questions in this discussion thread ,

### Is your feature request related to a problem? Please describe.

google searched  grpc + nodejs + protoset  seems nothing, and searched this issues board as well, 
 no one has discussed about protoset support before;    

I see the compiled protoset binary format is a good solution to bundle many related `*.proto` files into a single one, it's pretty well supported by protoc compiler, and in Go language pkg `github.com/golang/protobuf`, and tools like grpcurl
https://github.com/fullstorydev/grpcurl#protoset-files

I hope this base library should include protoset support, for tools like the grpc-dynamic-gateway to be able to load from a single protoset binary file

### Describe the solution you'd like

the `golang/protobuf` package has good support of protoset support, I think this grpc-node should have similar support, to parse one binary protoset file only, instead of multiple plaintext `*.proto` files
https://github.com/golang/protobuf/tree/master/protoc-gen-go/descriptor

### Describe alternatives you've considered

No one discussed   grpc + nodejs + protoset before.

### Additional

make better documentation about how to use this `grpc.load` method: does it already support parameter as `{ file: p, root: include }` ??  does it already support multiple proto_path ? does it already support the protoset binary proto definition?

https://grpc.io/grpc/node/grpc.html#.load__anchor

```
<static> load
Load a gRPC object from a .proto file.
```

```console
$ protoc --help
Usage: protoc [OPTION] PROTO_FILES
Parse PROTO_FILES and generate output based on the options given:
  -IPATH, --proto_path=PATH   Specify the directory in which to search for
                              imports.  May be specified multiple times;
                              directories will be searched in order.  If not
                              given, the current working directory is used.

  -oFILE,                     Writes a FileDescriptorSet (a protocol buffer,
    --descriptor_set_out=FILE defined in descriptor.proto) containing all of
                              the input files to FILE.
```

nicolasnoble:
Sorry, but what you're describing isn't a feature request for grpc-node, but for the third party package `protobufjs`.

There's two ways of using grpc-node. One is the static codepath, which is using protoc at build time to generate service files, and one is the dynamic codepath, which is using the third-party protobufjs package under the hood, where you load your proto files at runtime.

[...]

PR #558, opened 2018-09-26
Documentation updates.

nicolasnoble:
Addresses concerns from #556.

### DIFF
--- a/packages/grpc-protobufjs/README.md
+++ b/packages/grpc-protobufjs/README.md
@@ -2,7 +2,7 @@
 
 A utility package for loading `.proto` files for use with gRPC, using the latest Protobuf.js package.
 Please refer to [protobuf.js' documentation](https://github.com/dcodeIO/protobuf.js/blob/master/README.md)
-to understands its usage and limitations.
+to understands its features and limitations.
 
 ## Installation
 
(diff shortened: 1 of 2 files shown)


packages/grpc-protobufjs/README.md
```diff
@@ -1,6 +1,8 @@
 # gRPC Protobuf Loader
 
 A utility package for loading `.proto` files for use with gRPC, using the latest Protobuf.js package.
+Please refer to [protobuf.js' documentation](https://github.com/dcodeIO/protobuf.js/blob/master/README.md)
+to understands its usage and limitations.
```
murgatroid99: I'm not sure about pointing to Protobuf.js for "usage". The usage of this library is defined in this documentation, not in the Protobuf.js documentation.

nicolasnoble: I wanted to say something of the ilk of "pros and cons", but since we don't have any comparison point, I'm not sure this idiom would be appropriate. Any suggestion ?

nicolasnoble: Its features, maybe?

murgatroid99: I think "features and limitations" would work here.

PR #561, opened 2018-09-28
grpc-js-core: make semver a prod dependency

cjihrig:
`semver` is now used in `index.ts`, meaning that it needs to be included in the `"dependencies"` section of the `package.json`, otherwise deployments that use `npm i --production` will fail.


nicolasnoble:
Whoops.

nicolasnoble:
I think this one and #560 warrant a 0.3.2.

PR #592, opened 2018-10-25
Updating head again.

nicolasnoble:
My previous attempt broke, so let's see if it reproduces.


nicolasnoble:
Okay, this was just a fluke. Good.

nicolasnoble:
Or I spoke too soon, apparently.

nicolasnoble:
Ah, I see: `ERROR: Aborting VM command due to timeout of 3600 seconds`. Our builds are now a bit too extensive. We really need to split up the builds.

PR #606, opened 2018-10-31
grpc-js-core: remove simple uses of lodash

cjihrig:
This commit removes lodash as a production dependency. It remains as a devDependency because it's used in tests, but the uses in the `src/` directory were easily replaced with vanilla JavaScript.


cjihrig:
It looks like there is a regression. I can reproduce it locally and will push up a fix.

cjihrig:
Fixed the regression.

Issue #602, opened 2018-10-30
Rename directories in `packages` directory


cjihrig:
### Is your feature request related to a problem? Please describe.
It would be nice if the directories inside of `packages` in this repo had the same names as their corresponding npm packages.

### Describe the solution you'd like
In an ideal world, I'd get rid of the monorepo completely. You could also rename the directories. I'm happy to do the work, but I suspect it would break internal tooling such as CI.

### Describe alternatives you've considered
n/a

### Additional context
n/a

nicolasnoble:
Yes, we wanted to do this for a while :-) Priorities...

cjihrig:
> Priorities...

Is this something the community can help with, or will it break a bunch of internal things?

murgatroid99:
Doing that shouldn't break anything internal, as long as any references within the repo are fixed. My main concern is that it might break external links into the repo.

PR #608, opened 2018-11-01
Rename grpc-js-core and grpc-protobufjs to match npm

cjihrig:
Fixes: https://github.com/grpc/grpc-node/issues/602 (although `grpc-tools` and `grpc-health-check` will need to be revisited if they get published on npm under the `@grpc` scope).


murgatroid99:
It looks like some references did not get changed.

PR #615, opened 2018-11-10
refactor: use individual lodash package for proto loader

JustinBeckwith:
This drops the `lodash` package in favor of `lodash.camelcase`

### DIFF
--- a/packages/proto-loader/package.json
+++ b/packages/proto-loader/package.json
@@ -38,12 +38,12 @@
     "build/src/*.js"
   ],
   "dependencies": {
-    "@types/lodash": "^4.14.104",
-    "@types/node": "^9.4.6",
-    "lodash": "^4.17.5",
+    "lodash.camelcase": "^4.3.0",
     "protobufjs": "^6.8.6"
   },
   "devDependencies": {
+    "@types/lodash.camelcase": "^4.3.4",
+    "@types/node": "^10.12.5",
     "clang-format": "^1.2.2",
     "gts": "^0.5.3",
     "typescript": "~2.7.2"
--- a/packages/proto-loader/src/index.ts
+++ b/packages/proto-loader/src/index.ts
@@ -18,7 +18,7 @@
 import * as Protobuf from 'protobufjs';
 import * as fs from 'fs';
 import * as path from 'path';
-import * as _ from 'lodash';
+import camelCase = require('lodash.camelcase');
 
 export interface Serialize<T> {
   (value: T): Buffer;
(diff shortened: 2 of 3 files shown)


packages/proto-loader/src/index.ts
```diff
@@ -18,7 +18,7 @@
 import * as Protobuf from 'protobufjs';
 import * as fs from 'fs';
 import * as path from 'path';
-import * as _ from 'lodash';
+import camelCase = require('lodash.camelcase');
```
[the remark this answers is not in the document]
JustinBeckwith: Yeah, it is a little weird. It's like this because the DefinitelyTyped package uses the `export = ` syntax instead of es modules syntax:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/lodash.camelcase/index.d.ts

The behavior is described here under "export = and import = require()":
https://www.typescriptlang.org/docs/handbook/modules.html

packages/proto-loader/package.json
```diff
@@ -38,12 +38,12 @@
...
   "dependencies": {
-    "@types/lodash": "^4.14.104",
-    "@types/node": "^9.4.6",
-    "lodash": "^4.17.5",
+    "lodash.camelcase": "^4.3.0",
     "protobufjs": "^6.8.6"
   },
   "devDependencies": {
+    "@types/lodash.camelcase": "^4.3.4",
+    "@types/node": "^10.12.5",
```
murgatroid99: How do you decide whether this should be in `dependencies` or `devDependencies`. I would think it would need to be in the `dependencies` because some of the exported types use the `Buffer` type, so users would need the `Buffer` type definition to compile against it.

nicolasnoble: The rule of thumb should be that if it's not needed for production usage, it's a dev dependency. In this case, the types should be only needed to compile the typescript into javascript, and shouldn't be needed to run in production. I think that having typescript types as a dev dependency is the correct thing to do in this case.

murgatroid99: I would say that the dependencies should contain anything that the user needs to *use* our library. The users never get our dev dependencies, so if type definitions that we depend on only appear in our dev dependencies, they will not resolve properly when users try to compile their code.

JustinBeckwith: I'm interested in @ofrobots opinion here too.  In general, I try to avoid including @types packages in the dependencies section, mostly cause that gets deployed to production.  In this particular case - is there a scenario where the user will need the node types, but not already have them installed?  I think we install the @types/nodejs package in all of our typescript projects anyway...

ofrobots: The generally understood convention in the community is along the lines that @nicolasnoble and @JustinBeckwith put it. However, @murgatroid99 is not incorrect. TypeScript users who need to use this library cannot use it unless they *also* depend on `@types/node`. IMO this is a flaw in either TypeScript or npm. Arguably it could be solved if there was a concept of a `typeDependency` in `package.json` that were to be installed only for dependents' builds but not their production deployments.

But given that doesn't exist, I think it would be better match user expectation (most of which are JS users) that `@types/node` was not installed in production. As @JustinBeckwith points out, it is not quite likely that there are TS user of this library doesn't already have a dependency on `@types/node`.

murgatroid99: I can accept that this is probably the right answer specifically with `@types/node`, but I think it's worth revisiting if we encounter a similar problem with a different type library.

PR #620, opened 2018-11-10
refactor: drop usage of _.isObject

JustinBeckwith:
**NOTE: It's very possible this one won't work and is a bad idea.**

This drops `_.isObject` in favor of the native `typeof foo === 'object'` approach.

### DIFF
--- a/packages/grpc-native-core/src/server.js
+++ b/packages/grpc-native-core/src/server.js
@@ -861,6 +861,10 @@ var defaultHandler = {
   }
 };
 
+function isObject(thing) {
+  return (typeof thing === 'object' || typeof thing === 'function') && thing !== null;
+}
+
 /**
  * Add a service to the server, with a corresponding implementation.
  * @param {grpc~ServiceDefinition} service The service descriptor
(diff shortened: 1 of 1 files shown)


packages/grpc-native-core/src/server.js
```diff
@@ -868,7 +868,7 @@ var defaultHandler = {
  *     names to method implementation for the provided service.
  */
 Server.prototype.addService = function(service, implementation) {
-  if (!_.isObject(service) || !_.isObject(implementation)) {
+  if (typeof service !== 'object' || typeof implementation !== 'object') {
```
cjihrig: This should probably also check for `null` explicitly since `typeof null === 'object'`.

murgatroid99: I agree. In addition, the lodash function also returns `true` for values with a `typeof` that outputs `"function"`. That may be undesirable, though.

murgatroid99:
The Windows failures are just Kokoro errors, rerunning to be sure.

PR #630, opened 2018-11-13
refactor: use Object.assign

JustinBeckwith:
This replaces the usage of `_.defaults` with `Object.assign`.  Note that the order of assignment with this change is reversed.


murgatroid99:
`Object.assign` modifies the first argument, so almost all of these changes are invalid.

JustinBeckwith:
Good catch :)  Updated to assign into a new obj.  PTAL!

murgatroid99:
I've never seen that MacOS failure before, but I also don't see how these changes could have caused it. I'm rerunning the tests again.

JustinBeckwith:
@murgatroid99 looks like the re-run fix it 🤷‍♂️

PR #635, opened 2018-11-15
refactor: drop usage of _.zipObject

JustinBeckwith:
Replaces `_.zipObject` with a reduce function.

### DIFF
--- a/packages/grpc-native-core/src/common.js
+++ b/packages/grpc-native-core/src/common.js
@@ -112,6 +112,20 @@ exports.getMethodType = function(method_definition) {
   }
 };
 
+/**
+ * Given an array of property names and an array of values,
+ * combine the two into an object map.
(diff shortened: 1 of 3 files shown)


packages/grpc-native-core/src/common.js
```diff
@@ -112,6 +112,20 @@ exports.getMethodType = function(method_definition) {
   }
 };
 
+/**
+ * Given an array of property names and an array of values,
+ * combine the two into an object map.
+ * Equivalent to _.zipObject.
```
murgatroid99: This should be marked `@private` in the JSDoc so that it doesn't show up in the generated documentation, since it's not part of the public API.

JustinBeckwith: Good call!

PR #636, opened 2018-11-15
refactor: replace _.assign with Object.assign

JustinBeckwith:
replace _.assign with Object.assign


murgatroid99:
The Mac test failure is just a timeout, almost certainly a flake.

PR #640, opened 2018-11-15
refactor: drop usage of _.get

JustinBeckwith:
refactor: drop usage of _.get


murgatroid99:
The only test failure is a timeout on MacOS. That happens from time to time and I'm not worried about it.

PR #641, opened 2018-11-16
refactor: drop usage of _.partial

JustinBeckwith:
refactor: drop usage of _.partial

### DIFF
--- a/packages/grpc-native-core/src/client.js
+++ b/packages/grpc-native-core/src/client.js
@@ -940,10 +940,9 @@ exports.makeClientConstructor = function(methods, serviceName,
     }
     var method_type = common.getMethodType(attrs);
     var method_func = function() {
-      return requester_funcs[method_type].apply(this,
-        [].slice.call(arguments).concat([
-          attrs.path, attrs.requestSerialize, attrs.responseDeserialize
-        ])
+      return requester_funcs[method_type].apply(this, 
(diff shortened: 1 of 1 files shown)


packages/grpc-native-core/src/client.js
```diff
@@ -939,9 +939,11 @@ exports.makeClientConstructor = function(methods, serviceName,
       throw new Error('Method names cannot start with $');
     }
     var method_type = common.getMethodType(attrs);
-    var method_func = _.partial(requester_funcs[method_type], attrs.path,
-                                attrs.requestSerialize,
-                                attrs.responseDeserialize);
+    var method_func = function() {
+      return requester_funcs[method_type].apply(this, arguments.concat([
+        attrs.path, attrs.requestSerialize, attrs.responseDeserialize
```
murgatroid99: I think this is putting those arguments at the end of the final list, instead of at the beginning.

murgatroid99:
OK, you didn't actually change the thing I was pointing out. You're still concatenating the preset arguments at the end of the arguments list, but they should go at the beginning.

JustinBeckwith:
🤦‍♂️

murgatroid99:
I think slicing the arguments list is now redundant.

murgatroid99:
The Linux build failure was a Kokoro error, and there's no way it was impacted by this change.

PR #642, opened 2018-11-16
refactor: drop usage of _.mapValues

JustinBeckwith:
refactor: drop usage of _.mapValues

### DIFF
--- a/packages/grpc-native-core/src/client.js
+++ b/packages/grpc-native-core/src/client.js
@@ -983,6 +983,7 @@ exports.getClientInterceptors = function(client) {
   return Object.keys(client.$method_definitions)
     .reduce((acc, key) => {
       acc[key] = client[key].interceptors;
+      return acc;
     }, {});
 };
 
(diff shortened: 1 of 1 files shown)


murgatroid99:
I'm not sure exactly why yet but the tests are failing with these errors:

```
  1) Client interceptors uses interceptors passed to the client constructor:
     TypeError: Cannot set property 'echoClientStream' of undefined
      at Object.keys.reduce (packages/grpc-native-core/src/client.js:281:243)
      at Array.reduce (<anonymous>)
      at Object.exports.getClientInterceptors (packages/grpc-native-core/src/client.js:281:169)
      at Context.<anonymous> (packages/grpc-native-core/test/client_interceptors_test.js:994:34)
  2) Client interceptors pass calls through when no interceptors provided with unary call:
     TypeError: Cannot set property 'echoClientStream' of undefined
      at Object.keys.reduce (packages/grpc-native-core/src/client.js:281:243)
      at Array.reduce (<anonymous>)
      at Object.exports.getClientInterceptors (packages/grpc-native-core/src/client.js:281:169)
      at Context.<anonymous> (packages/grpc-native-core/test/client_interceptors_test.js:189:36)
```

PR #643, opened 2018-11-16
refactor: drop usage of _.each

JustinBeckwith:
refactor: drop usage of _.each


murgatroid99:
The Mac failure is just a single test timeout, which is almost definitely spurious and not related to this change.

PR #644, opened 2018-11-16
refactor: drop usage of _.omit

JustinBeckwith:
refactor: drop usage of _.omit


murgatroid99:
For some reason the health check package tests are failing with this error, but I can't really tell why:

```
     TypeError: options.getOwnPropertyNames is not a function
      at ServiceClient.Client (packages/grpc-native-core/src/client.js:162:210)
      at new ServiceClient (packages/grpc-native-core/src/client.js:269:374)
      at Context.<anonymous> (packages/grpc-health-check/test/health_test.js:47:20)
```

nicolasnoble:
Isn't it because "options" is `null` or `undefined` ?

murgatroid99:
That should be handled up in line 369

JustinBeckwith:
I think I got it this thing 😆

PR #647, opened 2018-11-16
native-core: remove duplicate package.json entry

cjihrig:
lodash is already listed as a prod dependency


nicolasnoble:
We are in the process of heavily moving the lodash dependency out, this may be redundant.

PR #651, opened 2018-11-20
Native: lazily load Protobuf.js

### LINKED ISSUE #626: Perf: lazy load protobufjs dependency in grpc-core
We were talking about speeding up load times, and mentioned.....

> The older version of Protobuf.js is basically baked into the grpc library, and we can't get rid of it or upgrade it. That's the whole reason we started building the proto-loader library. Even though we can't get rid of it, we can probably switch to lazily loading it so that most users don't see the hit for loading it.

murgatroid99:
Fixes #626.

### DIFF
--- a/packages/grpc-native-core/index.js
+++ b/packages/grpc-native-core/index.js
@@ -26,8 +26,6 @@ var SSL_ROOTS_PATH = path.resolve(__dirname, 'deps', 'grpc', 'etc', 'roots.pem')
 
 var _ = require('lodash');
 
-var ProtoBuf = require('protobufjs');
-
 var client = require('./src/client.js');
 
 var server = require('./src/server.js');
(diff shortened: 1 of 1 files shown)


packages/grpc-native-core/index.js
```diff
@@ -122,6 +120,7 @@ var loadObject = exports.loadObject;
  * @return {Object<string, *>} The resulting gRPC object
  */
 exports.load = util.deprecate(function load(filename, format, options) {
+  const ProtoBuf = require('protobufjs');
```
cjihrig: It probably doesn't matter for a deprecated function, but if it's called more than once, then doing something similar to [this](https://github.com/nodejs/node/blob/08c14d98cdb029c72549a6709c2fd0befda5a463/lib/internal/dgram.js#L27-L29) might be even more performant (if just slightly). Even though `require()` has a cache, there is still some additional overhead to using it.

murgatroid99: I think it's fine to keep it the way I wrote it, not only because `load` is deprecated, but also because it's expected to be called very infrequently, and it shouldn't generally be on the hot path.

murgatroid99:
The Mac failures are probably just a test timeout, but I'll rerun just in case.

PR #654, opened 2018-11-23
Fetch options also in method

w4-hojin:
Fetch options also in method


murgatroid99:
Can you explain what the purpose/goal of this change is?

w4-hojin:
I am implementing https://github.com/grpc-ecosystem/grpc-gateway with node js. But loaded Grpc loses its 'options' information. 
For example, 
https://github.com/konsumer/grpc-dynamic-gateway/blob/master/index.js#L44 
It also parses itself, instead of using the output of 'grpc-loader'.
Please consider submitting this code. Thank you!

murgatroid99:
I'm sorry for the delay responding here.

First, I am unable to accept any change from you until you sign the CLA linked in the comment above.

Second, one of the important design considerations in this library is to ensure that its own types are independent of Protobuf.js types. So, this PR needs to modify the `MethodDefinition` type definition to include a complete type for the `options` field. If that type would include any types defined by Protobuf.js, then you also need to transform the `options` object into one that uses more generic types.

murgatroid99:
Superseded by #2230.

PR #665, opened 2018-11-30
fix: require package.json directly

mrfelton:
# Problem

When running jest tests against a project that includes `grpc-js` I'm getting the following module resolve related error from jest:

```
 FAIL  test/unit/lnd/util.spec.js
 Test suite failed to run

    Cannot find module '../../package' from 'channel.js'

      at Resolver.resolveModule (node_modules/jest-resolve/build/index.js:221:17)
      at Object.<anonymous> (node_modules/@grpc/grpc-js/build/src/channel.js:16:36)
```

This is caused by the following code:

```javascript
require('../../package')
```

The problem is that grpc-js is relying on the node module resolver to load in package.json when there is really no need for it to do this.

This is the only place in the code where `package.json` is required in this way. All other places it is required as a file (`require('../../package.json')`

# Solution

Require package.json as a file rather than relying on the node module resolver:

change:
```javascript
require('../../package')
```

to:
```javascript
require('../../package.json')
```

### DIFF
--- a/packages/grpc-js/src/channel.ts
+++ b/packages/grpc-js/src/channel.ts
@@ -15,7 +15,7 @@ import {Metadata} from './metadata';
 import {MetadataStatusFilterFactory} from './metadata-status-filter';
 import {Http2SubChannel} from './subchannel';
 
-const {version: clientVersion} = require('../../package');
+const {version: clientVersion} = require('../../package.json');
 
 const MIN_CONNECT_TIMEOUT_MS = 20000;
 const INITIAL_BACKOFF_MS = 1000;


thelinuxfoundation:

* User @mrfelton isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

murgatroid99:
This is primarily an issue with Jest. There is a clear [specification of the require algorithm](https://nodejs.org/api/modules.html#modules_all_together), and it includes resolving `.json` files as our code depends on. So, please bring this up with Jest first, and if they do not fix their bug then I will be more willing to make this change here.

mrfelton:
@murgatroid99 there are advantages to requiring `package.json` as a file rather than a package.

1) performance - no need to invoke the module resolve system needlessly.
2) consistency - everywhere else in this codebase it is required as a file rather than a module. Why is the file being handled as a module in this case but as a file everywhere else?

murgatroid99:
It looks like this is not actually a Jest bug. There is a [Jest configuration option](https://jestjs.io/docs/en/configuration.html#modulefileextensions-array-string) that affects what file extensions it looks for when loading modules. By default it includes `json`. Are you modifying that configuration option to remove `json`? If so, changing that configuration will be the fastest way to resolve this problem.

Regarding your other comments, I do not believe that there is any meaningful performance here. With or without this change, the name resolution algorithm is being applied. The only difference is which step of the algorithm it ends on. You do have a good point regarding consistency, though, and I am OK with merging this to make that improvement.

mrfelton:
Good catch @murgatroid99 - yes, it seems that we did have that setting set.

I've updated here - https://github.com/LN-Zap/zap-desktop/pull/975/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R188

PR #668, opened 2018-12-05
Support for protobufjs's bundled definitions

litichevskiydv:
The `protobufjs` library in the` common.js` module contains definitions for standard types such as `google.protobuf.Empty`,` google.protobuf.Timestamp`, etc. This fix allows you to use them in a loadable proto-file without providing files containing them.


murgatroid99:
I would have expected this to work anyway, without any changes. Are you currently experiencing problems loading those `.proto` files?

litichevskiydv:
I signed it

litichevskiydv:
I signed it.

murgatroid99:
Before I merge this, I want to get a better understanding of why this is failing. There may be a better solution.

murgatroid99:
Thank you for the bug report and the contribution, but I believe that the solution in #676 is both simpler and more comprehensive.

PR #678, opened 2018-12-11
Native: Protobuf.js 6: use nested again

murgatroid99:
Fix the fix in #674.


nicolasnoble:
We may need to introduce some "well known structure" test :)

PR #684, opened 2018-12-18
Update proto-loader to 0.3.1

murgatroid99:
This is a patch because the only changes are #615 and #676. #615 has a minor potential behavior change, but only in rare cases that were previously errors.


nicolasnoble:
All the failures are caused by that annoying NaN thing.

murgatroid99:
In the previous run just the Mac tests failed, and this time the Windows tests just timed out

PR #685, opened 2018-12-18
Pin nan to non-buggy recent version

murgatroid99:
Nan version 2.12.0 is buggy; we are specifically having a problem with nodejs/nan#83, so we pin to 2.11.1. We should undo this once nodejs/nan#833 is merged and published.


nicolasnoble:
It looks like they will release the fix within a few hours, so this one kludge here should really be ultra-temporary.

PR #703, opened 2019-01-14
Add message and enum type information to package definition output.

### LINKED ISSUE #407: Add RequestType and ResponseType Metadata

murgatroid99:
This is an implementation of [gRFC proposal L43](https://github.com/grpc/proposal/blob/master/L43-node-type-info.md), adding message and enum type information to the output of the `load` and `loadSync` functions. When loaded into a gRPC implementation using `loadPackageDefinition` those objects will appear in the expected place in the object tree. They will also be included in the `requestType` and `responseType` message properties.

This supersedes PRs #408 and #448, and fixes #407.

### DIFF
--- a/packages/proto-loader/test/descriptor_type_test.ts
+++ b/packages/proto-loader/test/descriptor_type_test.ts
@@ -1,0 +1,53 @@
+import * as assert from 'assert';
+
+import * as proto_loader from '../src/index';
+
+// Relative path from build output directory to test_protos directory
+const TEST_PROTO_DIR = `${__dirname}/../../test_protos/`;
+
+type TypeDefinition = proto_loader.EnumTypeDefinition | proto_loader.MessageTypeDefinition;
+
+function isTypeObject(obj: proto_loader.AnyDefinition): obj is TypeDefinition {
+  return 'format' in obj;
+}
+
+describe('Descriptor types', () => {
+  it('Should be output for each enum', (done) => {
+    proto_loader.load(`${TEST_PROTO_DIR}/enums.proto`).then((packageDefinition) => {
+      assert('Enum1' in packageDefinition);
+      assert(isTypeObject(packageDefinition.Enum1));
+      // Need additional check because compiler doesn't understand asserts
+      if(isTypeObject(packageDefinition.Enum1)) {
+        const enum1Def: TypeDefinition = packageDefinition.Enum1;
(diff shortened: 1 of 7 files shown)


packages/proto-loader/test/descriptor_type_test.ts
```diff
@@ -0,0 +1,53 @@
...
+      if(isTypeObject(packageDefinition.LongValues)) {
+        const longValuesDef: TypeDefinition = packageDefinition.LongValues;
+        assert.strictEqual(longValuesDef.format, 'Protocol Buffer 3 DescriptorProto');
+      }
+
+      assert('SequenceValues' in packageDefinition);
+      assert(isTypeObject(packageDefinition.SequenceValues));
+      if(isTypeObject(packageDefinition.SequenceValues)) {
+        const sequenceValuesDef: TypeDefinition = packageDefinition.SequenceValues;
+        assert.strictEqual(sequenceValuesDef.format, 'Protocol Buffer 3 DescriptorProto');
```
nicolasnoble: I am kind of wondering if we don't want a bit deeper comparisons here to validate the contents of what is returned? At least maybe not just `.format`, but `.type` as well?

nicolasnoble:
LGTM otherwise, after we can see tests are green thanks to #705.

packages/proto-loader/test/descriptor_type_test.ts
murgatroid99: I think this is a good idea, but it's a significant amount of work so we should do it at a later date. The `type` fields are just defined as `object` in TypeScript, so it does not like me looking at specific fields of those objects. The best solution for that is to probably write up full TypeScript definitions for those objects, but that is a lot of work and I would like to get this change usable first.

murgatroid99:
The Windows build timed out as usual, and the Linux Build failed with what looks like a CI flake. Rerunning the tests to be sure.

PR #725, opened 2019-02-01
Target grpc-js build at a higher ES standard

murgatroid99:
According to [this StackOverflow post](https://stackoverflow.com/a/50466512/159388), the ES2017 target should work for Node 8.10 and above, which is our compatibility boundary anyway.


cjihrig:
It would be nice to get this out in a release for the native `async/await` support and improved debugging experience. Is there any timeline on the next grpc-js release?

PR #732, opened 2019-02-12
grpc-tools build script fixes part 2

murgatroid99:
Maybe it will work this time

### DIFF
--- a/packages/grpc-tools/build_binaries.ps1
+++ b/packages/grpc-tools/build_binaries.ps1
@@ -33,11 +33,11 @@ foreach ($Arch in $ArchList) {
     $Generator = "Visual Studio 14 2015"
   }
 
-  & cmake.exe . --config Release
+  & cmake.exe .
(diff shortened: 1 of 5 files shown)


nicolasnoble:
Nope, still not found. Try adding
```
gci env:*
ls Env:
```
so we can see what's going on.

packages/grpc-tools/build_binaries.ps1
```diff
@@ -1,7 +1,11 @@
+$ErrorActionPreference = "Stop"
+
 Install-PackageProvider -Name NuGet -RequiredVersion 2.8.5.201 -Force
 Import-PackageProvider -Name NuGet -RequiredVersion 2.8.5.201
 Install-Module -Force -Name 7Zip4Powershell
 
+$env:Path += ";C:\Program Files\CMake\bin"
```
nicolasnoble: We probably still want to file a bug for this one.

Issue #744, opened 2019-02-20
grpc_tools_node_protoc in grpc-tools@1.7.0 fails with ENOENT


josephharrington:
### Problem description
When trying to run `protoc` from the `grpc-tools@1.7.0` package on linux, it fails with "No such file or directory".

This also manifests as `ENOENT` when trying to run `grpc_tools_node_protoc`:
```
# ./node_modules/.bin/grpc_tools_node_protoc
/toolstest/node_modules/grpc-tools/bin/protoc.js:41
    throw error;
    ^

Error: spawn /toolstest/node_modules/grpc-tools/bin/protoc ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:246:19)
    at onErrorNT (internal/child_process.js:427:16)
    at processTicksAndRejections (internal/process/next_tick.js:76:17)
    at process.runNextTicks [as _tickCallback] (internal/process/next_tick.js:51:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:804:11)
    at internal/main/run_main_module.js:21:11
```

From what I can tell, this is due to the `protoc` executable being a 32-bit binary in 1.7.0, whereas it was a 64-bit binary in 1.6.6:
```
$ npm install grpc-tools@1.6.6 && file node_modules/grpc-tools/bin/protoc
node_modules/grpc-tools/bin/protoc: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.18, BuildID[sha1]=1b2e70e2faad598dfc7849be2c4bd1e11190dde7, stripped

$ npm install grpc-tools@1.7.0 && file node_modules/grpc-tools/bin/protoc
node_modules/grpc-tools/bin/protoc: ELF 32-bit LSB executable, Intel 80386, version 1 (GNU/Linux), dynamically linked, interpreter /lib/ld-linux.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=d2f865292439a6ba1bdf92921b0961c147916f01, not stripped
```

### Reproduction steps 
(run on macos via docker. similarly reproducible on linux.)
```
$ docker run --rm -it node /bin/bash
root:/# mkdir toolstest && cd toolstest/
root:/# npm install grpc-tools@1.7.0
root:/# npx grpc_tools_node_protoc --version
spawn /toolstest/node_modules/grpc-tools/bin/protoc ENOENT
[...]

PR #750, opened 2019-02-20
grpc-tools: Force 64 bit Linux build

murgatroid99:
Hopefully fix the underlying problem from #744.


nicolasnoble:
Maybe quickly test and close the associated issue if this properly addressed it?

PR #765, opened 2019-03-06
Fix #764 close client duplex stream when bidi stream is closed from server with error

### LINKED ISSUE #764: BIDIStream doesnt emit status event after its closed from server with error
### Problem description
When Bidi stream is closed from server with error, client duplex stream doesnt emit status or error.

### Reproduction steps
see [#468](https://github.com/googleapis/nodejs-pubsub/issues/468)

### Environment
 - OS: any
 - Node version: any
 - Package name and version [e.g. gRPC@1.15.1]

ajaaym:
Fix #764

### DIFF
--- a/packages/grpc-native-core/src/client_interceptors.js
+++ b/packages/grpc-native-core/src/client_interceptors.js
@@ -1117,6 +1117,9 @@ function _getBidiStreamingInterceptor(method_definition, channel, emitter) {
             emitter.emit('error', err);
             return;
           }
+        } else if (!emitter.finished) {
+          // when bidistream is closed from server with error it doesnt have read_status
+          emitter._readsDone();
         }
         response.status.metadata = Metadata._fromCoreRepresentation(
           response.status.metadata);


murgatroid99:
I'm letting the tests run to see what happens, but I don't think this is a real problem, and even if it is this is not the right solution. The entire point of that `_readsDone` thing is to make sure that all incoming messages are handled before emitting a status, particularly because deserialization failures are supposed to result in an error status for an otherwise successful call. Calling that function from a context other than reading a message is simply incorrect.

The core APIs used in this section of the code make the guarantee that when there is a status, the next read will return `null` to signal the end of the stream. This is when that `_readsDone` function is normally called. Unless you have evidence either that this guarantee is broken or that the Node library is mishandling that `null` message, I see no reason to make any change to this library to attempt to fix this problem.

ajaaym:
@murgatroid99 Thanks for the input, I was not able to build locally so couldnt run test. I provided repro in issue, putting this fix did fixed that issue locally. I will close this PR and let you provide correct solution for this.

Issue #694, opened 2019-01-07
MaxListenersExceededWarning: Possible EventEmitter memory leak detected


shanshanzhu:
### Problem description
Node warning observed
 ```MaxListenersExceededWarning: Possible EventEmitter memory leak detected``` when retrying connection for serverStreaming.

### Reproduction steps
1. Set up grpc client connection without grpc server
2. Receive ` 'Internal HTTP2 error'` error  when the grpc server is not available
3. Create retry logic to reconnect to grpc server whenever there is an error emitted from the stream.
4. After retrying for  11~15 times,  we see this warning:
```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 connect listeners added. Use emitter.setMaxListeners() to increase limit
    at _addListener (events.js:280:19)
    at Http2Channel.addListener (events.js:297:10)
    at Http2Channel.once (events.js:341:8)
    at connecting.Promise (node_modules/@grpc/grpc-js/build/src/channel.js:248:26)
    at new Promise (<anonymous>)
    at Http2Channel.connect (node_modules/@grpc/grpc-js/build/src/channel.js:229:35)
    at Http2Channel._startHttp2Stream (node_modules/@grpc/grpc-js/build/src/channel.js:173:42)
    at Http2CallStream.sendMetadata (node_modules/@grpc/grpc-js/build/src/call-stream.js:325:22)
    at ServiceClientImpl.makeServerStreamRequest (node_modules/@grpc/grpc-js/build/src/client.js:171:14)
    at apply (node_modules/lodash/lodash.js:478:17)
    at ServiceClientImpl.wrapper (node_modules/lodash/lodash.js:5323:16)
```
### Environment
 - MacOS 10.14 (18A391)
 - Node version [v8.12.0]
 - Node installation method [e.g. nvm]
 - Package name and version [grpc-js@0.3.2]

### Additional context

Adding `process.on('warning', e => console.warn(e.stack));` will show the stack trace.

Above this line of code

https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/channel.ts#L340

If we add
```
                        console.log(this.listenerCount('connect'));
                        console.log(this.listenerCount('shutdown'));
                        this.removeListener('shutdown', onShutdown);
```
We see incrementing of `connect` listenerCount

Adding
```
 this.removeListener('connect', onConnect);
```
 to `onConnect` function fix this incrementing, and also fix the error.

[...]

PR #768, opened 2019-03-07
grpc-js: deadline filter: reject promise if call ends

murgatroid99:
This fixes a problem I encountered while trying to reproduce #694. Currently if the channel never connects the promise sticks around forever.

Edit: The original test that exposed this problem no longer shows it with these changes.

### DIFF
--- a/packages/grpc-js/src/deadline-filter.ts
+++ b/packages/grpc-js/src/deadline-filter.ts
@@ -60,6 +60,7 @@ export class DeadlineFilter extends BaseFilter implements Filter {
                    resolve(metadata);
                    this.channel.removeListener(
                        'connectivityStateChanged', handleStateChange);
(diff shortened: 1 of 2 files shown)


packages/grpc-js/src/deadline-filter.ts
```diff
@@ -62,7 +62,13 @@ export class DeadlineFilter extends BaseFilter implements Filter {
                        'connectivityStateChanged', handleStateChange);
                  }
                };
+               const handleStatus = () => {
```
cjihrig: Should `handleStatus()` and `handleStateChange()` remove the new `'status'` listener?

murgatroid99: You're right. I added that in `handleStateChange`, but it shouldn't be needed in `handleStatus` because that is a `once` listener.

PR #781, opened 2019-03-14
Removing support for Electron 4.0 and adding support for Electron 4.1.

nicolasnoble:
We won't be able to support the 4.0 Electron branch anymore, but we should be able to support the 4.1 one.


nicolasnoble:
All 3 remaining Windows tests have completed successfully, but are struggling to report their results. Merging.

PR #795, opened 2019-03-20
Array.isArray instead of instanceof for consistent checking

raunaqrox:
I'm running proto-loader from inside a [haraka](https://github.com/haraka/Haraka/) plugin.
When I send `includeDirs` as Array, `instanceof` fails. 
When I do `Array.isArray` it works correctly. 
Turns out `instanceof` does not work correctly for arrays in different contexts!


thelinuxfoundation:

* User @raunaqrox isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

raunaqrox:
I have signed CLA

PR #796, opened 2019-03-20
Add metadata options

murgatroid99:
This is an implementation of [proposal L48](https://github.com/grpc/proposal/blob/master/L48-node-metadata-options.md). For grpc-js this just adds the API to pass the options in. They currently have no effect, but this is required to make it a working drop-in TypeScript replacement.


murgatroid99:
The previous Windows test run failed on Node 6 and 9. Running again to see if that is consistent.

murgatroid99:
OK, the tests in this PR were predicated on the false assumption that "waitForReady" was the default behavior of the core.

murgatroid99:
The only test failures now are in the "default behavior" tests, and those are really existing bugs that are unrelated to this change. So I am going to remove those tests from this PR and move them to another PR.

PR #806, opened 2019-03-29
Fixed load/loadSync filename parameter type

### LINKED ISSUE #805: Incorrect load/loadSync method signature
The protobuf.js `load`/`loadSync` methods accept a filename parameter that is [either a string, or an array of strings](https://github.com/dcodeIO/protobuf.js/blob/6fa4c3487c50f9e2647a384bf64cfb009752b6a7/src/index-light.js#L17).

But the grpc proto-loader `load`/`loadSync` method defines the method to be a [single string only](https://github.com/grpc/grpc-node/blob/master/packages/proto-loader/src/index.ts#L268).

This parameter is not used other than to be passed as is to protobuf.js, so effectively it can be either, and if using from JavaScript, then you can pass an array of strings and it works, but an array of strings won't compile if using from TypeScript. So the `filename` parameter type should be changed to a union of `string` and `string[]`.

jroper:
Fixes #805.

Changed the type of the filename parameter to the proto-loader load method to match the type of the protobuf.js load method that it is passed to.


thelinuxfoundation:

* User @jroper isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

jroper:
CLA signed.

PR #809, opened 2019-03-31
grpc-js: add ServerCredentials support

cjihrig:
At gRPC Conf, I discussed porting the code in [`grpc-server-js`](https://www.npmjs.com/package/grpc-server-js) to `@grpc/grpc-js` with @murgatroid99. I think it makes the most sense to do this as smaller changes that are easier to author and review. `ServerCredentials` is a relatively isolated piece of code, so I decided to start there. The code in this PR is ported to TypeScript from https://github.com/cjihrig/grpc-server-js/blob/master/lib/server-credentials.js.

### DIFF
--- a/packages/grpc-js/src/index.ts
+++ b/packages/grpc-js/src/index.ts
@@ -27,6 +27,7 @@ import {LogVerbosity, Status} from './constants';
 import * as logging from './logging';
 import {Deserialize, loadPackageDefinition, makeClientConstructor, Serialize} from './make-client';
 import {Metadata} from './metadata';
+import {KeyCertPair, ServerCredentials} from './server-credentials';
 import {StatusBuilder} from './status-builder';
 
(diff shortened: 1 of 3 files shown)


cjihrig:
> In general, it would help to know how this is supposed to interact with other Node APIs. How are these objects eventually going to be passed to the relevant TLS or HTTP2 API?

They will be passed to `http2.createSecureServer()`

cjihrig:
@murgatroid99 updated! I think I've addressed all of your previous comments.

cjihrig:
I think something went wrong with the CI run...

murgatroid99:
It looks like we have a confluence of different failures here. The two Linux build failures are different problems we already know about and are working on correcting, the MacOS failure is probably spurious, and our CI just doesn't seem to be running Windows jobs for some reason. I'll rerun it soon to try to get some Windows results, but I don't think there are any problems here that are attributable to the code in this PR.

Issue #771, opened 2019-03-12
proto-loader: consider providing missing common google protos


cjihrig:
### Is your feature request related to a problem? Please describe.
protobufjs provides [some common Google types](https://github.com/dcodeIO/protobuf.js/blob/6fa4c3487c50f9e2647a384bf64cfb009752b6a7/src/common.js#L12-L18), but [not others](https://github.com/dcodeIO/protobuf.js/blob/6fa4c3487c50f9e2647a384bf64cfb009752b6a7/src/common.js#L34-L41). This is pretty confusing behavior, particularly if you've successfully used one, but run into 'unresolvable extensions' errors on another.

### Describe the solution you'd like
It would be great if the proto-loader module provided these. The change is straightforward, and I'm happy to send a PR if the maintainers are OK with it.

### Describe alternatives you've considered
This could be fixed in protobufjs, but I'm honestly not sure how maintained that repo currently is.

It can also be worked around in user code, but as previously mentioned, the current behavior is extremely confusing (and frustrating).

murgatroid99:
The first option for changing this should be Protobuf.js. If the change is rejected there then we can see about modifying this library to handle it.

cjihrig:
I agree. The problem I see is that protobufjs doesn't seem very responsive. For example, I opened an issue (unrelated to this) in October that hasn't been responded to at all. I'll open another issue, but how long do we wait if there is no response?

cjihrig:
https://github.com/dcodeIO/protobuf.js/issues/1180

xblv:
meet the same problem... could we have some workarounds here in proto-loader level?

PR #811, opened 2019-04-01
proto-loader: include Google well known protos

cjihrig:
This is a draft PR. It is not complete yet because there was some push back in #771 about implementing this in protobufjs, and I'd like feedback on the general approach taken. Since almost 3 weeks have passed with no response from protobufjs, I decided to open a WIP PR.

This commit ensures that the Google well known protos are available because protobufjs only exposes a subset of them.

Fixes: https://github.com/grpc/grpc-node/issues/771

### DIFF
--- a/packages/proto-loader/src/index.ts
+++ b/packages/proto-loader/src/index.ts
@@ -293,3 +293,20 @@ export function loadSync(filename: string, options?: Options): PackageDefinition
   loadedRoot.resolveAll();
   return createPackageDefinition(root, options!);
 }
+
+// Load Google's well-known proto files that aren't exposed by Protobuf.js.
+{
+  // Protobuf.js exposes: any, duration, empty, field_mask, struct, timestamp,
+  // and wrappers. compiler/plugin is excluded in Protobuf.js and here.
+  const wellKnownProtos = ['api', 'descriptor', 'source_context', 'type'];
+  const sourceDir = path.join(path.dirname(require.resolve('protobufjs')),
+                              'google', 'protobuf');
+
+  for (const proto of wellKnownProtos) {
+    const file = path.join(sourceDir, `${proto}.proto`);
+    const descriptor = Protobuf.loadSync(file).toJSON();
+
+    // @ts-ignore
+    Protobuf.common(proto, descriptor.nested.google.nested);
+  }
+}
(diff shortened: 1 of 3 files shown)


murgatroid99:
Most of the well-known protos are already loadable with Protobuf.js, and the ones that aren't are located in `protobufjs/google/protobuf`, so we should just be able to use those instead of getting them from another library like `grpc-tools`. If something is missing from both of those sets, it's probably worth just leaving out until someone needs it. The `compiler/plugin.proto` file, for example, is only useful for writing a plugin for `protoc`.

We probably want to use something like [get-installed-path](https://www.npmjs.com/package/get-installed-path) to consistently access that directory.

cjihrig:
> Most of the well-known protos are already loadable with Protobuf.js, and the ones that aren't are located in protobufjs/google/protobuf, so we should just be able to use those instead of getting them from another library like grpc-tools.

Right. I wasn't sure if there was any chance of the ones bundled with protobufjs being out of date. Using their copy simplifies things, and if I've checked properly, only `compiler/plugin.proto` is missing, which should be fine.

Do you want the well known protos available all of the time, or behind an option?

murgatroid99:
It should just load them all the time. I don't think optimizing for load time is worthwhile because it's something you should be doing infrequently, and the alternative of a load failure when you don't pass the option is not valuable.

cjihrig:
OK, I've made them load unconditionally. I also added a second commit to fix the lint. If you're OK with this, I can remove the draft status.

murgatroid99:
Instead of having the extra build step, I would prefer to just load the 3 or 4 extra files that we need directly from the protobuf.js directory at runtime either when the package is loaded or the first time a user tries to load a proto file.

xblv:
Hi @murgatroid99 @cjihrig could we publish a new version of proto-loader npm package? on npm this still of version 0.4.0.
Or since I am quite new with npm with git, Is there a way to directly use up-to-date code of proto-loader package? thanks.

nicolasnoble:
@xblv check https://docs.npmjs.com/cli/install, there are several parts talking about how to install from something local, from git, github, or other means.

murgatroid99:
This has now been published in version 0.5.0 of the library.

PR #812, opened 2019-04-02
V1.19 upmerge

murgatroid99:
We especially need to get some build fixes into master.


murgatroid99:
The Windows jobs didn't run. Rerunning

murgatroid99:
The Linux failure is the one we already know about.

PR #813, opened 2019-04-02
Disable a warning that is breaking our builds under gcc 8

murgatroid99:
We need to have our build working more than we need to fix this warning that we didn't even have to deal with before.


murgatroid99:
This approach isn't going to work.

PR #814, opened 2019-04-03
Update grpc submodule to fix gcc8 build failures

murgatroid99:
Let's see if this resolves everything.


murgatroid99:
The test logs for the Windows jobs show that they finished and succeeded, so the incomplete indicator is just a CI reporting issue.

PR #815, opened 2019-04-03
v1.20.0-pre1 submodule update and version bump

murgatroid99:
Finally we can actually publish a new version.


murgatroid99:
The Windows jobs have actually passed

PR #817, opened 2019-04-07
grpc-js: define Server API contract

cjihrig:
This PR builds on the work in https://github.com/grpc/grpc-node/pull/809. Again, the goal is to keep these PRs down to a manageable size (~700 lines here is already pushing that), so the server is not 100% functional.

This commit defines the `Server` API contract, and implements the `Server` functionality (with tests), minus the actual handling of requests. This commit intentionally does not publicly expose the `Server` functionality in `index.ts` yet.

### DIFF
--- a/packages/grpc-js/src/server-call.ts
+++ b/packages/grpc-js/src/server-call.ts
@@ -1,0 +1,148 @@
+/*
+ * Copyright 2019 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


packages/grpc-js/src/server-call.ts
```diff
@@ -0,0 +1,171 @@
...
+import {Metadata} from './metadata';
+
+
+export class ServerUnaryCall<RequestType> extends EventEmitter {
+  cancelled: boolean;
+  metadata: Metadata;
+  request: RequestType|null;
+  private call: ServerCall;
+
+  constructor(call: ServerCall, metadata: Metadata) {
```
murgatroid99: For this constructor and others, arguments that are assigned directly to instance members should be declared and assigned directly in the constructor, like `constructor(private call: ServerCall, public metadata: Metadata)`, then you can delete both the declaration lines and the assignment lines for those members.

cjihrig: I've updated the PR to addressed this. I'm not sure if there is a tslint rule for this or not. If so, it might be worth enabling.

PR #824, opened 2019-04-10
Re-add typescript dependency on packages that use it

murgatroid99:
It turns out that the `prepare` scripts need a direct dependency on Typescript.


nicolasnoble:
Did you mean to update the submodule?

PR #827, opened 2019-04-15
Update protobufjs module

### LINKED ISSUE #572: Cannot compile TypeScript hello world client/server
I was following the [quickstart document](https://grpc.io/docs/quickstart/node.html) to implement the hello world client server example in TypeScript.

I can compile and run the code fine if I have `noImplicitAny` set to `false` in `tsconfig.json`.  However, if I have it set to `true`, I get the following error when running `tsc -p .`:
```
node_modules/grpc/index.d.ts:2:55 - error TS7016: Could not find a declaration file for module 'protobufjs'. '.../grpc-example/node_modules/protobufjs/dist/protobuf.js' implicitly has an 'any' type.

2   import { Message, Service as ProtobufService } from "protobufjs";
                                                        ~~~~~~~~~~~~
```
Also, running `npm install @types/protobufjs` does not resolve the issue.

### Environment
 - OS name, version and architecture: OS X Mojave
 - Node version: 8.9.1
 - Package name and version: @grpc/proto-loader@0.3.0, grpc@1.15.1

jdforsythe:
Match version to that in @grpc/proto-loader

Will now include type definitions from `protobufjs` module

Fixes #572

### DIFF
--- a/packages/grpc-native-core/package.json
+++ b/packages/grpc-native-core/package.json
@@ -33,7 +33,7 @@
     "lodash.clone": "^4.5.0",
     "nan": "^2.0.0",
     "node-pre-gyp": "^0.12.0",
-    "protobufjs": "^5.0.3"
+    "protobufjs": "^6.8.6"
   },
   "devDependencies": {
     "body-parser": "^1.15.2",


murgatroid99:
This is a major version bump that breaks the API. It will not work. That is one of the primary reasons `@grpc/proto-loader` exists in the first place: to use the new major version of Protobuf.js.

PR #828, opened 2019-04-15
Require @types/protobuf.js at version 5.x as a dependency

### LINKED ISSUE #572: Cannot compile TypeScript hello world client/server

jdforsythe:
Fixes #572

### DIFF
--- a/packages/grpc-native-core/package.json
+++ b/packages/grpc-native-core/package.json
@@ -29,6 +29,7 @@
     "node-pre-gyp"
   ],
   "dependencies": {
+    "@types/protobufjs": "^5.0.31",
     "lodash.camelcase": "^4.3.0",
     "lodash.clone": "^4.5.0",
     "nan": "^2.0.0",


murgatroid99:
Would you mind making the same PR against the `grpc@1.20.x` branch? That would help us get it out more quickly.

Ten0:
The following seems to have broken our own system. We were using protobufjs 6 for generating the types, and independently using grpc for making the actual requests. Since the latest gRPC update, we couldn't generate the types because compilation was using the wrong protobufjs version (5).
We were getting the following kind of errors:

```
@types/protobufjs/5.0.31/node_modules/@types/protobufjs/index.d.ts"' has no exported member 'RPCImpl'
```

```
n/node_modules/.registry.npmjs.org/@types/protobufjs/5.0.31/node_modules/@types/protobufjs/index.d.ts"' has no exported member 'IParseOptions'.

51 export declare type Options = Protobuf.IParseOptions & Protobuf.IConversionOptions & {
```

Locking gRPC to 1.20 solved it but I'm unsure that's the right way to solve it. Any insight ?
We're using pnpm by the way.

PR #830, opened 2019-04-17
Bump to v1.20.0

murgatroid99:
I need to write something here.


brandonros:
What else needs to be done to get 1.20.0 to npm?

nicolasnoble:
Building packages, and publishing. Pressing buttons, that is.

brandonros:
@nicolasnoble Thanks for the swift response. Am I able to push those buttons? If not, are those buttons are expected to be pushed by the end of day today?

murgatroid99:
Currently this is something only we can do, and I have done it. v1.20.0 is now out.

brandonros:
Thank you so much!

PR #832, opened 2019-04-21
grpc-js: support unary and server streaming rpcs

cjihrig:
This commit adds support for unary and server streaming RPCs.

Server and bidi streaming will be added, but first I'd like to extract [this logic](https://github.com/grpc/grpc-node/blob/ad9eb795d6a194716f95ce3c1d70e5bc78b60f7d/packages/grpc-js/src/call-stream.ts#L294-L351) to make it reusable by the server.

### DIFF
--- a/packages/grpc-js/src/server-call.ts
+++ b/packages/grpc-js/src/server-call.ts
@@ -16,20 +16,74 @@
  */
 
 import {EventEmitter} from 'events';
+import * as http2 from 'http2';
 import {Duplex, Readable, Writable} from 'stream';
+
 import {ServiceError} from './call';
+import {Status} from './constants';
 import {Deserialize, Serialize} from './make-client';
 import {Metadata} from './metadata';
 
+function noop(): void {}
 
-export class ServerUnaryCall<RequestType> extends EventEmitter {
+export type PartialServiceError = Partial<ServiceError>;
+
+type DeadlineUnitIndexSignature = {
+  [name: string]: number
+};
+
+const GRPC_ACCEPT_ENCODING_HEADER = 'grpc-accept-encoding';
+const GRPC_ENCODING_HEADER = 'grpc-encoding';
+const GRPC_MESSAGE_HEADER = 'grpc-message';
+const GRPC_STATUS_HEADER = 'grpc-status';
+const GRPC_TIMEOUT_HEADER = 'grpc-timeout';
+const DEADLINE_REGEX = /(\d{1,8})\s*([HMSmun])/;
+const deadlineUnitsToMs: DeadlineUnitIndexSignature = {
+  H: 3600000,
+  M: 60000,
+  S: 1000,
+  m: 1,
+  u: 0.001,
+  n: 0.000001
+};
+const defaultResponseHeaders = {
+  // TODO(cjihrig): Remove these encoding headers from the default response
+  // once compression is integrated.
+  [GRPC_ACCEPT_ENCODING_HEADER]: 'identity',
+  [GRPC_ENCODING_HEADER]: 'identity',
+  [http2.constants.HTTP2_HEADER_STATUS]: http2.constants.HTTP_STATUS_OK,
+  [http2.constants.HTTP2_HEADER_CONTENT_TYPE]: 'application/grpc+proto'
+};
+const defaultResponseOptions = {
+  waitForTrailers: true
+} as http2.ServerStreamResponseOptions;
+
+
+export type ServerSurfaceCall = {
+  cancelled: boolean; getPeer(): string;
+  sendMetadata(responseMetadata: Metadata): void
+};
+
+export type ServerUnaryCall<RequestType, ResponseType> =
+    ServerSurfaceCall&{request: RequestType | null};
+export type ServerReadableStream<RequestType, ResponseType> =
+    ServerSurfaceCall&Readable;
+export type ServerWritableStream<RequestType, ResponseType> =
+    ServerSurfaceCall&Writable&{request: RequestType | null};
+export type ServerDuplexStream<RequestType, ResponseType> =
+    ServerSurfaceCall&Duplex;
+
+export class ServerUnaryCallImpl<RequestType, ResponseType> extends EventEmitter
+    implements ServerUnaryCall<RequestType, ResponseType> {
   cancelled: boolean;
   request: RequestType|null;
 
-  constructor(private call: ServerCall, public metadata: Metadata) {
+  constructor(
+      private call: Http2ServerCallStream<RequestType, ResponseType>,
+      public metadata: Metadata) {
     super();
     this.cancelled = false;
-    this.request = null;  // TODO(cjihrig): Read the unary request here.
+    this.request = null;
   }
 
   getPeer(): string {
@@ -37,17 +91,20 @@ export class ServerUnaryCall<RequestType> extends EventEmitter {
   }
 
   sendMetadata(responseMetadata: Metadata): void {
-    throw new Error('not implemented yet');
+    this.call.sendMetadata(responseMetadata);
   }
 }
 
 
-export class ServerReadableStream<RequestType> extends Readable {
+export class ServerReadableStreamImpl<RequestType, ResponseType> extends
+    Readable implements ServerReadableStream<RequestType, ResponseType> {
   cancelled: boolean;
+  private done = false;
 
   constructor(
-      private call: ServerCall, public metadata: Metadata,
-      private deserialize: Deserialize<RequestType>) {
+      private call: Http2ServerCallStream<RequestType, ResponseType>,
+      public metadata: Metadata,
+      private _deserialize: Deserialize<RequestType>) {
     super();
     this.cancelled = false;
   }
@@ -57,21 +114,32 @@ export class ServerReadableStream<RequestType> extends Readable {
   }
 
   sendMetadata(responseMetadata: Metadata): void {
-    throw new Error('not implemented yet');
+    this.call.sendMetadata(responseMetadata);
+  }
+
+  _done(): void {
+    this.done = true;
+    this.on('data', noop);
   }
 }
 
 
-export class ServerWritableStream<RequestType, ResponseType> extends Writable {
+export class ServerWritableStreamImpl<RequestType, ResponseType> extends
+    Writable implements ServerWritableStream<RequestType, ResponseType> {
   cancelled: boolean;
   request: RequestType|null;
 
   constructor(
-      private call: ServerCall, public metadata: Metadata,
-      private serialize: Serialize<ResponseType>) {
-    super();
+      private call: Http2ServerCallStream<RequestType, ResponseType>,
+      public metadata: Metadata, private _serialize: Serialize<ResponseType>) {
+    super({objectMode: true});
     this.cancelled = false;
-    this.request = null;  // TODO(cjihrig): Read the unary request here.
+    this.request = null;
+
+    this.on('error', (err) => {
+      this.call.sendError(err as ServiceError);
+      this.end();
+    });
   }
 
   getPeer(): string {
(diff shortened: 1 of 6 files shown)


cjihrig:
Hm, looking at the CI, I think this might require updating the supported semver range to `^8.13.0 || >=10.0.0` or `^8.13.0 || >=10.10.0`. `8.13.0` puts `http2` out of experimental. The `9.x` branch has been end-of-life for almost 10 months now, and should be dropped if possible. I think all versions from 10 and up should be fine, but `10.10.0` is when `http2` came out of experimental.

packages/grpc-js/src/server-call.ts
```diff
@@ -146,3 +246,209 @@ export type Handler<RequestType, ResponseType> = {
...
+      const metadata = this.status.metadata;
+
+      if (metadata) {
+        trailersToSend =
+            Object.assign(trailersToSend, metadata.toHttp2Headers());
+      }
+
+      this.stream.sendTrailers(trailersToSend);
+    });
+    this.stream.respond(headers, defaultResponseOptions);
```
murgatroid99: Why is this line after the `'wantTrailers'` event listener? That makes it easy to miss that this line is in the `sendMetadata` scope.

packages/grpc-js/src/server-call.ts
```diff
@@ -16,38 +16,89 @@
...
+export type PartialServiceError = Partial<ServiceError>;
+
+type DeadlineUnitIndexSignature = {
+  [name: string]: number
+};
+
+const kGrpcMessageHeader = 'grpc-message';
+const kGrpcStatusHeader = 'grpc-status';
+const kGrpcTimeoutHeader = 'grpc-timeout';
+const kDeadlineRegex = /(\d{1,8})\s*([HMSmun])/;
```
murgatroid99: These `k`-prefixed constants don't match the coding style of the rest of the codebase. You should instead use `ALL_CAPS_SNAKE_CASE`.

packages/grpc-js/src/server-call.ts
```diff
@@ -57,40 +108,93 @@ export class ServerReadableStream<RequestType> extends Readable {
...
-    throw new Error('not implemented yet');
+    this.call.sendMetadata(responseMetadata);
+  }
+
+  async _write(chunk: ResponseType, encoding: string, callback: Function) {
+    try {
+      this.outstandingWrites++;
+      const response = await this.call.serializeMessage(chunk);
+      this.call.write(response);
+      this.outstandingWrites--;
```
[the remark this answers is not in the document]
cjihrig: Regarding `outstandingWrites` tracking, an issue that I encountered is calling `stream.write()` immediately followed by `stream.end()` with non-identity compression (for example in a server streaming RPC). The other compression schemes use Node's asynchronous `zlib` calls. Without the tracking, there is a window between `serializeMessage()` finishing and the underlying stream's `write()` method being called. During that window, `end()` is called. When the `write()` to the stream finally happens, you get a 'write after end' error.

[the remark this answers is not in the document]
murgatroid99: The problem there is with the `end()` implementation. It should do the minimum work to handle, then call `super.end()`. The call to `this.call.end()` and the rest of the logic should be in an `_final` method. The stream code should delay the `_final` invocation until all writes are complete.

packages/grpc-js/src/server-call.ts
```diff
@@ -146,3 +246,209 @@ export type Handler<RequestType, ResponseType> = {
...
+  sendMetadata(customMetadata?: Metadata) {
+    if (this.stream.headersSent === true) {
+      return;
+    }
+
+    const custom = customMetadata ? customMetadata.toHttp2Headers() : null;
+    // TODO(cjihrig): Include compression headers.
+    const headers = Object.assign(defaultResponseHeaders, custom);
+
+    this.stream.once('wantTrailers', () => {
```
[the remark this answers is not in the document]
cjihrig: I don't think that we need to do any additional tracking here. The `'wantTrailers'` event comes from calling `this.stream.respond()` in this function (`sendMetadata()`), and every response path goes through this. That's also why I had originally written the code the way I did in https://github.com/grpc/grpc-node/pull/832#discussion_r277436868.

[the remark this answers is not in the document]
murgatroid99: On the application side, sending metadata and sending trailers are separate actions, at least in server streaming and bidirectional streaming methods. Sending metadata does not imply that status has been set.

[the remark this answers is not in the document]
cjihrig: Just to clarify, can you provide an example server streaming handler that does what you're referring to? I'd like to add a test for the behavior.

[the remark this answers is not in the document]
murgatroid99: After rereading this code and the documentation for this event I think this is OK.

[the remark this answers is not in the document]
cjihrig: Cool. In that case, I think I've addressed all of your comments.

packages/grpc-js/src/server-call.ts
```diff
@@ -146,3 +252,213 @@ export type Handler<RequestType, ResponseType> = {
...
+        this.status.metadata = metadata;
+      }
+
+      this.end(response);
+    } catch (err) {
+      this.sendError(err, Status.INTERNAL);
+    }
+  }
+
+  sendError(error: ServiceError, code = Status.UNKNOWN) {
```
murgatroid99: Instead of this I suggest the following:

```ts
sendStatus(statusObj: StatusObject) {
    if (this.cancelled === true) {
      return;
    }

     if (this.deadline !== null) {
      clearTimeout(this.deadline);
      this.deadline = null;
    }
  this.sendMetadata();
  this.stream.once('wantTrailers', () => {
      let trailersToSend = {
        [GRPC_STATUS_HEADER]: status.code,
        [GRPC_MESSAGE_HEADER]: encodeURI(status.details as string)
      };
      const metadata = status.metadata;

       if (metadata) {
        trailersToSend =
            Object.assign(trailersToSend, metadata.toHttp2Headers());
      }

       this.stream.sendTrailers(trailersToSend);
    });
  this.stream.end();
}
```

Then `end` should be removed, and the trailer code should be removed from `sendMetadata`. The code to translate a `ServiceError` to a `StatusObject` would have to be pushed up a layer, but this would significantly simplify the logic of ending a response

cjihrig: Renaming this and switching to a `StatusObject` input shouldn't be too big of a deal, but what about the following:

> Then `end` should be removed

The server and bidi streams still have to signal down to the HTTP2 layer to end the call. The deadline timer is also cleared there, along with any future cleanup code that might be added. I imagine this would result in the code being duplicated, or still being contained in a function, just with a name other than `end`.

> the trailer code should be removed from `sendMetadata`

Don't we still want this behavior in places other than `sendStatus()` (for example `sendUnaryMessage()`)?

murgatroid99: The idea here is that the only path to `stream.end()` would be going through `sendStatus()`. I included the existing code from `end()` in my suggested code. I think a reasonable alternative would be to keep `end` with the cleanup code, but make it private and call it only from here. And the idea is that `sendUnaryMessage()` would also call `sendStatus()` at the end; it's already calling `sendError()` conditionally a couple of times, it could simplify the code to instead call `sendStatus()` unconditionally, perhaps with a default simple status.

The point of this suggestion is to ensure that any code path for ending a call goes through setting some status, and that setting the trailers goes through the same path, because it is essentially the same action.

cjihrig: Thanks for the explanation. I've updated the PR with `sendStatus()`. I kept `sendError()` as a helper for translating `ServiceError`s before calling `sendStatus()`.

cjihrig:
@murgatroid99 is this ready for a CI run?

murgatroid99:
Yeah, I didn't realize it wasn't running.

packages/grpc-js/src/server-call.ts
```diff
@@ -146,3 +253,218 @@ export type Handler<RequestType, ResponseType> = {
...
+      this.sendStatus(
+          {code: Status.OK, details: 'OK', metadata} as StatusObject);
+    } catch (err) {
+      err.code = Status.INTERNAL;
+      this.sendError(err);
+    }
+  }
+
+  sendStatus(statusObj: StatusObject) {
+    if (this.cancelled === true) {
```
murgatroid99: `this.cancelled` is a boolean. You don't need to explicitly compare it with `true`.

packages/grpc-js/src/server-call.ts
```diff
@@ -146,3 +253,218 @@ export type Handler<RequestType, ResponseType> = {
...
+      if (error.hasOwnProperty('details')) {
+        status.details = error.details;
+      }
+    }
+
+    this.sendStatus(status);
+  }
+
+  write(chunk: Buffer) {
+    if (this.cancelled === true) {
```
murgatroid99: Again, don't need to do an explicit comparison here.

packages/grpc-js/src/server-call.ts
```diff
@@ -146,3 +253,218 @@ export type Handler<RequestType, ResponseType> = {
...
+      return;
+    }
+
+    this.sendMetadata();
+    return this.stream.write(chunk);
```
murgatroid99: This propagates the return value of `this.stream.write`, but it looks like it doesn't propagate the `'drain'` event, which the next layer up listens for.

packages/grpc-js/src/server.ts
```diff
@@ -227,4 +229,108 @@ export class Server {
...
+  private _setupHandlers(): void {
+    if (this.http2Server === null) {
+      return;
+    }
+
+    this.http2Server.on(
+        'stream',
+        (stream: http2.ServerHttp2Stream,
+         headers: http2.IncomingHttpHeaders) => {
+          if (this.started !== true) {
```
murgatroid99: `this.started` is a boolean. You don't have to explicitly compare it with `true`.

nicolasnoble: This seems to be a common pattern with the patch.

packages/grpc-js/src/server-call.ts
```diff
@@ -133,9 +240,9 @@ export type handleBidiStreamingCall<RequestType, ResponseType> =
     (call: ServerDuplexStream<RequestType, ResponseType>) => void;
 
 export type HandleCall<RequestType, ResponseType> =
-    handleUnaryCall<RequestType, ResponseType>|
-    handleClientStreamingCall<RequestType, ResponseType>|
-    handleServerStreamingCall<RequestType, ResponseType>|
+    handleUnaryCall<RequestType, ResponseType>&
```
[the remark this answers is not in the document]
cjihrig: It seemed wrong to me as well. When I change it to `|`, I get the following errors. I'm assuming it's a lack of TypeScript knowledge on my part, but it's also not clear why `&` does work:

```
src/server.ts:299:7 - error TS2345: Argument of type 'ServerUnaryCallImpl<RequestType, ResponseType>' is not assignable to parameter of type 'ServerSurfaceCall & { request: RequestType | null; } & Readable & Writable & { request: RequestType | null; } & Duplex'.
  Type 'ServerUnaryCallImpl<RequestType, ResponseType>' is missing the following properties from type 'Readable': readable, readableHighWaterMark, readableLength, _read, and 13 more.

299       emitter,
          ~~~~~~~

src/server.ts:328:3 - error TS2554: Expected 2 arguments, but got 1.

328   handler.func(stream);
      ~~~~~~~~~~~~~~~~~~~~

  src/server-call.ts:224:6
    224      callback: sendUnaryData<ResponseType>) => void;
             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    An argument for 'callback' was not provided.
```

cjihrig:
I've addressed all of the comments other than https://github.com/grpc/grpc-node/pull/832#discussion_r280637935

packages/grpc-js/src/server-call.ts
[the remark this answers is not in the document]
murgatroid99: I see. This is a problem with [function type contravariance with argument types](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)#Function_types). When you declare that a `handleCall` function can be any of those functions, then when you call a function with that type, Typescript wants to make sure that the argument(s) are valid no matter which function it actually is. That's not how we want it to actually work.

The solution is to have specific `UnaryHandler`, `ServerStreamingHandler`, etc. types, and have the `handleUnary`, `handleServerStreaming`, etc. functions in `server.ts` take those specific `Handler` types as arguments. Each of those handler types should specify `func` as the specific `HandleCall` type and `type` as the specific corresponding string. Also, they should be declared with `export interface`, not `export type`. Then `Handler` should just be each of those or-ed together, just like `HandleCall`.

Then, in theory, the switch statement in `server.ts` should be enough for Typescript to infer which `Handler` type it has. If not, TypeScript will throw errors for the 4 function calls in the switch statement; in that case it is appropriate to use type assertions like `as UnaryHandler<RequestType, ResponseType>`.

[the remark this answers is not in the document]
cjihrig: I've pushed up another commit ¯\_(ツ)_/¯

SProst:
@cjihrig I have a question regarding this PR -- does this enable using the server in addition to the client in this library? If so, it would appear that the server module export is still throwing a "Not yet implemented" error in index.js. If I have misunderstood the purpose of this PR, my apologies.

cjihrig:
@SProst the server is intentionally not exposed yet, as only the unary and server streaming RPC types have been ported over. If you still want to access the incomplete server implementation anyway, you can technically do so via `require('@grpc/grpc-js/build/src/server')`

SProst:
@cjihrig Took me too long to figure that out today, but I was able to actually get things working with your https://github.com/cjihrig/grpc-server-js module. I appreciate the clarification.

PR #833, opened 2019-04-23
Adding node 12.

nicolasnoble:
Let's do this as 1.20.1 - this is early enough that we want to add this.


nicolasnoble:
Note that the main repository also has fixes we may want, so this wouldn't be the only 1.20.1 change.

nicolasnoble:
The errors are a chicken-and-egg problem, merging anyway.

PR #841, opened 2019-04-26
grpc-js: don't export private symbol

cjihrig:
The `kChannel` symbol was being exported due to a TypeScript issue. That issue has been resolved, so it seems OK to remove the export.

While I'm here, should I rename `kChannel` to something that follows the project's style (per https://github.com/grpc/grpc-node/pull/832#discussion_r277438065)?


murgatroid99:
A rename would probably be good.

cjihrig:
The full test suite passed (minus the Mergeable check because the commit message contains the string "exp"). Pushed another commit for the rename.

PR #842, opened 2019-04-26
Revert "Protobufjs typescript types backport"

murgatroid99:
Reverts grpc/grpc-node#838. It looks like this caused more problems than it solved.


murgatroid99:
This needs to get out quickly, and the repository was in an OK state before #838.

PR #843, opened 2019-04-26
grpc-js: require non-experimental http2

cjihrig:
This commit bumps the required semver range to versions of Node that include a non-experimental http2 module.


murgatroid99:
It looks like there is still the old version range somewhere because it tried to run some tests under 9.x

cjihrig:
Found a couple other places that didn't exactly match the previous semver range string. Hopefully that's all of them.

cjihrig:
The break was intentional, but is also the reason that the version was bumped from 0.3.x to 0.4.0. Node 8.10 wasn't supported prior to this change though:

```
> process.version
'v8.10.0'
> require('@grpc/grpc-js');
Error: @grpc/grpc-js only works on Node ^8.11.2 || >=9.4
```

PR #848, opened 2019-05-01
Build Electron 5 artifacts

murgatroid99:
Apparently that's out now


nicolasnoble:
There's a codesmell from Electron here. They are forcibly defining `OPENSSL_THREAD` from the build command line somehow. That's invalid, especially if they are not providing symbols for OpenSSL.

nicolasnoble:
Right, the Windows tests are passing. I have a strong hint as to what's going on here.

PR #859, opened 2019-05-07
grpc-js: add client streaming RPC support

cjihrig:
This commit adds client streaming RPC support.

### DIFF
--- a/packages/grpc-js/src/server-call.ts
+++ b/packages/grpc-js/src/server-call.ts
@@ -24,6 +24,7 @@ import {StatusObject} from './call-stream';
 import {Status} from './constants';
 import {Deserialize, Serialize} from './make-client';
 import {Metadata} from './metadata';
+import {StreamDecoder} from './stream-decoder';
 
 function noop(): void {}
 
@@ -100,6 +101,7 @@ export class ServerUnaryCallImpl<RequestType, ResponseType> extends EventEmitter
 export class ServerReadableStreamImpl<RequestType, ResponseType> extends
     Readable implements ServerReadableStream<RequestType, ResponseType> {
   cancelled: boolean;
+  private decoder: StreamDecoder;
 
   constructor(
       private call: Http2ServerCallStream<RequestType, ResponseType>,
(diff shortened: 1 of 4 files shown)


packages/grpc-js/src/server-call.ts
```diff
@@ -100,13 +101,52 @@ export class ServerUnaryCallImpl<RequestType, ResponseType> extends EventEmitter
...
 
   constructor(
       private call: Http2ServerCallStream<RequestType, ResponseType>,
       public metadata: Metadata,
       private _deserialize: Deserialize<RequestType>) {
     super({objectMode: true});
     this.cancelled = false;
+    this.decoder = new StreamDecoder();
+
+    const http2Stream = this.call._getHttp2Stream();
```
murgatroid99: This really feels like an abstraction layering violation. The `Http2ServerCallStream` owns the http2 stream object, and all interaction with the http2 stream should go through it.

cjihrig: I'm open to suggestions. I can think of a few other ways of doing this, but I'd like to avoid guessing wrong and rewriting it multiple times.

nicolasnoble: So, I don't think that @murgatroid99 was hinting at a particular way of writing this. The gist of his remark is that if we look at the current structure of the code, you have this class you wrote a little while back named `Http2ServerCallStream` with a comment on it that says it's wrapping the HTTP2 request.

But now, in this new pull request, you're bypassing this wrapper, and you are accessing the `ServerHttp2Stream` object directly. So, from an overall design point of view, this seems inconsistent. On one hand, you're wrapping the features of `ServerHttp2Stream`, mutating some of this class' behavior such as the `write` function, and on the other hand, you're bypassing these mutations. Thinking about the future maintainer of this code, one needs precise rules and contracts about what's being wrapped and why.

If you're willing to try and bypass the wrapping you introduced yourself, would you mind sharing your thoughts on what the contracts here should be?

murgatroid99: Directly listening to the `'data'` event won't work well because the stream classes have special handling around that event: they will start reading as soon as you add that listener.

The strategy used on the client side is for the internal call stream class to be a duplex stream of `Buffer` messages, and when the surface call needs to send or receive a unary message it just writes or reads only one message. I think it would be reasonable to do that here too.

Another option I can see is to have all of the `Server*Impl` classes directly own and interact with the `ServerHttp2Stream`, with helper functions or a helper class for de-duplicating some code.

packages/grpc-js/src/server-call.ts
```diff
@@ -107,6 +108,19 @@ export class ServerReadableStreamImpl<RequestType, ResponseType> extends
...
     this.cancelled = false;
+    this.call.setupReadable(this);
+  }
+
+  _read(size: number) {
+    this.call.resume();
+  }
+
+  deserialize(input: Buffer): RequestType|null {
+    if (input === null || input === undefined) {
```
murgatroid99: The function type declaration says this will never be null or undefined. And this is intended to only be called by our own internal code. And the code that calls it explicitly checks for null first. So this whole condition should be unnecessary.

cjihrig: OK. I've removed it.

PR #860, opened 2019-05-08
Make Windows build use OpenSSL built in to Node

murgatroid99:
Maybe this will just work.

EDIT: No such luck


murgatroid99:
I don't think this is going to work, because I don't think nodejs/node#4932 was ever actually fixed.

nicolasnoble:
It might be worth re-opening it somehow.

nicolasnoble:
By the way, I meant re-opening the nodejs bug :-)

murgatroid99:
See the latest update on that bug for why I'm trying this.

murgatroid99:
OK, based on further updates to that issue it looks like this will not be a viable option in the near future.

PR #866, opened 2019-05-11
feat: export package version

alexander-fenster:
Since we are entering a phase when different client libraries will use different gRPC implementations (most will be migrated to `@grpc/grpc-js` but some will still stay with `grpc`), we'd like to have an easy way for the client library to set the gRPC version in the headers without actually knowing if the `grpc` instance they work with is a C-core or JS implementation.

Can we have this `grpc.version` added to both implementations to make this tracking easier? (or feel free to suggest another options!)

### DIFF
--- a/packages/grpc-js/src/index.ts
+++ b/packages/grpc-js/src/index.ts
@@ -250,3 +250,7 @@ export const InterceptingCall = () => {
 };
 
 export {GrpcObject} from './make-client';
+
+const packageJson = require('../../package.json');
+export const version =
+    `${packageJson.name.replace(/.*\//, '')}/${packageJson.version}`;
--- a/packages/grpc-native-core/index.d.ts
+++ b/packages/grpc-native-core/index.d.ts
@@ -947,6 +947,11 @@ declare module "grpc" {
   };
 
   /**
+   * Package version.
+   */
+  export const version: string;
+
+  /**
    * Metadata generator function.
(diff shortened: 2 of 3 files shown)


murgatroid99:
First, both libraries put some variant of this information in the `user-agent` header, or at least they're supposed to. Does that resolve your problem of wanting that information in a header?

Second, library users can already do `require('grpc/package.json')` or `require('@grpc/grpc-js/package.json')` and then format the information from that however they want.

alexander-fenster:
Libraries use `grpc/${version}` header. @JustinBeckwith can know more about tracking and if `user-agent` header is sufficient. Justin, do you know if can we go with `user-agent` and not set `grpc` at all?

As for the second item, it won't be possible since the gax code can accept any `grpc` instance from the caller and it does not know if it's one or another implementation.

murgatroid99:
If the current `user-agent` behavior does not solve your problem, this change needs to go through our [API design review process](https://github.com/grpc/proposal) because it is a (non-breaking) API change.

murgatroid99:
I don't think it's a good idea to export this information pre-formatted for a specific use-case. I would prefer to either export the whole `package.json` as a single object or export `name` and `version` verbatim.

alexander-fenster:
I'm totally fine with either of those options - I just wanted to have the conversation started :)

PR #874, opened 2019-05-17
grpc-js: update to gts@1.x.x

cjihrig:
This commit updates the `gts` dependency to 1.x.x. All of the changes come from running `npm run fix` and find+replace'ing a handful of `assert` methods, as instructed by the tool.

### DIFF
--- a/packages/grpc-js/gulpfile.ts
+++ b/packages/grpc-js/gulpfile.ts
@@ -17,7 +17,6 @@
 
 import * as gulp from 'gulp';
 
-import * as fs from 'fs';
 import * as mocha from 'gulp-mocha';
 import * as path from 'path';
 import * as execa from 'execa';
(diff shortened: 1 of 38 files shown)


cjihrig:
If I'm reading the CI output properly, it looks like it's failing on Node 6 due to the presence of an `async` function in `gts`. I guess this will require sticking on an older `gts` version, or dropping support for Node 6, which is already end-of-life.

murgatroid99:
We're not going to drop Node 6 support on the main grpc library because the latest version of the formatter for `grpc-js` doesn't run on Node 6. Probably the right thing here is to remove that `gts clean` command from the overall `gulp cleanAll` script.

cjihrig:
> We're not going to drop Node 6 support on the main grpc library because the latest version of the formatter for grpc-js doesn't run on Node 6.

Sorry, I meant drop Node 6 specifically for grpc-js. Is that possible, given that the module doesn't even run on Node 6?

murgatroid99:
The grpc-js tests already don't run on Node 6. What's happening here is that we have a collection of setup scripts run as `gulp setup` and a collection of teardown scripts run as `gulp cleanAll` that are run before and after the full collection of test suites for each version of Node. The test scripts check for specific versions before running, but the setup and teardown scripts don't, which is why this is failing. I think the right solution is to simply remove that teardown script for `grpc-js`, because it's probably not really necessary.

cjihrig:
Would you be open to something like https://github.com/grpc/grpc-node/pull/874/commits/4e0162cf222318e1123f239a21267074a7a7475d (if it works), which skips all of the tasks on unsupported versions of Node?

murgatroid99:
If this works it's fine but I was thinking that a simpler solution for this specific problem would be to just remove the `grpc-js` `cleanAll` task from the global `cleanAll` task.

PR #887, opened 2019-05-28
grpc-js: Change how filters access connectivity information

murgatroid99:
Instead of having filters explicitly use the channel's `connectivityStateChanged` listener to track connectivity state, make the filter stack's initial metadata promise only resolve when the channel becomes connected. Then any filter can just await that promise to wait for the channel to connect. This should fix the warning reported in https://github.com/grpc/grpc-node/issues/694#issuecomment-496566860.


murgatroid99:
The person who originally reported the issue says that it fixes that problem. I'm treating that as a review.

PR #897, opened 2019-05-31
grpc-js: load semver range from package.json

cjihrig:
This commit loads the required semver range from the `package.json` file, instead of hard-coding the string in `index.ts`.

This should make it a bit easier to update the required version in the future. After this change, the version string will exist in:

- This module's `package.json` file. Ideally, this would be the only non-docs location.
- The top level `PACKAGE-COMPARISON.md` file.
- The top level's `test/gulpfile.js`. I'm happy to update that as well if it's OK.


cjihrig:
It looks like the CI never ran.

PR #899, opened 2019-06-03
Add test for not waiting for ready by default

murgatroid99:
This is a test that was failing from #796 that doesn't directly test to that change

### DIFF
--- a/packages/grpc-native-core/test/surface_test.js
+++ b/packages/grpc-native-core/test/surface_test.js
@@ -941,6 +941,18 @@ describe('Other conditions', function() {
     call.write({});
     call.end();
   });
+  it('client should not wait for ready by default', function(done) {
+    this.timeout(15000);
+    const disconnectedClient = new Client('foo.test.google.com:50051', grpc.credentials.createInsecure());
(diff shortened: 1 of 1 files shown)


packages/grpc-native-core/test/surface_test.js
```diff
@@ -941,6 +941,18 @@ describe('Other conditions', function() {
     call.write({});
     call.end();
   });
+  it('client should not wait for ready by default', function(done) {
+    this.timeout(15000);
+    const disconnectedClient = new Client('foo.test.google.com:50051', grpc.credentials.createInsecure());
+    const metadata = new grpc.Metadata({waitForReady: false});
```
imjoey: If understanding correctly, I'm a bit of confused by the `{waitForReady: false}` argument. Shall we remove it for the **by default** test case? Thanks.

murgatroid99: I think you're right. It looks like that test implementation is the result of some evolution in the PR it was copied from.

murgatroid99:
Fortunately, this will probably work now because we no longer test on the Node version where this was failing.

PR #901, opened 2019-06-04
Fix dropped messages when multiple arrived in one HTTP/2 frame

### LINKED ISSUE #895: v0.4.1 closes streams prematurely
Original issue: https://github.com/googleapis/nodejs-firestore/issues/667

Firestore relies on grpc-js for all its operations. With the recent 0.4.1 release, our RunQuery operations terminate before the client receives all results. We use a unidirectional stream and receive a seemingly random number of events before the `on('end')` callback fires. With 15 messages, I was never able to receive all messeages before the stream closed. We seem to receive typically between 8 and 9 responses.

The RunQuery operation has a deadline of 60s, but at least the query doesn't last 60 seconds. It might be that the channel exists for 60 seconds when we receive the  `on('end')` event, but I no longer have v0.4.1 to reproduce.
 
I was able to consistently reproduce the above issue using v0.4.1 of grpc-js. This issue does not exist in v0.4.0. The code snippet in the link above is enough to repro.

murgatroid99:
This fixes #895. This depends on specific HTTP/2 framing behavior so I don't know how to turn it into a unit test, but I suspect that this bug was also the cause of grpc/grpc#18967. If so that can act as a regression test.

CC @cjihrig.

### DIFF
--- a/packages/grpc-js/src/call-stream.ts
+++ b/packages/grpc-js/src/call-stream.ts
@@ -289,9 +289,9 @@ export class Http2CallStream extends Duplex implements Call {
       });
       stream.on('trailers', this.handleTrailers.bind(this));
       stream.on('data', (data: Buffer) => {
-        const message = this.decoder.write(data);
+        const messages = this.decoder.write(data);
 
-        if (message !== null) {
+        for (const message of messages) {
           this.tryPush(message);
         }
       });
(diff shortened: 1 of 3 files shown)


cjihrig:
Thank you for the heads up.

PR #908, opened 2019-06-06
Pure JS: Fixed two bugs with goaway handling

### LINKED ISSUE #734: Error: Getting metadata from plugin failed with error: New streams cannot be created after receiving a GOAWAY
### Problem description
We currently use `@grpc/grpc-js` with Envoy in production, and we get this mystifying error `Error: Getting metadata from plugin failed with error: New streams cannot be created after receiving a GOAWAY` occasionally during normal RPCs. We currently only use unary calls and not streams. It might have to do with something when we do a new deployment. Does anyone have any insight on how to debug this error?

### Reproduction steps
It's difficult to reproduce because this doesn't happen in our staging environment.

### Environment
 - alpine linux 3.6
 - Node version 10.13
 - `@grpc/grpc-js` 0.3.2

### Additional context
```
Error: Getting metadata from plugin failed with error: New streams cannot be created after receiving a GOAWAY
    at Http2CallStream.call.on (/home/node/app/node_modules/@grpc/grpc-js/build/src/client.js:88:45)
    at Http2CallStream.emit (events.js:187:15)
    at Http2CallStream.EventEmitter.emit (domain.js:442:20)
    at Http2CallStream.endCall (/home/node/app/node_modules/@grpc/grpc-js/build/src/call-stream.js:66:18)
    at Http2CallStream.<anonymous> (/home/node/app/node_modules/@grpc/grpc-js/build/src/call-stream.js:337:18)
    at Generator.next (<anonymous>)
    at fulfilled (/home/node/app/node_modules/@grpc/grpc-js/build/src/call-stream.js:4:58)
```

murgatroid99:
First, I think this fixes #734. When a Node http2 client session receives a GOAWAY from the server, it sets its closed flag and emits a `'goaway'` event immediately, and then emits the `'close'` and `'error'` (if needed) events a tick or two later. We listen for the `'close'` event to know when a session should be considered unusable, and if the channel tries to start a stream in between the `'goaway'` event and the `'close'` event, it ends with an error. The fix is to simply stop using the subchannel as soon is the `'goaway'` event comes in.

Second, RST_STREAM error codes were not getting handled correctly in the stream code, including artificial RST_STREAM errors created by the Node http2 code when a GOAWAY is received.

### DIFF
--- a/packages/grpc-js/src/call-stream.ts
+++ b/packages/grpc-js/src/call-stream.ts
@@ -298,10 +298,10 @@ export class Http2CallStream extends Duplex implements Call {
       stream.on('end', () => {
         this.tryPush(null);
       });
-      stream.on('close', async errorCode => {
+      stream.on('close', async () => {
         let code: Status;
         let details = '';
-        switch (errorCode) {
+        switch (stream.rstCode) {
           case http2.constants.NGHTTP2_REFUSED_STREAM:
             code = Status.UNAVAILABLE;
             break;
@@ -329,11 +329,9 @@ export class Http2CallStream extends Duplex implements Call {
         this.endCall({ code, details, metadata: new Metadata() });
       });
       stream.on('error', (err: Error) => {
-        this.endCall({
-          code: Status.INTERNAL,
-          details: 'Internal HTTP2 error',
-          metadata: new Metadata(),
-        });
+        /* We need an error handler here to stop "Uncaught Error" exceptions
+         * from bubbling up. However, errors here should all correspond to
+         * "close" events, where we will handle the error more granularly */
       });
       if (!this.pendingRead) {
         stream.pause();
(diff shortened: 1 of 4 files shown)


packages/grpc-js/src/call-stream.ts
```diff
@@ -130,7 +130,16 @@ export class Http2CallStream extends Duplex implements Call {
   private endCall(status: StatusObject): void {
     if (this.finalStatus === null) {
       this.finalStatus = status;
-      this.emit('status', status);
+      /* We do this asynchronously to ensure that no async function is in the
+       * call stack when we return control to the application. If an async
+       * function is in the call stack, any exception thrown by the application
+       * (or our tests) will bubble up and turn into promise rejection, which
+       * will result in an UnhandledPromiseRejectionWarning. Because that is
+       * a warning, the error will be effectively swallowed and execution will
```
nicolasnoble: Note this isn't a warning any longer in node 12. But yes.

PR #912, opened 2019-06-10
grpc-js: handle multiple messages in single 'data' event

cjihrig:
This commit adds support for receiving multiple messages in a single `'data'` event from the underlying HTTP2 stream. It also handles potential out of order messages due to asynchronous deserialization of messages.

### DIFF
--- a/packages/grpc-js/src/server-call.ts
+++ b/packages/grpc-js/src/server-call.ts
@@ -117,6 +117,10 @@ export class ServerReadableStreamImpl<RequestType, ResponseType>
   }
 
   _read(size: number) {
+    if (!this.call.consumeUnpushedMessages(this)) {
+      return;
+    }
+
     this.call.resume();
   }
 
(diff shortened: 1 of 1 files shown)


cjihrig:
> I think this looks correct.

I have a [test](https://github.com/cjihrig/grpc-server-js/blob/24c4be6bf81fb7f369931781ad85f3b18bd6a489/test/server.js#L603-L648) for most of the behavior here, but it does depend on an implementation detail of how messages are serialized. I can port it over if you want.

I have another question as well. What else do you think needs to be implemented before the server code can be exposed?

murgatroid99:
I'm not sure about that test. It doesn't look like it properly accounts for the grpc length prefix framing. But I think it's a good general approach to testing this.

I actually think the server might be ready to go. It seems to have all of the basic features. The next big step is to expose the API and enable the corresponding test fixtures in `test/gulpfile.ts`. If those tests pass then I'd say we're good to go.

PR #917, opened 2019-06-13
Update grpc-tools to 1.8.0

murgatroid99:
I want to publish #914.

### DIFF
--- a/packages/grpc-tools/package.json
+++ b/packages/grpc-tools/package.json
@@ -1,6 +1,6 @@
 {
   "name": "grpc-tools",
-  "version": "1.7.3",
+  "version": "1.8.0",
   "author": "Google Inc.",
   "description": "Tools for developing with gRPC on Node.js",
   "homepage": "https://grpc.io/",


murgatroid99:
That Linux test failure is just an infrastructure failure. There's no way a simple version change caused any test failures.

jedrivisser:
Does something need to be done to publish this to https://node-precompiled-binaries.grpc.io/
The latest version I see there is 1.6.6

murgatroid99:
1.7 and 1.8 are there, according to what I can see. What specific error are you experiencing?

jedrivisser:
I might not understand it right

I can download 1.6.6 from: https://node-precompiled-binaries.grpc.io/grpc-precompiled-binaries/node/grpc-tools/v1.6.6/linux-x64.tar.gz

https://node-precompiled-binaries.grpc.io/grpc-precompiled-binaries/node/grpc-tools/v1.8.0/linux-x64.tar.gz does not exist

If I go to https://node-precompiled-binaries.grpc.io and ctrl+F `grpc-tools/v1.` I can find versions up to 1.6.6

murgatroid99:
The URL pattern changed between 1.6.6 and 1.7.0. The URL for that package for 1.8.0 is http://node-precompiled-binaries.grpc.io/grpc-tools/v1.8.0/linux-x64.tar.gz. The purpose of these tarballs is for use with the `grpc-tools` npm package, which describes the URL pattern in its `package.json` file.

jedrivisser:
Thanks for pointing that out. I see the file is there. It seems it is just not indexed in the XML hosted on https://node-precompiled-binaries.grpc.io/

PR #918, opened 2019-06-13
Delete another directory in Windows build script

murgatroid99:
This directory is taken from the NodeJS build error logs in #917.


murgatroid99:
The Linux Electron Build and Windows Tests failures appear to be related to network failures when running `node-gyp`. Rerunning to see if that's just temporary.

murgatroid99:
I switched to node-gyp 4 because I saw the following problems that appeared specific to node-gyp 5: nodejs/node-gyp#1778, and some odd linkage failures on Windows that I have been unable to replicate outside of the CI environment. However, node-gyp 4 appears to have a different problem: many of these builds are failing with strange network failures that I have also been unable to replicate outside of the CI environment.

PR #921, opened 2019-06-20
Bump to 1.22.0-pre1

murgatroid99:
Bump to 1.22.0-pre1


murgatroid99:
This test that is failing was intended to test the cares resolver, and it's failing with the native resolver, like the others. I'm just going to disable it. We can re-enable those tests when we switch to the cares resolver by default.

PR #923, opened 2019-06-20
grpc-js: expose Server implementation publicly

cjihrig:
This PR exposes the grpc-js `Server` implementation as public API and enables interop testing.

### DIFF
--- a/packages/grpc-js/src/server.ts
+++ b/packages/grpc-js/src/server.ts
@@ -374,7 +374,11 @@ export class Server {
           }
         } catch (err) {
           const call = new Http2ServerCallStream(stream, null!);
-          err.code = Status.INTERNAL;
+
+          if (err.code === undefined) {
+            err.code = Status.INTERNAL;
+          }
+
           call.sendError(err);
         }
       }
--- a/packages/grpc-js/src/server.ts
+++ b/packages/grpc-js/src/server.ts
@@ -91,6 +91,7 @@ export class Server {
     string,
     UntypedHandler
   >();
+  private sessions = new Set<http2.ServerHttp2Session>();
   private started = false;
 
   constructor(options?: object) {}
@@ -223,7 +224,22 @@ export class Server {
   }
 
   forceShutdown(): void {
-    throw new Error('Not yet implemented');
+    // Close the server if it is still running.
+    if (this.http2Server && this.http2Server.listening) {
+      this.http2Server.close();
+    }
+
+    this.started = false;
+
+    // Always destroy any available sessions. It's possible that one or more
+    // tryShutdown() calls are in progress. Don't wait on them to finish.
+    this.sessions.forEach(session => {
+      // Cast NGHTTP2_CANCEL to any because TypeScript doesn't seem to
+      // recognize destroy(code) as a valid signature.
+      // tslint:disable-next-line:no-any
+      session.destroy(http2.constants.NGHTTP2_CANCEL as any);
+    });
+    this.sessions.clear();
   }
 
   register<RequestType, ResponseType>(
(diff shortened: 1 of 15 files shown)


murgatroid99:
I assume the grpc submodule modification in the first commit is unintentional. Can you undo that?

cjihrig:
It was indeed unintentional, and I'm not even sure what caused it. Any quick pointers on how to back that particular change out? I've tried a few things so far (`git submodule foreach --recursive git reset --hard`, etc.), and none of them seem to work.

murgatroid99:
The simplest way I've found to undo that is to checkout an up to date `master`, do `git submodule update --init --recursive`, then switch back to the branch and commit the new state of the submodule.

I've had that happen to me a couple of times, by not updating the submodule and then doing `git add -u`.

cjihrig:
Thanks! I believe it's fixed now.

cjihrig:
It looks like the unimplemented `bind()` and `forceShutdown()` methods are causing problems.

murgatroid99:
Right. `bind` can be replaced with `bindAsync` in the tests. `forceShutdown` might actually be needed. It's called as a cleanup function to make sure test jobs always end. The implementation in the other library cancels all outstanding calls, then shuts down the server as normal but without a callback.

cjihrig:
OK. To implement both shutdown methods properly, I think the server will need to start tracking all active sessions and streams. The session tracking is needed for `tryShutdown()` to gracefully close everything properly - the Node HTTP2 docs recommend calling the server's `close()` method, as well as calling each active session's `close()` method. `forceShutdown()` needs access to all of the streams in order to cancel the calls. The server could just `destroy()` all of the sessions, but the calls wouldn't be properly cancelled that way.

murgatroid99:
I think tracking streams is unnecessary. Destroying all active sessions should be sufficient for `forceShutdown`. The docs say that sends a final GOAWAY frame with an `INTERNAL_ERROR` code by default, which should serve the purpose here of informing the client that the server is no longer handling streams on that session. We could also try sending a `CANCEL` error code; it's not clear if that's valid for GOAWAYs.

cjihrig:
@murgatroid99 I've pushed up a few changes. I've moved from `bind()` to `bindAsync()`, added a `forceShutdown()` implementation, and improved `tryShutdown()`.

packages/grpc-js/src/server.ts
```diff
@@ -223,7 +224,25 @@ export class Server {
   }
 
   forceShutdown(): void {
-    throw new Error('Not yet implemented');
+    if (this.http2Server === null) {
+      throw new Error('server is not running');
```
murgatroid99: This should not be treated as an error. Just make it a no-op if there's no server. The same is true for `tryShutdown`.

packages/grpc-js/src/server.ts
```diff
@@ -266,10 +285,39 @@ export class Server {
       return;
     }
 
-    this.http2Server.close((err?: Error) => {
-      this.started = false;
-      callback(err);
+    let pendingChecks = 0;
+    let callbackError: Error | undefined;
```
murgatroid99: I think we're better off just omitting all of the error handling code from this function. The docs don't say that `Http2Session#close` pass an error to the callback, and the current `tryShutdown` implementation doesn't pass any meaningful errors to its own callback.

murgatroid99:
If I'm reading those Linux test logs correctly, for 3 different tests where the request is supposed to end with a specific error from the server, in the native client-pure js server case, the native client is emitting an INTERNAL status code instead of the expected code.

cjihrig:
I pushed a fix up for two of the tests. The third is related to [this call](https://github.com/grpc/grpc-node/blob/2e13f46141c66c11b574a6d865b4422646e8a2de/packages/grpc-js/src/server.ts#L352), I believe. It's interesting that it only fails with the native client. I believe it will fail with an `INTERNAL` status on grpc-js@0.4.2, but pass with `UNAVAILABLE` on 0.4.3.

murgatroid99:
It's possible that that test failure is actually a timing issue. If so, #925 should fix that. Once that's merged in, you might have to merge master into this branch. That might fix the failure.

cjihrig:
I've rebased this since #925 landed.

cjihrig:
@murgatroid99 can you kick off the CI again. I think I may have solved the problem with the last test (https://github.com/grpc/grpc-node/pull/923/commits/e9946c3f6cc38e73c9c5585816e5856634798294).

It seems like the native client was making the final unary call to the server, but the connection wasn't actually in tact. I was able to make the test fail reliably locally by adding a timeout between `server1.tryShutdown()` and `server2.bindAsync()`. The changes in the latest commit made it begin passing reliably for me - then I removed the extra timeout.

murgatroid99:
I think there may be an actual issue with what the server is sending that is resulting in the native library emitting INTERNAL errors. I can dig into it tomorrow.

murgatroid99:
I have not yet been able to reproduce this error on my own computer. This is going to be tricky. First, those test logs have gotten a little weird. I pushed a commit skipping the connectivity test to verify that it's actually the only thing that's failing.

murgatroid99:
That looks like confirmation. That Windows test failure is on a test I already disabled in another branch. I think I'm OK merging in this state; I'm not entirely confident in the construction of that connectivity test, and I don't think it's a strong enough signal to be worth blocking the server.

BugDiver:
So happy to see this in progress. It is a much-awaited feature 🎉 🎉

PR #925, opened 2019-06-24
Connectivity test: ensure all calls end before ending the test

murgatroid99:
Avoid potential timing issues with making the final calls and calling `forceShutdown` on the server in the `after` hook.

### DIFF
--- a/test/api/connectivity_test.js
+++ b/test/api/connectivity_test.js
@@ -76,6 +76,13 @@ describe('Reconnection', function() {
     server2.forceShutdown();
   });
   it('Should end with either OK or UNAVAILABLE when querying a server that is shutting down', function(done) {
+    let pendingCalls = 0;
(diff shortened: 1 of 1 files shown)


test/api/connectivity_test.js
```diff
@@ -76,6 +76,13 @@ describe('Reconnection', function() {
     server2.forceShutdown();
   });
   it('Should end with either OK or UNAVAILABLE when querying a server that is shutting down', function(done) {
+    let pendingCalls = 0;
+    let testDone = false;
+    function maybeDone() {
+      if (testDone && pendingCalls == 0) {
```
cjihrig: It shouldn't matter in this case, but maybe use `===` instead of `==`.

murgatroid99:
I already set that Windows test to be skipped on the 1.22 branch. It's just flaky.

PR #942, opened 2019-07-08
Resolve out of order operations on Node 12.6.0

tcf909:
In Node 12.6.0 there appears to be an additional tick incurred which allows this.finalStatus to be set before handleFilteredRead() can process the message. Removing this unnecessary Promise allows handleFilteredRead() to process the message and call  before this.finalStatus is set.

### DIFF
--- a/packages/grpc-js/src/call-stream.ts
+++ b/packages/grpc-js/src/call-stream.ts
@@ -186,7 +186,7 @@ export class Http2CallStream extends Duplex implements Call {
     }
     this.isReadFilterPending = true;
     this.filterStack
-      .receiveMessage(framedMessage)
+      .receiveMessage((async () => framedMessage)())
       .then(
         this.handleFilteredRead.bind(this),
         this.handleFilterError.bind(this)
--- a/packages/grpc-js/src/call-stream.ts
+++ b/packages/grpc-js/src/call-stream.ts
@@ -186,7 +186,7 @@ export class Http2CallStream extends Duplex implements Call {
     }
     this.isReadFilterPending = true;
     this.filterStack
-      .receiveMessage(Promise.resolve(framedMessage))
+      .receiveMessage(framedMessage)
       .then(
         this.handleFilteredRead.bind(this),
         this.handleFilterError.bind(this)


thelinuxfoundation:

* User @tcf909 isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

tcf909:
reverify CLA

murgatroid99:
The filter stack itself is asynchronous, so there are still going to be ticks in between calling `filterReceiveMessage` and calling `handleFilteredRead`. This change looks fine but I think a bigger change will be needed to completely fix the problem you describe. The right way to fix this is probably to more robustly ensure that all incoming messages are processed before emitting a status based on the received trailers.

tcf909:
Totally agree. The bigger change should also fix the race to set this.finalStatus that happens between an error condition and a normal close condition.

This was the smallest code we could change to make sure we could still run in production on node 8, 10 and 12.

murgatroid99:
OK, this change isn't actually valid. You can't just pass a regular object in place of a promise.

tcf909:
Interestingly enough you can since the filters are all async functions and they end up converting the result to a promise.

We were able to determine that our use of the Bluebird promise library with Promise.resolve() was what was triggering the issue. The async aspect of this code still needs to be cleaned up though since its 100% dependent on code running in the ordered ticks it expects (which is far from a certainty).

We will take a look at a broader fix and see about reopening this pr then.

tcf909:
I've updated the pr with the change we are using in production now. It conforms to the expectation of a promise being passed, but bypasses 3rd party promise libraries by using an async function executed immediately.

Hopefully this quick fix will save anyone else from having issues in the future until a broader design change can be made.

murgatroid99:
I am very hesitant to try to support an environment in which globals like `Promise` may be modified in ways I can't predict. If this third party `Promise` library is behaving in a way that is essentially incompatible with the built in `Promise`, I would argue that the responsibility is on that library and not this one to ensure compatibility.

tcf909:
Definitely understand your perspective. Unfortunately, Promises were a little late to the game and there are a lot of third-party libraries that implement the spec, but the spec doesn't go as far to specify the difference between process.nextTick, setImmediate, etc. We use to have a list of which libraries used the microtask queue (nextTick(), and which used the normal event queue (setImmediate, etc.)

Beyond that, the microtask queue went through a big overhaul somewhere in 10.x - 12.0 which changed the timing of promises all together in Node.js.

We support a lot of user code (that we don't control), unfortunately. So things like global.Promise = <insert better library here> happen (which I think you will find happens more than you might think in the node.js ecosystem -- again, unfortunately.)

We only ran into the issue as we looked to prepare for supporting 12.x for our users. Feel free to reject (we've setup up an auto-applied patch), but just know there will be others that run in to this.

tcf909:
You might think of it as your own version of {}.prototype.hasOwnProperty.call() #insert_bad_joke_here_about_unfortunate_standards_that_grow_from_wild

tcf909:
You might also note that Bluebird is by far one of the most popular Promise libraries out there...(as I'm sure you probably already know).

murgatroid99:
I am glad you brought this to my attention and I would like to do something about it. But my preferred solution would be to ensure that the code isn't racy even considering the variety of possible `Promise` behaviors, rather than trying to micromanage the timings like this.

6utt3rfly:
Thank you for this PR! I found that when I upgraded to node 12, while using Bluebird promises, I could no longer use pagination for google-cloud modules (i.e. pubsub.getTopics, monitoring.getDescriptors, etc.) since gax-nodejs uses grpc-js. This patch fixes that (or I could pass in full grpc to google modules instead of using grpc-js).

bensimian:
Hi, 

We recently started the upgrade to node 12 and are also experiencing this issue with the usage of Bluebird. Any ideas on landing a fix?

tcf909:
> Hi,
> 
> We recently started the upgrade to node 12 and are also experiencing this issue with the usage of Bluebird. Any ideas on landing a fix?

We are still running this patch in productions without any issues. The fundamental problem is how grpc-node currently handles its internal state. 

The maintainer hopes that a patch will fix the fundamental problem, but we decided that messing with something as fundamental as the internal state management without a ton more discovery and leg work put a lot more of our code at risk.

tcf909:
> Hi,
> 
> We recently started the upgrade to node 12 and are also experiencing this issue with the usage of Bluebird. Any ideas on landing a fix?

We are still running this patch in productions without any issues. The fundamental problem is how grpc-node currently handles its internal state. 

The maintainer hopes that a patch will fix the fundamental problem, but we decided that messing with something as fundamental as the internal state management without a ton more discovery and leg work put a lot more of our code at risk.

murgatroid99:
I believe this has been resolved by other changes in grpc-js 0.7.0. Please try that version out and see if it works for you. If not, please file a new issue and we can take a look at the current code.

Issue #941, opened 2019-07-08
Not enough responses received


mrfelton:
### Problem description

When connecting to some gRPC servers with grpc-js I'm getting the error message `Not enough responses received` when trying to call a method against one of the exposed services.

I'm trying to understand exactly what this error message means or what could be the cause of it. Any light that you could shed on it would be appreciated.

### Reproduction steps
I don't really have reproducible steps that I can provide at this point. I'm more just lloking for information on possible causes of this error.

### Environment
 - OS name, version and architecture: mac
 - Node version [e.g. 8.10.0]: 12
 - Node installation method [e.g. nvm]: nvm
 - If applicable, compiler version [e.g. clang 3.8.0-2ubuntu4]
 - Package name and version [e.g. gRPC@1.12.0]: grpc-js

### Additional context
N/A

mrfelton:
To add a little more info to this. I switched out `grpc-js` for `grpc` and I get a different error:

`Received http2 header with status: 404`

This is actually the error that I was expecting since I know that the service in question is not available.

So, this leads me to believe that there is a difference in how grpc and grpc-js handle the 404 case, with grpc handling it correctly and grpc-js providing a very obscure error message.

mrfelton:
A little more info from trying to debug this.

Whilst this doesn't fix this issue, I believe there is a bug in the code here:

https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/call-stream.ts#L246-L268

You are checking the http headers against strings, but the status is returned in the headers as a Number. So this switch will always fall back to the default case and set the status code to `UNKNOWN`

mrfelton:
What I can't quite figure out is that even with that fixed and `mappedStatusCode` getting set correctly to `UNIMPLEMENTED` in the case of a 404, this status code does not make its way to the response.

murgatroid99:
Thank you for pointing this out. I totally missed that `:status` would be represented as a number. And I think I see what's happening here: [this error](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/client.ts#L144) is overriding the one from the HTTP status code.

PR #943, opened 2019-07-08
fix: properly handle http error status codes

mrfelton:
This is an attempt to get a fix for #941

There are 2 things here:

1) Correctly check http error status codes as number rather than string. This is probably correct.

2) Trigger `handleTrailers` in the case of an http error. This is probably not the right thing to do, but I'm not very familiar with the codebase or protocol and I'm literally just hacking away trying to get the result I'm looking for. This does fix the error I'm experiencing, although probably not the right way to go about it.

### DIFF
--- a/packages/grpc-js/src/call-stream.ts
+++ b/packages/grpc-js/src/call-stream.ts
@@ -243,24 +243,24 @@ export class Http2CallStream extends Duplex implements Call {
     } else {
       this.http2Stream = stream;
       stream.on('response', (headers, flags) => {
-        switch (headers[HTTP2_HEADER_STATUS]) {
+        switch (headers[':status']) {
           // TODO(murgatroid99): handle 100 and 101
-          case '400':
+          case 400:
             this.mappedStatusCode = Status.INTERNAL;
             break;
-          case '401':
+          case 401:
             this.mappedStatusCode = Status.UNAUTHENTICATED;
             break;
-          case '403':
+          case 403:
             this.mappedStatusCode = Status.PERMISSION_DENIED;
             break;
-          case '404':
+          case 404:
             this.mappedStatusCode = Status.UNIMPLEMENTED;
             break;
-          case '429':
-          case '502':
-          case '503':
-          case '504':
+          case 429:
+          case 502:
+          case 503:
+          case 504:
             this.mappedStatusCode = Status.UNAVAILABLE;
             break;
           default:


packages/grpc-js/src/call-stream.ts
```diff
@@ -245,29 +245,32 @@ export class Http2CallStream extends Duplex implements Call {
...
+          case 503:
+          case 504:
             this.mappedStatusCode = Status.UNAVAILABLE;
             break;
           default:
             this.mappedStatusCode = Status.UNKNOWN;
         }
 
-        if (flags & http2.constants.NGHTTP2_FLAG_END_STREAM) {
+        if (flags & http2.constants.NGHTTP2_FLAG_END_STREAM || this.mappedStatusCode !== Status.OK) {
```
murgatroid99: The problem here is that if the `END_STREAM` flag is not set, then there will still be actual trailers later. I think there might be a better solution here that doesn't involve discarding those trailers.

mrfelton: Yeah, I figured. I'm open to suggestions!

murgatroid99: The simplest solution for now would probably be to just remove that `end` event handler I linked to in the issue. I think that check actually doesn't exist in the other implementation.

mrfelton: That doesn't work because the status code set in `mappedStatusCode` only gets handled in `handleTrailers`, which never gets called again.

mrfelton: Something worth noting here is that in the case that I'm working with, the problem arrises when grpc is is being served behind nginx with `grpc_pass`.

In this case, nginx serves the 404 error - so I don't think the trailers ever come.

mrfelton: After some further debugging it seems that the nginx proxy was incorrectly setup in my case. So, I have updated this PR to remove the changes related to the handing of the trailers - it now just fixes the http status code checking (number vs string)

packages/grpc-js/src/call-stream.ts
```diff
@@ -245,22 +245,22 @@ export class Http2CallStream extends Duplex implements Call {
       stream.on('response', (headers, flags) => {
         switch (headers[HTTP2_HEADER_STATUS]) {
```
murgatroid99: This line now has to say `headers[':status']` so that TypeScript can recognize that the value will be a number.

mrfelton: Updated

murgatroid99:

murgatroid99:
0.5.1 is now out with this fix.

PR #945, opened 2019-07-08
Build electron 4.2.x artifacts

CapOM:
Cherry-pick ec91a16


CapOM:
Hi @murgatroid99 please take a look, thx

murgatroid99:
Thank you for helping with this, but we're not going to be rebuilding versions earlier than 1.22.

CapOM:
Ah ok no worries.

PR #959, opened 2019-07-10
Fix 'noImplicitAny' issue with declaration file.

CapOM:
Cherry-pick 26e30ab136d8ec00106fd8d67844e5df651a4df8

The latest release of grpc-node cannot be used in a typescript project that uses -noImplicitAny.  This is due to a signature being added without a specified return type.  This causes the return type to implicitly be 'any' which TS explicitly disallows.  For projects that use strict flags to catch lots of bugs, this prevents usage of this version of grpc entirely.


murgatroid99:
Thank you for this. I actually thought the change was on the other package and I didn't realize it needed to be backported.

Issue #965, opened 2019-07-12
Namespace '"grpc"' has no exported member 'ServerWritableStream'.


tatianafrank:
Im writing a node GRPC app using typescript and have a method that is streaming on the server side only. But I get a warning in my IDE when I try to define the type of the call as a 'ServerWritableStream', it says grpc doesnt have that exported member. If I try to use 'ClientReadableStream' (I dont know the difference) as the type then the error I get is that "write" and "end" dont exist as methods on that type. 

```
watch(call: grpc.ServerWritableStream<grpc.Call>): void { 
    // I think the type should be grpc.ClientReadableStream<grpc.Call> but throws error that says "Namespace '"grpc"' has no exported member 'ServerWritableStream'"
    try {
      call.write({ status: ServingStatus.SERVING });
      call.end();
    } catch (error) {
      call.write({ status: ServingStatus.NOT_SERVING });
      call.end();
    }
  }
```
Protobuf:
`rpc Watch(HealthCheckRequest) returns (stream HealthCheckResponse);`

### Environment
 - mac
 - Node version 10.16
 - Node installation method- NVM
 - gRPC@1.21.1
 - @grpc/proto-loader@0.3.0

murgatroid99:
The `Server` and `Client` prefix in those types depend on whether you are writing server code or client code. It looks like you are writing a server method implementation so `ServerWritableStream` is correct there. It is a bug that that type is not exported.

murgatroid99:
Also,`ServerWritableStream` takes two type parameters, the request message type and the response message type.

tatianafrank:
Revisited this morning and it looks like changing it from ServerWrit**a**bleStream to ServerWrit**ea**bleStream fixed the issue!

PR #972, opened 2019-07-16
Improve server-related types exported by grpc-js

murgatroid99:
This mainly makes the server call types and `Server#registerService` function signature more specific, the server error types more permissive, and exports some more types needed for writing service implementations.

Also, this fixes a typo in the native library's `index.d.ts` that got brought up in #965. That class is essentially internal, so code written using the typo works. I have added an alias to avoid breaking that.

### DIFF
--- a/packages/grpc-js/src/server-call.ts
+++ b/packages/grpc-js/src/server-call.ts
@@ -18,7 +18,6 @@
 import { EventEmitter } from 'events';
 import * as http2 from 'http2';
 import { Duplex, Readable, Writable } from 'stream';
-import * as util from 'util';
 
 import { StatusObject } from './call-stream';
 import { Status } from './constants';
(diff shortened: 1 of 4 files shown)


packages/grpc-js/src/server-call.ts
```diff
@@ -18,13 +18,14 @@
 import { EventEmitter } from 'events';
 import * as http2 from 'http2';
 import { Duplex, Readable, Writable } from 'stream';
+import * as util from 'util';
```
cjihrig: `util.isNumber()` and `util.isString()` are deprecated in favor of `typeof value === 'number'` and `typeof value === 'string'`, respectively.

packages/grpc-js/src/server-call.ts
```diff
@@ -490,21 +495,21 @@ export class Http2ServerCallStream<
...
         ? error.message
         : 'Unknown Error',
-      metadata: error.hasOwnProperty('metadata')
+      metadata: ('metadata' in error && error.metadata !== undefined)
         ? error.metadata
         : new Metadata(),
     };
 
-    if (error.hasOwnProperty('code') && Number.isInteger(error.code)) {
+    if ('code' in error && util.isNumber(error.code) && Number.isInteger(error.code)) {
```
cjihrig: The `util.isNumber()` (or `typeof error.code === 'number')` check shouldn't be necessary due to the `Number.isInteger()` call.

murgatroid99: I think Typescript was not happy with calling `Number.isInteger` on that value without verifying that it was a number.

cjihrig: I think you're right: https://github.com/Microsoft/TypeScript/issues/21199 😕

cjihrig: I think you're right: https://github.com/Microsoft/TypeScript/issues/21199 😕

PR #973, opened 2019-07-17
Add dev-dependencies and fix tests so they exit after running

bjornstar:
My goal was to be able to successfully:

```
git clone git@github.com:grpc/grpc-node.git
cd grpc-node/packages/grpc-js
npm i
npm run test
```

What I did:

1. Specify all of the dependencies required to run the tests in `package.json` as `devDependencies`
2. Add teardown steps to two tests that prevented the process from exiting when the tests completed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "through2": "^2.0.3",
     "ts-node": "^8.1.0",
     "tslint": "^5.5.0",
-    "typescript": "~3.3.3333",
+    "typescript": "^3.5.3",
     "xml2js": "^0.4.19"
   },
   "contributors": [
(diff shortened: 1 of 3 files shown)


bjornstar:
Can someone take a look at this?

murgatroid99:
Currently the `package.json` in the root directory has some shared dev dependencies, and we assume those are present in our tests. This means that the tests we currently have won't guard against some other dev dependency being introduced without being in this `package.json` file.

Also, the tests are failing now. It is likely that one of the new dependencies has a version that is not consistent with the version we are currently using.

bjornstar:
I've bumped the version of typescript to the latest available, it seems to have done the trick.

bjornstar:
The tests are passing for me, are there more changes you would like to see?

bjornstar:
Could someone take a look at this?

PR #979, opened 2019-07-31
Deadline exceeded

Nobotym:
When timeout for connection is exceeded a new connection is started. I think if timeout is exceeded there is should not more tries.

### DIFF
--- a/packages/grpc-js/src/call-stream.ts
+++ b/packages/grpc-js/src/call-stream.ts
@@ -365,6 +365,12 @@ export class Http2CallStream extends Duplex implements Call {
       metadata
     );
   }
+  
+  private closeChannel() {
+    if (this.channel) {
+      this.channel.close();
+    }
+  }
 
   private destroyHttp2Stream() {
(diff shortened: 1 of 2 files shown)


thelinuxfoundation:

* User @Nobotym isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

Nobotym:
I signed it

Nobotym:
I signed it

packages/grpc-js/src/call-stream.ts
```diff
@@ -377,6 +383,7 @@ export class Http2CallStream extends Duplex implements Call {
   }
 
   cancelWithStatus(status: Status, details: string): void {
+    this.closeChannel();
```
murgatroid99: I don't know what you think this will do, but it's not correct. There can be multiple calls on a single channel, and a call ending with an error should not close its channel.

murgatroid99:
The channel code has been reworked in version 0.6.x. It no longer has the described behavior.

PR #981, opened 2019-08-03
unary request and client stream promise support

EduardoRFS:
Support `then()` when unary request or client stream, essentially both APIs are the ones that already exposes a callback, so I just removed the need for a callback and make the return of both thenable.

No breaking change, every test pass(and two new ones), the API only became more open, now allow to avoid a callback.

Usage example:
```javascript
// with callback
it('with unary call', function(done) {
  registry = new CallRegistry(done, expected_calls, true);
  var message = { value: 'foo' };
  client.echo(message, options, function(err) {
    if (!err) { // do what if error happen? yeah nothing
      registry.addCall('response');
    }
  });
});

// async await version
it('with unary call', async function(done) {
  registry = new CallRegistry(done, expected_calls, true);
  var message = { value: 'foo' };
  await client.echo(message, options); // will throw if fail
  registry.addCall('response'); 
});
```

### DIFF
--- a/packages/grpc-native-core/index.d.ts
+++ b/packages/grpc-native-core/index.d.ts
@@ -1089,8 +1089,8 @@ declare module "grpc" {
       argument: RequestType | null,
       metadata: Metadata | null,
       options: CallOptions | null,
-      callback: requestCallback<ResponseType>,
-    ): ClientUnaryCall;
+      callback?: requestCallback<ResponseType>,
+    ): ClientUnaryCall & PromiseLike<ResponseType>;
 
     /**
      * Make a client stream request to the given method, using the given serialize
@@ -1109,8 +1109,8 @@ declare module "grpc" {
       deserialize: deserialize<ResponseType>,
       metadata: Metadata | null,
       options: CallOptions | null,
-      callback: requestCallback<ResponseType>,
-    ): ClientWritableStream<RequestType>;
+      callback?: requestCallback<ResponseType>,
+    ): ClientWritableStream<RequestType> & PromiseLike<ResponseType>;
 
     /**
      * Make a server stream request to the given method, with the given serialize
--- a/packages/grpc-native-core/src/client.js
+++ b/packages/grpc-native-core/src/client.js
@@ -559,8 +559,10 @@ Client.prototype.makeUnaryRequest = function(path, serialize, deserialize,
     intercepting_call.halfClose();
   });
 
-  // avoid UnhandledPromiseRejectionWarning
-  promise.catch(() => {});
+  if (typeof callProperties.callback === 'function') {
+    // avoid UnhandledPromiseRejectionWarning
+    promise.catch(() => {});
+  }
 
   emitter.then = promise.then.bind(promise);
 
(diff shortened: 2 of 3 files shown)


thelinuxfoundation:

* User @EduardoRFS isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

EduardoRFS:
CLA?

murgatroid99:
In general, we require API changes to go through our [gRFC proposal process](https://github.com/grpc/proposal).

In addition, the problem I see here is that for existing users and anyone who uses the callback API, this will introduce new `UnhandledPromiseRejectionWarning`s whenever a call has an error. According to that warning's text, those will be fatal errors in some future Node version. When that happens, anyone who uses the callback API instead of the promise API will have their process exit every time a call has an error, even though they are "handling" it in the callback. That is unacceptable.

EduardoRFS:
@murgatroid99 

> In general, we require API changes to go through our gRFC proposal process.

Even non breaking ones? 

> UnhandledPromiseRejectionWarnings whenever a call has an error

I already thought about it(when I ran the tests it was flooded with that)
```javascript
if (typeof callProperties.callback === 'function') {
  // avoid UnhandledPromiseRejectionWarning
  promise.catch(() => {});
}
```

murgatroid99:
Yes, we do require proposals for non-breaking API changes. That is most of the content in that proposal repository.

EduardoRFS:
@murgatroid99 See it will take some time to write an proposal(probably more broadly using the "same" approach"), so would be really be helpful if you could give me a feedback about making APIs `PromiseLike`

murgatroid99:
One of the primary purposes of making a proposal is to get feedback about the design.

It does seem a little weird to me to not have the full `Promise` prototype here. Do you know how well `PromiseLike` objects interact with existing APIs such as `Promise.all`? Other than that, this solution looks like a good way to add promise support, but I am somewhat worried about increasing the complexity of these APIs. What are the benefits of adding this as compared with leaving it alone and leaving it up to users to "promisify" the API?

EduardoRFS:
@murgatroid99 It should work with every API that supports promise, the base of a Promise it's a thenable(object that have a then). https://promisesaplus.com/

The advantage is that we can by base allow it to be a the standard library for basic usages without a lot of wrappers, wrappers are easy to do, one problem is typescript support, with static generated code.

I could easily fork it or make a wrapper, but then I need to fork the entire ecosystem(well I almost did)

- I'm waiting this to make a PR for [grpc_tools_node_protoc_ts](https://github.com/EduardoRFS/grpc_tools_node_protoc_ts/commit/adcd407331a94ec0134db56db32c14dd73c6b161#diff-a6b6b3213488bbfaa24afd1f6499efd6) with the right types.
- [One PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37354#) on `@types/node` for stream api that supports generics. 
- If I could merge the one in `@types/node` I already have one more for this repo with the right types.

After everything I get a 100% typed code with an nice usage, without any major breaking change. ex: [api.spec.ts](https://github.com/EduardoRFS/hash-test/blob/master/discounts/tests/api.spec.ts).

Also it's easier for myself(and for everyone) if everything became part of the mainstream. Anyway, I will write the proposal today or tomorrow, then we could think better about it.

gwer:
It's only been three years. Any news? 😸 

Promises out of the box would be very helpful.

PR #995, opened 2019-08-15
feat: start introducing verbose logging

bcoe:
`grpc-js` has a logging module, which can be configured to output debug information at different levels of verbosity, let's start using it to dig deeper into encoding issues we're seeing periodically.

### DIFF
--- a/packages/grpc-js/src/logging.ts
+++ b/packages/grpc-js/src/logging.ts
@@ -18,7 +18,7 @@
 import { LogVerbosity } from './constants';
 
 let _logger: Partial<Console> = console;
-let _logVerbosity: LogVerbosity = LogVerbosity.DEBUG;
+let _logVerbosity: LogVerbosity = LogVerbosity.ERROR;
 
 export const getLogger = (): Partial<Console> => {
   return _logger;
--- a/packages/grpc-js/src/metadata.ts
+++ b/packages/grpc-js/src/metadata.ts
@@ -19,6 +19,9 @@ import * as http2 from 'http2';
 const LEGAL_KEY_REGEX = /^[0-9a-z_.-]+$/;
 const LEGAL_NON_BINARY_VALUE_REGEX = /^[ -~]*$/;
 
+import {LogVerbosity} from './constants';
+import {log} from './logging';
+
 export type MetadataValue = string | Buffer;
 export type MetadataObject = Map<string, MetadataValue[]>;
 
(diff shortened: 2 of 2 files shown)


thelinuxfoundation:

* User @bcoe isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

packages/grpc-js/src/metadata.ts
```diff
@@ -39,6 +42,7 @@ function normalizeKey(key: string): string {
 }
 
 function validate(key: string, value?: MetadataValue): void {
+  log(LogVerbosity.DEBUG, `validate key = "${key}" value type = "${typeof value}"`);
```
cjihrig: Is it valid to use `printf()` style logging in gRPC? Since this defaults to using `console.error()`, it should at least work in that case - I'm just not sure if it's officially supported by gRPC.

If `printf()` style is supported, I'd recommend using it in this particular case. Template strings can get expensive, since they are evaluated even if nothing is actually logged (see today's Node security releases, which include a commit to move to fewer template strings).

packages/grpc-js/src/logging.ts
```diff
@@ -18,7 +18,7 @@
 import { LogVerbosity } from './constants';
 
 let _logger: Partial<Console> = console;
-let _logVerbosity: LogVerbosity = LogVerbosity.DEBUG;
+let _logVerbosity: LogVerbosity = LogVerbosity.ERROR;
```
cjihrig: Does this match the default verbosity of the compiled addon?

packages/grpc-js/src/metadata.ts
murgatroid99: `log` isn't part of the grpc public API, so we can do whatever we want with it. The `setLogger` API documentation says that the logger should be a "`console`-like object", so I think it's reasonable to call it as though it was `console.error`.

packages/grpc-js/src/logging.ts
murgatroid99: Yes, I'm pretty sure it does.

packages/grpc-js/src/metadata.ts
bcoe: talking with @murgatroid99 offline, I'm a little concerned logging this might be a bit too noisy ... I think we'd do better to figure out some more targeted log lines that could help us start to figure out what the heck is going on when keys are corrupted.

murgatroid99:
This has been superseded by a combination of #1065 and #1174

PR #1001, opened 2019-08-20
Add MAINTAINERS.md

murgatroid99:


murgatroid99:
OK, I have moved a couple of people to the emeritus list and revoked their write permissions. Those are the only two I'm sure about removing.

PR #1004, opened 2019-08-21
Update from master.

nicolasnoble:
Tweaking binding.gyp.template to avoid properties starting with a $.


packages/grpc-native-core/binding.gyp
```diff
@@ -194,25 +204,45 @@
...
+              '-Ithird_party/upb',
+              '-Isrc/core/ext/upb-generated',
               '-fvisibility=hidden',
           ],
           'OTHER_CPLUSPLUSFLAGS': [
               '-g',
               '-Wall',
               '-Wextra',
               '-Werror',
+              '$(W_NO_UNKNOWN_WARNING_OPTION)',
```
murgatroid99: Is this supposed to be here?

nicolasnoble: Gah, I'm not properly filtering all of them out. Thanks.

nicolasnoble: Fixed.

murgatroid99:
It looks like the Mac builds are failing because XCode doesn't recognize `-Wno-maybe-uninitialized`. Based on the error message it looks like that might be resolved by passing `-Wno-unknown-warning-option`.

murgatroid99:
This has been superseded by more recent updates.

PR #1007, opened 2019-08-22
Move EventEmitter inside declare module "grpc"

### LINKED ISSUE #1006: node_modules/grpc/index.d.ts - Imports are not permitted in module augmentations
### Problem description
Im using grpc with typescript but the typescript compile command fails with the following error: 'Imports are not permitted in module augmentations. Consider moving them to the enclosing external module.'

### Reproduction steps
Install grpc in node project with typescript. Add a file that imports grpc. `run tsc -p tsconfig.json` wirh target=es6 and module=commonjs. 

### Environment
 - macOS Mojave
 - Node version 11.10
 - Node installation method : nvm
 - Package name and version:  gRPC@^1.23.2

### Additional context

sandersn:
Fixes #1006 

For sad historical reasons, Typescript's syntax for ambient modules, which is what grpc-native-core's index.d.ts tries to declare, is the same as module augmentations. The way to tell the difference between the two is whether the file has imports or exports outside the declaration. #1002 adds such a declaration, turning the ambient module into a module augmentation. Not only is there no grpc module to augment, module augmentations are subject to a bunch of restrictions since they are only supposed to augment the contents of an existing module. That's the cause of the new errors.

I'm not sure if merging into the grpc@1.23.x branch is correct. That's what I targetted with this change.

### DIFF
--- a/packages/grpc-native-core/index.d.ts
+++ b/packages/grpc-native-core/index.d.ts
@@ -17,6 +17,7 @@
  */
 
 declare module "grpc" {
+  // add imports here, inside the "grpc" module, to keep it as an ambient module
   import { Message, Service as ProtobufService } from "protobufjs";
   import { EventEmitter } from "events";
   import { Duplex, Readable, Writable } from "stream";
--- a/packages/grpc-native-core/index.d.ts
+++ b/packages/grpc-native-core/index.d.ts
@@ -1,4 +1,3 @@
-import { EventEmitter } from "events";
 
 /*
  * Copyright 2019 gRPC authors.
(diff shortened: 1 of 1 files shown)


thelinuxfoundation:

* User @sandersn isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

murgatroid99:
Thank you so much for this. That `EventEmitter` import completely slipped my notice, and I didn't even realize it made that kind of difference.

murgatroid99:
Would you mind adding a comment indicating that imports need to be inside of the module declaration, so that we can avoid this in the future?

sandersn:
Done. Warning: I work with this terminology every day so I can't tell whether the message is at all comprehensible.

PR #1013, opened 2019-08-27
Use cross docker only for artifacts

veblush:
Instead of having two docker images, only one latest docker image is used. From node.js 12, GCC 6.3 or later is required ([ref](https://medium.com/@nodejs/introducing-node-js-12-76c41a1b3f3f)) and only cross docker based on `debian:stretch` has it.

To address the problem where generated artifacts from GCC 6.x have a dependency against newer version of libstdc++, `_GLIBCXX_USE_CXX11_ABI=0` is defined for linux to use the old version of API explicitly. ([ref](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html))

### DIFF
--- a/packages/grpc-native-core/binding.gyp
+++ b/packages/grpc-native-core/binding.gyp
@@ -205,6 +205,11 @@
             '-Wno-error=deprecated-declarations'
           ],
         },
+      }],
+      ['OS == "linux"', {
+        'defines': [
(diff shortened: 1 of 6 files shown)


veblush:
Closing this because grpc-note won't have updates with gRPC Core.

Issue #896, opened 2019-05-31
Make grpc-js not wait for ready by default


murgatroid99:
The default core behavior is to end a stream with a non-OK status (usually UNAVAILABLE) as soon as the channel enters the TRANSIENT_FAILURE state, with an option `wait_for_ready` to keep waiting instead. The grpc-js library currently always behaves as though the `wait_for_ready` option is set. That should be changed to align the behavior with the C core.

LINKIWI:
Hi @murgatroid99, curious if there has been any progress on this bug. This bug prevents fail-fast behavior in my client, which is necessary for my use case. I'm happy to contribute if the fix is accessible to someone new to gRPC internals.

murgatroid99:
I'm working on it but it's taking some time. I'm fixing this as part of a larger overhaul of how channel connectivity is handled.

murgatroid99:
This has been fixed as of 0.6.x.

Issue #767, opened 2019-03-07
grpc-js: A channel that never connects keeps the process alive


murgatroid99:
If a channel is trying to connect and it never succeeds and never gets closed by the user, it appears to keep the process alive, even if there are no active calls. This differs from the behavior of the other library, which only keeps the process alive with an active call or other pending asynchronous action from the point of view of the surface API.

murgatroid99:
Update: It looks like what is keeping the process alive is alternately the connection we are trying to establish and the backoff timer.

murgatroid99:
This can be fixed by `unref`ing timers as long as there are no pending calls on the channel. That change should probably wait until the upcoming connection logic changes as part of the load balancer support.

ThomWright:
This might be what is preventing my tests from finishing. I have some tests which launch the pubsub emulator and send/receive some messages using it.

The tests used to pass and exit fine. I have updated the pubsub library like so, which starts using `@google/grpc-js` in favour of `grpc`:

```diff
-    "@google-cloud/pubsub": "^0.28.1",
+    "@google-cloud/pubsub": "^0.30.1",
```

My tests now hang. Using [wtfnode](https://www.npmjs.com/package/wtfnode) I see:

```
[WTF Node?] open handles:
- File descriptors: (note: stdio always exists)
  - fd 1 (tty) (stdio)
  - fd 2 (tty) (stdio)
- Timers:
  - (500 ~ 500 ms) (anonymous) @ myproject/__tests__/test.ts:13
  - (2782 ~ 2 s) (anonymous) @ myproject/node_modules/google-gax/node_modules/@grpc/grpc-js/build/src/channel.js:125
  - (2546 ~ 2 s) (anonymous) @ myproject/node_modules/google-gax/node_modules/@grpc/grpc-js/build/src/channel.js:125
  - (2777 ~ 2 s) (anonymous) @ myproject/node_modules/google-gax/node_modules/@grpc/grpc-js/build/src/channel.js:125
  - (2420 ~ 2 s) (anonymous) @ myproject/node_modules/google-gax/node_modules/@grpc/grpc-js/build/src/channel.js:125
  - (2996 ~ 2 s) (anonymous) @ myproject/node_modules/google-gax/node_modules/@grpc/grpc-js/build/src/channel.js:125
  - (2716 ~ 2 s) (anonymous) @ myproject/node_modules/google-gax/node_modules/@grpc/grpc-js/build/src/channel.js:125
[...]

PR #1015, opened 2019-08-29
Name resolution load balancing

murgatroid99:
This adds a custom name resolver and load balancer internal API with basic implementations. The name resolver implementation handles retrieving and service config data from DNS TXT records, but most of the service config options are not implemented. The only implemented load balancer implementation is "pick first". It will attempt to establish connections to every address returned by the resolver, and will use whichever one connects first.

This change fixes the following bugs:

 - Calls on channels that quickly fail to establish a connection will end with an `UNAVAILABLE` error instead of waiting forever or until the call deadline. The `waitForReady` option to override that behavior is now supported. (#896)

 - Channels with no calls no longer keep the event loop alive. (#767)

Some parts of this design are not in this PR and will be added in future PRs:

 - UDS resolver
 - Round robin load balancer
 - Channel argument handling for the features implemented here


packages/grpc-js/src/channel.ts
```diff
@@ -128,240 +105,220 @@ export interface Channel {
...
+        this.updateState(connectivityState);
+      },
+      requestReresolution: () => {
+        // This should never be called.
+        throw new Error(
+          'Resolving load balancer should never call requestReresolution'
+        );
+      },
+    };
+    // TODO: check channel arg for default service config
```
nicolasnoble: Add your username to the TODO

packages/grpc-js/src/compression-filter.ts
```diff
@@ -138,7 +138,9 @@ class UnknownHandler extends CompressionHandler {
   compressMessage(message: Buffer): Promise<Buffer> {
     return Promise.reject<Buffer>(
       new Error(
-        `Received message compressed wth unsupported compression method ${this.compressionName}`
+        `Received message compressed wth unsupported compression method ${
```
nicolasnoble: That's a bit of a weird looking change. What's the reasoning?

packages/grpc-js/src/index.ts
```diff
@@ -283,3 +283,13 @@ export const InterceptingCall = () => {
 };
 
 export { GrpcObject } from './make-client';
+
+import * as resolver from './resolver';
+import * as load_balancer from './load-balancer';
+
+function setup() {
```
nicolasnoble: The typical paradigm for these is rather something along these lines:

```JS
(function() {
  resolver.registerAll();
  load_balancer.registerAll();
}());
```

packages/grpc-js/src/compression-filter.ts
murgatroid99: I think this came from the code formatter.

paambaati: There’s a typo in `wth` — it should be `with`.

felipegcampos:
Good job on these bugs. I think this PR might fix 2 problems we're currently experiencing. Do you have any estimations on when it should be released?

murgatroid99:
@felipegcampos We want to get it out soon, but it's a lot of code so it's going to take some time for it to be reviewed.

felipegcampos:
> @felipegcampos We want to get it out soon, but it's a lot of code so it's going to take some time for it to be reviewed.

Perfect. I was reviewing myself. However, I don't know if I'm of much help since I have not been checking the codebase since we started with grpc. However, let me know if I can help somehow.

PR #1032, opened 2019-09-13
Update master to 1.24.

nicolasnoble:
Prereq for publishing 1.24.


murgatroid99:
There are also two instances of `-Werror` in the template

Issue #1027, opened 2019-09-11
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. error


apatora:
### Problem description
I am using the package @google-cloud/bigtable which relies on this one and when I ran npm install today it updated @grpc/grpc-js from 0.5.2 to 0.5.3. I am getting the following error :
```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added. Use emitter.setMaxListeners() to increase limit
    at _addListener (events.js:256:17)
    at TLSSocket.addListener (events.js:272:10)
    at TLSSocket.Readable.on (_stream_readable.js:826:35)
    at Http2CallStream.attachHttp2Stream (/usr/src/node_modules/@google-cloud/bigtable/node_modules/@grpc/grpc-js/build/src/call-stream.js:275:35)
    at Http2SubChannel.startCallStream (/usr/src/node_modules/@google-cloud/bigtable/node_modules/@grpc/grpc-js/build/src/subchannel.js:110:20)
    at finalMetadata.then.metadataValue (/usr/src/node_modules/@google-cloud/bigtable/node_modules/@grpc/grpc-js/build/src/channel.js:205:28)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

### Reproduction steps
Run `npm i @google-cloud/bigtable` in a project. Then try to execute for example :
```
const Bigtable = require('@google-cloud/bigtable')

for (let j = 0; j < 100; j++) {
    Bigtable.instance('INSTANCE').table('TABLE').insert([{ key: 'key', data: { data: { some: 'data' } } }]
}
```

### Environment
I am running the code in the following docker image : `node:10.16.0`

murgatroid99:
I think this is the result of #1021. I'll need to figure out a better way to do that.

ddehghan:
also happens on node:12.10.0-alpine

ddehghan:
for now i rolled back to ver "0.5.2" and it fixed the problem.

yatingoswami:
Also facing the same issue on 12.4.0-stretch
Getting pod restarts every 4 hours.

Cant rollback version as its part of google-gax which is nested inside @google-cloud/pubsub.
Is there a proper solution for this

apatora:
@yatingoswami You can solve the issue for now by putting : 
```
"resolutions": {
  "@grpc/grpc-js": "0.5.2"
}
```
[...]

PR #1034, opened 2019-09-19
Hopefully fix memory leak with socket close listener

murgatroid99:
This problem was reported in issue #1027. I think that was the issue, that the callback was never getting cleaned up.

Note: this only tries to fix the memory leak, not the warning. Further changes will be needed to fix the warning but those changes will conflict with #1015 so I want to get that PR merged first.

### DIFF
--- a/packages/grpc-js/src/call-stream.ts
+++ b/packages/grpc-js/src/call-stream.ts
@@ -112,8 +112,7 @@ export class Http2CallStream extends Duplex implements Call {
   // This is populated (non-null) if and only if the call has ended
   private finalStatus: StatusObject | null = null;
 
-  private socketCloseCallback: () => void = 
-    this.socketCloseCallback = () => {
+  private socketCloseCallback: () => void = () => {
       this.endCall({
         code: Status.UNAVAILABLE,
         details: 'Connection dropped',
--- a/packages/grpc-js/src/call-stream.ts
+++ b/packages/grpc-js/src/call-stream.ts
@@ -112,6 +112,15 @@ export class Http2CallStream extends Duplex implements Call {
   // This is populated (non-null) if and only if the call has ended
   private finalStatus: StatusObject | null = null;
 
+  private socketCloseCallback: () => void = 
+    this.socketCloseCallback = () => {
+      this.endCall({
+        code: Status.UNAVAILABLE,
+        details: 'Connection dropped',
+        metadata: new Metadata(),
+      });
+    };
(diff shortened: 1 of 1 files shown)


packages/grpc-js/src/call-stream.ts
```diff
@@ -112,6 +112,15 @@ export class Http2CallStream extends Duplex implements Call {
   // This is populated (non-null) if and only if the call has ended
   private finalStatus: StatusObject | null = null;
 
+  private socketCloseCallback: () => void =
```
cjihrig: Should this be something like the following?

```js
private socketCloseCallback = () => {
  this.endCall(...);
};
```

murgatroid99: Yes, that's a copy/paste error.

packages/grpc-js/src/call-stream.ts
```diff
@@ -347,13 +358,7 @@ export class Http2CallStream extends Duplex implements Call {
...
        * call. This should ensure that this endCall happens sooner than the one
        * in the stream.on('close', ...) handler. */
-      stream.session.socket.on('close', () => {
-        this.endCall({
-          code: Status.UNAVAILABLE,
-          details: 'Connection dropped',
-          metadata: new Metadata(),
-        });
-      });
+      stream.session.socket.on('close', this.socketCloseCallback);
```
stephenplusplus: Would a `once` listener have the same effect, or is this potentially meant to be called multiple times?

murgatroid99: A `once` listener should also work here, but it shouldn't make much of a different because I have to explicitly clean it up either way. And this is just a temporary solution to part of the problem.

stephenplusplus: Fair enough. I saw there were other conditions in the `endCall` handler that had to be met in order to trigger the `removeListener` call. Although, if `once` would also work here, that seems to mean that you could also just let Node's internals handle the removal for you.

No big deal, just thought it might simplify things, but I'm not close enough to the code to know what I'm talking about :)

murgatroid99: The memory leak happens with connections that persist for a long time, which is also when the event will never trigger and never get automatically removed.

murgatroid99:
I'm closing this in favor of #1036. I now know that the warning itself is causing problems so a full revert seems most prudent.

PR #1039, opened 2019-09-20
Stop testing EOL versions of node

### LINKED ISSUE #1038: Drop support for node versions that are EOL
I tried updating some devDependencies in #1035 and ran into test failures in node v6 when newer libraries tried to use the spread operator - `...`

Node v6 has been [EOL](https://github.com/nodejs/Release) for about 5 months now. It would be great if we can drop support for v6 so that we can start using modern javascript and libraries written in modern javascript.

I see tests are running on 6, 7, 8, 9, 10, 11, and 12. Only 8, 10 and 12 are not EOL now.

My proposal is to only run tests on 8, 10 and 12 as they are the only supported versions of node at the moment.

bjornstar:
Resolves #1038 

In order to use modern javascript syntax and libraries using that syntax we need to stop supporting node v6. At the same time, we can also stop testing other versions of node that are officially EOL.

This PR removes testing against node v6, 7, 9, and 11 while moving the initial node version from v8 to v10.

I also added `node-gyp` as a devDependency since it is required to build.

### DIFF
--- a/run-tests.bat
+++ b/run-tests.bat
@@ -38,7 +38,7 @@ call npm install || goto :error
 SET JUNIT_REPORT_STACK=1
 SET FAILED=0
 
-for %%v in (6 7 8 9 10 11 12) do (
+for %%v in (8 10 12) do (
   call nvm install %%v
   call nvm use %%v
   if "%%v"=="4" (
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -26,7 +26,7 @@ set -ex
 cd $ROOT
 
 if [ ! -n "$node_versions" ] ; then
-  node_versions="6 7 8 9 10 11 12"
+  node_versions="8 10 12"
 fi
 
 set +ex


package.json
```diff
@@ -50,6 +51,9 @@
       "name": "Google Inc."
     }
   ],
+  "engines": {
+    "node": "^8.13.0 || >=10.10.0"
```
murgatroid99: I don't really see a point to this, because this package isn't published and this range isn't binding anyway.

run-tests.bat
```diff
@@ -24,12 +24,12 @@ powershell -c "(Get-WmiObject -Class Win32_ComputerSystem).SystemType"
 powershell -c "& { iwr https://raw.githubusercontent.com/grumpycoders/nvm-ps/master/nvm.ps1 | iex }"
 
 SET PATH=%APPDATA%\nvm-ps;%APPDATA%\nvm-ps\nodejs;%PATH%
-SET JOBS=8
+SET JOBS=4
```
murgatroid99: What is the purpose of this change?

package.json
bjornstar: It's advisory the same way it is in `grpc-js`, I've removed it.

murgatroid99:
A dev dependency on node-gyp shouldn't be necessary. It comes with npm.

bjornstar:
> A dev dependency on node-gyp shouldn't be necessary. It comes with npm.

It doesn't It's being installed by the test scripts so I thought to include it. I'll move it into another PR since it's unrelated to this change.

murgatroid99:
node-gyp does come with npm. We install it in the test script to work around a bug in old versions of npm that are included by default in some versions of node that we test, including node 8. But it is not always necessary, and as new node versions are released it should be necessary in fewer cases.

murgatroid99:
That last commit is not acceptable. We still want to publish binaries for older versions of Node.

bjornstar:
> That last commit is not acceptable. We still want to publish binaries for older versions of Node.

Understood, I've backed out all the changes except which node versions to test on.

PR #1042, opened 2019-09-24
Return UNAVAILABLE status on TCP disconnect

murgatroid99:
This is a second attempt at what #1021 tried to do. We only add the listener if the call is not closed and we remove the listener when the call does close, so it should not leak memory. And each socket only gets one event listener, so we should not see the warning about too many listeners for one event.

Also, since this solution required the calls having a reference to the subchannel, I also moved the reference handling code there.


murgatroid99:
I just deleted a test file because it uses an API that I modified here that now takes a Subchannel, and I can't make a mock Subchannel class.

PR #1049, opened 2019-09-26
Add UDS resolver

murgatroid99:
The first followup to #1015.


murgatroid99:
Testing this is currently hard because as far as I know neither Node server implementation can currently bind UDS addresses.

PR #1051, opened 2019-09-26
Add GOVERNANCE.md and CONTRIBUTING.md

srini100:


murgatroid99:
When I added the MAINTAINERS.md file I had it link to these files in grpc-community, just like [the file I based it on in the grpc-go repo](https://github.com/grpc/grpc-go/blob/master/MAINTAINERS.md). Do we no longer want that grpc-community repo to be the reference for those files?

srini100:
These files also link to grcp-community files. We need GOVERNANCE.md and CONTRIBUTING.md to be explicitly present in the root directory.

PR #1055, opened 2019-09-30
Properly use filtered metadata when starting requests

murgatroid99:
This should fix metadata elements from `CallCredentials` (plus a couple of other sources) not being sent to the server. This should have been caught by metadata credentials interop tests; I will need to figure out why we missed this.


nicolasnoble:
Lots of timeouts... is that okay?

Issue #1064, opened 2019-10-04
Error: No connection established


apatora:
### Problem description
Since `google-gax` forced the `grpc-js` version to 0.6.4 (https://github.com/googleapis/gax-nodejs/commit/628db9e84c61017dfaba22ac7c2215e2c79c9745), I am having the following issue :
```
Error: No connection established
    at Http2CallStream.call.on (/usr/src/node_modules/@google-cloud/bigtable/node_modules/@grpc/grpc-js/build/src/call.js:68:41)
    at Http2CallStream.emit (events.js:203:15)
    at Http2CallStream.EventEmitter.emit (domain.js:448:20)
    at process.nextTick (/usr/src/node_modules/@google-cloud/bigtable/node_modules/@grpc/grpc-js/build/src/call-stream.js:75:22)
    at process._tickCallback (internal/process/next_tick.js:61:11)
```

### Reproduction steps
The bug only happens after one hour. We are using `grpc-js` inside `@google-cloud/bigtable` and our bigtable instance is autoscaling.

alexander-fenster:
Thank you for the report! Just to confirm, does that mean that the script should have an hour of idle time with no network requests to experience that, or just an hour of runtime?

harscoet:
Similar issue on [@google-cloud/pubsub@1.1.0](https://github.com/googleapis/nodejs-pubsub)
After about 10 minutes of idle time with no network requests, any call to "publish" callbacks after 10 minutes (maxRetryDelayMillis) to say "Retry total timeout exceeded before any response was received"
To be able to find this error "No connection established", I had to add logs in the google-gax module

reichjustin:
We are seeing the exact same issues as @harscoet with `@google-cloud/pubsub@1.1.0`

apatora:
It happens after an hour of runtime (I had this issue on our main API). Then it starts to throw those errors on separate processes progressively.

<img width="1834" alt="Capture d’écran 2019-10-04 à 18 21 54" src="https://user-images.githubusercontent.com/15031808/66223548-0a31c580-e6d4-11e9-90c8-26e7bdfd6550.png">

weipublic:
Same issue when using `@google-cloud/datastore` which depends on `google-gax`

alexander-fenster:
We are looking into this and will either fix it ASAP, or release an update to google-gax that pins `@grpc/grpc-js` to an older version.

murgatroid99:
[...]

PR #1070, opened 2019-10-08
grpc-js: pick-first: fix bad state transition when reconnecting connected LB

murgatroid99:
`connectToAddressList` can be called while there is a current pick, and it should not update the state to override that current pick. I think this is what is causing the remaining reports of the problem in #1064 after the partial fix in 0.6.6.


murgatroid99:
The build jobs are not impacted by this change

PR #1079, opened 2019-10-21
Upmerge from 1.23.x

murgatroid99:
I guess all of the 1.23.x branch didn't make it into 1.24.x.


murgatroid99:
The other user signed the old CLA. I'm just moving the commit around.

Issue #1085, opened 2019-10-24
`@grpc/grpc-js` channel and memory leak (with a minimal reproduction)


merlinnot:
### Problem description

`@grpc/grpc-js` v0.6.9 seemingly leaks both memory and open network connections.

I did my best to provide a minimal possible reproduction by using gRPC directly, without any other layers on top, such as Google SDKs. Total size of the reproduction is ~50 LOC of readable TypeScript and uses `@grpc/grpc-js`, `@grpc/proto-loader`, `ramda` and `path` modules and packages exclusively. I hope it will make it easier to pinpoint and resolve an issue.

I'd appreciate if you could confirm that the issue is reproducible by following the steps below.

### Reproduction steps

1. Download [reproduction-grpc.zip](https://github.com/grpc/grpc-node/files/3768863/reproduction-grpc.zip).
2. Unzip the file.
3. Navigate to the extracted directory.
4. Run `npm install`.
5. Run Firestore Emulator, either through `gcloud` or using a [Docker container](https://github.com/ridedott/firestore-emulator-docker).
6. Adjust the hostname and port in `configuration.ts` file if necessary. 
7. Run `node -r ts-node/register --expose-gc index.ts`.
8. Heap size will be logged, see it steadily growing. Garbage collection is forced before every log statement to represent accurate results.
9. Run `lsof -i tcp:4500 -n -P`, adjust the port in this script to the one which Firestore Emulator is running on. A list of open sockets will be displayed.
10. Wait 10 seconds, repeat step 9. Wait 10 seconds, repeat step 9. Wait 10 seconds, ... A steadily increasing number of open sockets can be observed.

### Environment
 - OS name, version and architecture: Darwin Kernel Version 19.0.0: Wed Sep 25 21:13:50 PDT 2019; root:xnu-6153.11.26~4/RELEASE_X86_64 x86_64
 - Node version: v10.16.3
 - Node installation method: nvm
 - If applicable, compiler version: -
 - Package name and version: `@grpc/grpc-js` v0.6.9

### Additional context

This might possibly be a root cause of https://github.com/googleapis/nodejs-firestore/issues/768.

murgatroid99:
[...]

PR #1088, opened 2019-10-28
grpc-js: Close http2 sessions that are dropped by their subchannels

murgatroid99:
This should fix the connection leak described in #1085.

### DIFF
--- a/packages/grpc-js/src/subchannel.ts
+++ b/packages/grpc-js/src/subchannel.ts
@@ -321,6 +321,9 @@ export class Subchannel {
         this.continueConnecting = false;
         break;
       case ConnectivityState.TRANSIENT_FAILURE:
+        if (this.session) {
+          this.session.close();
+        }
         this.session = null;
         this.stopKeepalivePings();
         break;
@@ -329,6 +332,9 @@ export class Subchannel {
          * should only transition to the IDLE state as a result of the timer
          * ending, but we still want to reset the backoff timeout. */
         this.stopBackoff();
+        if (this.session) {
+          this.session.close();
+        }
         this.session = null;
         this.stopKeepalivePings();
         break;


Legogris:
@murgatroid99 : Any estimate for when this will be released? We have severly degraded performance after upgrading to 0.6.9 and I want to know if this fixes it.

Also, what is the recommended version to run at this point in time? The memory leak in 0.5.9 is a complete breakage and the last version that looks stable so far if 0.5.4, but I haven't tried every versino in-between yet.

murgatroid99:
I was going to bundle this change with a feature release because I did not consider it high priority. This bug should have only a very minor impact for regular usage of the library. Version 0.6.9 is currently the recommended version; if you have a problem with it I suggest filing an issue.

Legogris:
@murgatroid99 I'm running this on a host that does a large number of outbound http connections - with 0.6.9 a large portion of those requests start timing out, and I would like to see if it's related to these danging sessions.

murgatroid99:
I think it's very unlikely to be related. Your problem sounds similar to one reported on earlier 0.6.x versions that according to all other reports was fixed in 0.6.9 (#1064, googleapis/nodejs-pubsub#770). Can you verify that every instance of `@grpc/grpc-js` in `npm ls` is version 0.6.9, and if so can you file an issue with more information about your problem?

Legogris:
@murgatroid99 Verified. I suspect that this is a duplicate of https://github.com/grpc/grpc-node/issues/1085. It manifests as outgoing HTTP connections failing (these connections don't involve `grpc-js` or PubSub) when publishing messages on PubSub. These issues are not present when using `0.5.4`.

Where can I submit logs?

Legogris:
@murgatroid99 I think I have a satisfactory explanation; Since we have a global http agent configured with `maxSockets` set to a finite value, the dangling sockets end up blocking the other http requests made from the app. Even with separate agents, a lot of dangling sockets will become an issue and it would be great to see this released soon.

murgatroid99:
OK, I see now how these might be connected. I'll publish a new version of grpc-js soon.

murgatroid99:
Version 0.6.10 is out now

PR #1089, opened 2019-10-28
Update to 1.25.0-pre1

murgatroid99:
I have verified that this PR is pointed at the 1.25.x branch.


splix:
Hi Michael! Do you have any ETA on when the release could be made?

murgatroid99:
We're tracking down some test failures. We'll be able to get this out when that's done.

PR #1091, opened 2019-10-30
build: fix typescript compilation errors

JustinBeckwith:
It looks like compilation was failing on master.  Not sure why the unit tests are failing.


murgatroid99:
The tests are failing because of DefinitelyTyped/DefinitelyTyped#39986, which was fixed in DefinitelyTyped/DefinitelyTyped#39992. That fix was published in `@types/node@12.12.1`, so no fix to our code should be needed.

PR #1095, opened 2019-10-30
Bump grpc-js to 0.6.10

murgatroid99:
I have been convinced that it's worth publishing #1088 now rather than waiting and folding it into a bigger release. This release will also include the round robin load balancer implementation in #1063. The only API change there is a newly supported key in the channel options object.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/grpc-js",
-  "version": "0.6.9",
+  "version": "0.6.10",
   "description": "gRPC Library for Node - pure JS implementation",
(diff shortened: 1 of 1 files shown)


Legogris:
I guess this depends on your release strategy, but I would advise releasing the bugfix separately as a patch release (`0.6.10`) and the load balancer feature as a new minor release (`0.7.0`).

Not only to be compatible with semver (patch releases only contain bug fixes) but also in the off event that there's unexpected behavior for some users with the round-robin implementation, it provides:

* A safe version for those users
* Much easier to track down where any issues lie

Appreciate the hard work!

Legogris:
Gotcha, missed the part that it's opt-in with a new flag. Thanks! :)

PR #1157, opened 2019-11-01
Enabling RTTI for Windows.

nicolasnoble:
See https://github.com/nodejs/node-gyp/issues/857 as to why this looks weird.

### DIFF
--- a/packages/grpc-native-core/templates/binding.gyp.template
+++ b/packages/grpc-native-core/templates/binding.gyp.template
@@ -154,10 +154,23 @@
             '_UNICODE',
             'NOMINMAX',
           ],
-          "msvs_settings": {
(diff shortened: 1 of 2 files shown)


packages/grpc-native-core/templates/binding.gyp.template
```diff
@@ -154,11 +154,24 @@
             '_UNICODE',
             'NOMINMAX',
           ],
-          "msvs_settings": {
-            'VCCLCompilerTool': {
-              'RuntimeLibrary': 1, # static debug
-            }
-          },
+          "configurations": [
```
murgatroid99: Configurations is a map, not a list. So `{}` not `[]`.

PR #1172, opened 2019-11-07
build(grpc-js): add full gts configuration and fix TSLint issues

merlinnot:
This PR introduces full gts configuration generated by `gts init` with manual changes to `tsconfig.json` to enable useful options from existing configuration. I also fixed all reported TSLint issues. `prettier.config.js` allows IDEs to automatically start formatting using a correct configuration.

### DIFF
--- a/packages/grpc-js/prettier.config.js
+++ b/packages/grpc-js/prettier.config.js
@@ -1,0 +1,5 @@
+module.exports = {
+  proseWrap: 'always',
+  singleQuote: true,
+  trailingComma: 'es5',
+};
--- a/packages/grpc-js/src/make-client.ts
+++ b/packages/grpc-js/src/make-client.ts
@@ -39,6 +39,7 @@ export interface MethodDefinition<RequestType, ResponseType> {
 }
 
 export interface ServiceDefinition {
+  // tslint:disable-next-line no-any
   [index: string]: MethodDefinition<any, any>;
 }
 
--- a/packages/grpc-js/src/load-balancer-round-robin.ts
+++ b/packages/grpc-js/src/load-balancer-round-robin.ts
@@ -107,7 +107,7 @@ export class RoundRobinLoadBalancer implements LoadBalancer {
   private calculateAndUpdateState() {
     if (this.subchannelStateCounts[ConnectivityState.READY] > 0) {
       const readySubchannels = this.subchannels.filter(subchannel => subchannel.getConnectivityState() === ConnectivityState.READY);
-      let index: number = 0;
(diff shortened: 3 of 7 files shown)


packages/grpc-js/prettier.config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  proseWrap: 'always',
+  singleQuote: true,
+  trailingComma: 'es5',
```
merlinnot: @murgatroid99 This is a default value, however we might want to go with `all` here to reduce noise in PRs. What's your take on this?

packages/grpc-js/src/make-client.ts
```diff
@@ -39,6 +39,7 @@ export interface MethodDefinition<RequestType, ResponseType> {
 }
 
 export interface ServiceDefinition {
+  // tslint:disable-next-line no-any
   [index: string]: MethodDefinition<any, any>;
```
merlinnot: We could go with the following here, but I'm not sure if it wouldn't be a breaking change (although a safer choice).

```suggestion
  [index: string]: MethodDefinition<unknown, unknown>;
```

murgatroid99: It's hard for me to tell right now what the impact of changing this would be. I think it would make sense to defer this change to when we can take a broader look at the possibility of changing `any` to `unknown` across the codebase.

packages/grpc-js/prettier.config.js
murgatroid99: I don't really like how code looks with the `all` value, and it's not entirely clear that it would result in valid TypeScript.

packages/grpc-js/src/make-client.ts
merlinnot: Sure, let's leave it as-is then.

packages/grpc-js/prettier.config.js
merlinnot: It most definitely results in valid TypeScript, as it's also a valid ECMAScript. ES6 added support for trailing commas (e.g. last property in the object). I generally tend to use `all`, as it gives cleaner PRs (if you add a property, only one line is changed, not two - current and previous with the comma).

I don't have a strong opinion on this tho, just a personal preference, so we can leave it as `es5` if you will.

murgatroid99: The documentation I found says that the `es5` option already adds commas after the last item in objects and arrays, and that the `all` option puts it in other places, including after the last function argument, which looks weird to me.

PR #1178, opened 2019-11-11
grpc-js: Change resolver to look at registeredResolvers when invoking getDefaultAuthority

### LINKED ISSUE #1054: grpc-js: Failed to parse target + adding a custom resolver
### Problem description
I'm trying to add a custom resolver to resolve hosts through Consul and am getting an error initializing the client: `Error: Failed to parse target consul://my-service`. It'd appear that the client by default will use the `DnsResolver` but the regex does not support any schemes besides `dns:`.

The way I'm trying to map my target to the resolver is through the target's scheme, but perhaps there's a better way to do this? Was using changes in https://github.com/grpc/grpc-node/pull/1049 and https://github.com/grpc/grpc-node/pull/1015 as reference.

### Reproduction steps

```js
registerResolver('consul:', ConsulResolver);

const client = new grpc.Client(
  'consul://my-service,
);
```

### Environment
 - OS X
 - Node version 10.16.0
 - Node installation via nvm
 - @grpc/grpc-js@0.6.2

jareddellitt:
Previously, it was calling `Object.keys` on a function which would never return the appropriate prefixes for registered resolvers and always fallback to the DNS resolver.

Fixes #1054

### DIFF
--- a/packages/grpc-js/src/resolver-dns.ts
+++ b/packages/grpc-js/src/resolver-dns.ts
@@ -140,7 +140,11 @@ function mergeArrays<T>(...arrays: T[][]): T[] {
   const result: T[] = [];
   for (
     let i = 0;
-    i < Math.max.apply(null, arrays.map(array => array.length));
+    i <
+    Math.max.apply(
+      null,
+      arrays.map(array => array.length)
+    );
     i++
   ) {
     for (const array of arrays) {
--- a/packages/grpc-js/src/resolver.ts
+++ b/packages/grpc-js/src/resolver.ts
@@ -125,7 +125,7 @@ export function createResolver(
  * @param target
  */
 export function getDefaultAuthority(target: string): string {
-  for (const prefix of Object.keys(registerDefaultResolver)) {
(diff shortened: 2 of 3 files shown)


thelinuxfoundation:

* User @jareddellitt isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

jareddellitt:
I signed it

packages/grpc-js/src/resolver-dns.ts
```diff
@@ -140,7 +140,11 @@ function mergeArrays<T>(...arrays: T[][]): T[] {
   const result: T[] = [];
   for (
     let i = 0;
-    i < Math.max.apply(null, arrays.map(array => array.length));
+    i <
```
jareddellitt: This is a formatting change that was required for my build to pass.

murgatroid99:
This fix is now out in version 0.6.12.

PR #1195, opened 2019-11-20
grpc-js: fix error messages truncating at commas

jamessharp:
At the moment, if an error message in grpc-js has a comma in it then it gets truncated just before the comma. This is presumably because commas are separators for multiple values of the same header, and `encodeURI` doesn't actually percent-encode a comma. Using `encodeURIComponent` does work (and that's what this PR does), but I'm not sure whether that is actually a bit overzealous.

### DIFF
--- a/packages/grpc-js/src/metadata-status-filter.ts
+++ b/packages/grpc-js/src/metadata-status-filter.ts
@@ -38,7 +38,7 @@ export class MetadataStatusFilter extends BaseFilter implements Filter {
       metadata.remove('grpc-status');
     }
     if (typeof metadataMap['grpc-message'] === 'string') {
-      details = decodeURIComponent(metadataMap['grpc-message'] as string);
+      details = decodeURI(metadataMap['grpc-message'] as string);
       metadata.remove('grpc-message');
     }
     return { code, details, metadata };
--- a/packages/grpc-js/src/metadata-status-filter.ts
+++ b/packages/grpc-js/src/metadata-status-filter.ts
@@ -38,7 +38,7 @@ export class MetadataStatusFilter extends BaseFilter implements Filter {
       metadata.remove('grpc-status');
     }
     if (typeof metadataMap['grpc-message'] === 'string') {
-      details = decodeURI(metadataMap['grpc-message'] as string);
+      details = decodeURIComponent(metadataMap['grpc-message'] as string);
       metadata.remove('grpc-message');
     }
     return { code, details, metadata };
--- a/packages/grpc-js/src/metadata.ts
+++ b/packages/grpc-js/src/metadata.ts
@@ -36,6 +36,10 @@ function isBinaryKey(key: string): boolean {
   return key.endsWith('-bin');
 }
 
+function isCustomMetadata(key: string): boolean {
+  return !key.startsWith('grpc-');
+}
+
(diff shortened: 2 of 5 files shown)


thelinuxfoundation:

* User @jamessharp isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

jamessharp:
I've signed the CLA

murgatroid99:
Can you make sure you signed the CLA linked in the second comment? The other one was supposed to be removed.

jamessharp:
Done 👍

murgatroid99:
On further reflection, I'm not sure this is the right solution. It's not clear from the spec if commas are required to be percent encoded, so we may need to assume that other implementations do not do so. I think a better way to approach this is to add a test of this scenario to `test/api` and make changes independently to make the `js client -> native server` and `native client -> js server` versions of the test work.

jamessharp:
I feel like something is missing from the spec (or, more likely, I'm missing something). Looking at https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md 

```
The value portion of Status-Message is conceptually a Unicode string description of the error, physically encoded as UTF-8 followed by percent-encoding. Percent-encoding is specified in RFC 3986 §2.1, although the form used here has different restricted characters. When decoding invalid values, implementations MUST NOT error or throw away the message. At worst, the implementation can abort decoding the status message altogether such that the user would received the raw percent-encoded form. Alternatively, the implementation can decode valid portions while leaving broken %-encodings as-is or replacing them with a replacement character (e.g., '?' or the Unicode replacement character).
```

I can't seem to find any reference as to what restricted characters "although the form used here has different restricted characters" is referring to. However, given "," is a delimiter then RFC3986 is explicit that it should be percent encoded. Perhaps we should explicitly just percent encode that?

Any idea why it was just `decodeURI` in the first place (that feels like an odd choice)?

Anyway - I'll get those tests in and we can have another look

murgatroid99:
I agree with your reading of the spec: I don't know what the set of restricted characters is. And fundamentally, it doesn't matter that much. The receiving end should be able to decode any percent-encoded values, so as long as we encode a superset of those restricted characters, it should be fine. I expect that both `encodeURI` and `encodeURIComponent` meet that criteria. And, for the same reason, it shouldn't really matter which `decode` function we use; in any case, we need to decode all percent-encoded characters.

The problem is that other servers may not be encoding commas, so we may still need to account for that in the client. The spec says that comma is a separator for `Custom-Metadata`, but it does not say that about `Status-Message`, so it would be reasonable for implementations not to assume that commas need to be encoded.

jamessharp:
That all makes sense

On digging further, and given what you've said about only custom-metadata being comma separated, then I think the commit above is a better solution. It looks like the problem was actually just on the client side with parsing the response. All headers were deemed to be comma separated.

My proposed fix is to only parse into metadata as comma-separated those headers that don't start 'grpc-' (which the spec says "should not be used by applications as Custom-Metadata"). I've also put in the interop tests

murgatroid99:
Don't worry about the build job failures. Those jobs have been flaky recently, and this change doesn't impact them.

PR #1196, opened 2019-11-21
Run make with verbose option

murgatroid99:


murgatroid99:
I don't think we're using this anymore

PR #1198, opened 2019-11-21
feat: allow headers other than authorization

bcoe:
Our upstream authentication library has gradually added more keys that provide meta-information to the upstream gRPC API.

We could not think of a reason why we were fetching only the `authorization` header, let's move away from this approach and simply populate all headers populated by `google-auth-library`.

CC: @murgatroid99, @alexander-fenster

### DIFF
--- a/packages/grpc-js/src/index.ts
+++ b/packages/grpc-js/src/index.ts
@@ -85,11 +85,11 @@ export interface OAuth2Client {
     callback: (
       err: Error | null,
       headers?: {
-        Authorization: string;
+        [index: string]: string;
       }
     ) => void
   ) => void;
-  getRequestHeaders: (url?: string) => Promise<{ Authorization: string }>;
+  getRequestHeaders: (url?: string) => Promise<{ [index: string]: string }>;
 }
 
 /**** Client Credentials ****/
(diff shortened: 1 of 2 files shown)


thelinuxfoundation:

* User @bcoe isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

alexander-fenster:
The same change needs to be done here for C-core gRPC:

https://github.com/grpc/grpc-node/blob/7a44ef8c01d3e722c48970476f906961c8a9c262/packages/grpc-native-core/src/credentials.js#L216-231

bcoe:
@murgatroid99 should we check `obj.hasOwnProperty`?  Or should we allow headers to potentially inherit from another object?

murgatroid99:
What we "allow" isn't really up to me. That object comes from google-auth-library, so it depends on what that library does. That being said, a simple solution here is to just use `for (const key of Object.keys(headers)`.

broady:
LGTM, but I think the changes @alexander-fenster describes should happen at the same time. It looks like it will need a little more refactoring.

bcoe:
@broady @murgatroid99 @alexander-fenster addressed initial review.

murgatroid99:
It probably makes sense to change the name of `getAuthorizationHeaderFromGoogleCredential`, now that it gets all of the headers.

PR #1206, opened 2019-12-03
Update google-auth-library in interop tests

murgatroid99:
I'm not really sure how to test this, because these test cases are only run in the cloud-to-prod auth test job, which doesn't run in this repo, and I'm not sure how to set up the environment for them.

### DIFF
--- a/test/interop/interop_client.js
+++ b/test/interop/interop_client.js
@@ -22,7 +22,7 @@ var fs = require('fs');
 var path = require('path');
 var grpc = require('../any_grpc').client;
 var protoLoader = require('../../packages/proto-loader');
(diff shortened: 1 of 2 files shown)


jtattermusch:
@murgatroid99 
as a note, here's how you could reproduce:
- check out grpc and grpc-node in two directories side by side.
- run `tools/run_tests/run_interop_tests.py --use_docker  ...`

the run_interop_tests.py expects node to be checked out in ../grpc-node

murgatroid99:
I ran that test script as described with this branch checked out in the grpc-node repo, and all of the `cloud_to_prod_auth` tests passed.

PR #1207, opened 2019-12-04
Put /bin in front of the PATH on Mac

murgatroid99:
Let's see if this makes the Mac jobs pass.


murgatroid99:
This didn't seem to work.

PR #1208, opened 2019-12-04
Attempting to mutate /bin/sh to stop the flakes.

nicolasnoble:


murgatroid99:
> ```
> /bin/sh: /bin/sh: cannot execute binary file
> ```

PR #1212, opened 2019-12-09
grpc-js: Add low-level call tracers

murgatroid99:
This will hopefully help with debugging individual calls that behave badly. In several places it prints message length instead of message contents to make the logs hopefully easier to follow while still providing something to match up between different parts of the log.

### DIFF
--- a/packages/grpc-js/src/call-stream.ts
+++ b/packages/grpc-js/src/call-stream.ts
@@ -28,6 +28,14 @@ import { ObjectDuplex, WriteCallback } from './object-stream';
 import { StreamDecoder } from './stream-decoder';
 import { ChannelImplementation } from './channel';
 import { Subchannel } from './subchannel';
+import * as logging from './logging';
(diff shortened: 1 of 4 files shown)


packages/grpc-js/src/call-stream.ts
```diff
@@ -177,6 +203,11 @@ export class Http2CallStream extends Duplex implements Call {
         (this.http2Stream as http2.ClientHttp2Stream).pause();
       }
     } else {
+      trace(
+        this.methodName +
+          ' unpushedReadMessages.push message of length ' +
```
nicolasnoble: Is this what you wanted it to say? This suspiciously looks like a copy/paste that hasn't been edited. That or the other one below.

murgatroid99: It looks right to me. They're both different, and each refers to the array being accessed on the following line.

PR #1222, opened 2019-12-13
grpc-js: Add unique ID number to call trace logs

murgatroid99:
This number is tracked in the channel file so that the channel can log that ID number when creating a call, linking the call number with the channel target and the method name. This change should make it easy to filter on the ID and see the entire lifetime of any single call.

### DIFF
--- a/packages/grpc-js/src/call-stream.ts
+++ b/packages/grpc-js/src/call-stream.ts
@@ -33,10 +33,6 @@ import { LogVerbosity } from './constants';
 
 const TRACER_NAME = 'call_stream';
 
-function trace(text: string): void {
-  logging.trace(LogVerbosity.DEBUG, TRACER_NAME, text);
-}
-
 const {
   HTTP2_HEADER_STATUS,
   HTTP2_HEADER_CONTENT_TYPE,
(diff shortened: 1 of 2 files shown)


packages/grpc-js/src/call-stream.ts
```diff
@@ -129,7 +125,8 @@ export class Http2CallStream extends Duplex implements Call {
     private readonly channel: ChannelImplementation,
     private readonly options: CallStreamOptions,
     filterStackFactory: FilterStackFactory,
-    private readonly channelCallCredentials: CallCredentials
+    private readonly channelCallCredentials: CallCredentials,
+    private readonly callNumber: number
```
nicolasnoble: Sort of design nit here: why does it have to be an actual number? Wouldn't a string be just as effective, and more flexible / extensible, if needs be?

murgatroid99: I think it depends on how this ID might get used in the future. If it's just used in logs, then it would be fine to use a string or number or whatever. But I think there are use cases where it makes more sense to have a number. Either way, I don't think it really matters. This is internal to the library, and can be changed in the future if there is a reason to do so.

PR #1232, opened 2020-01-09
Update submodule, test --no-duplicate-basename-check option

murgatroid99:
If the `node-gyp` information about that option is correct, the Mac build will fail because there are two files with the same name and libtool on Mac can't handle that.

### DIFF
--- a/packages/grpc-native-core/package.json
+++ b/packages/grpc-native-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "grpc",
-  "version": "1.24.0-pre1",
+  "version": "1.27.0-dev",
   "author": "Google Inc.",
   "description": "gRPC Library for Node",
   "homepage": "https://grpc.io/",
@@ -22,7 +22,7 @@
     "build": "node-pre-gyp build",
     "electron-build": "node-pre-gyp configure build --runtime=electron --disturl=https://atom.io/download/atom-shell",
     "coverage": "istanbul cover ./node_modules/.bin/_mocha test",
-    "install": "node-pre-gyp install --fallback-to-build --library=static_library",
+    "install": "node-pre-gyp install --fallback-to-build --library=static_library --no-duplicate-basename-check",
     "prepack": "git submodule update --init --recursive && npm install"
   },
   "bundledDependencies": [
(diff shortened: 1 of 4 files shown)


packages/grpc-native-core/package.json
```diff
@@ -22,7 +22,7 @@
     "build": "node-pre-gyp build",
     "electron-build": "node-pre-gyp configure build --runtime=electron --disturl=https://atom.io/download/atom-shell",
     "coverage": "istanbul cover ./node_modules/.bin/_mocha test",
-    "install": "node-pre-gyp install --fallback-to-build --library=static_library",
+    "install": "node-pre-gyp install --fallback-to-build --library=static_library --no-duplicate-basename-check",
```
mpminardi: I think you need a `--` before `--no-duplicate-basename-check` to pass it through to node-gyp.

e.g: `"install": "node-pre-gyp install --fallback-to-build --library=static_library -- --no-duplicate-basename-check",`

(this should let the build go farther but it seems like from the error text that the conflicting files might need to be renamed in the submodule?).

murgatroid99: I considered that but the error log contains this line:

```
gyp ERR! command "/home/kbuilder/.nvm/versions/node/v8.17.0/bin/node" "/home/kbuilder/.nvm/versions/node/v8.17.0/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "configure" "--fallback-to-build" "--library=static_library" "--no-duplicate-basename-check" "--module=/tmpfs/src/github/grpc-node/packages/grpc-native-core/src/node/extension_binary/node-v57-linux-x64-glibc/grpc_node.node" "--module_name=grpc_node" "--module_path=/tmpfs/src/github/grpc-node/packages/grpc-native-core/src/node/extension_binary/node-v57-linux-x64-glibc" "--napi_version=4" "--node_abi_napi=napi" "--napi_build_version=0" "--node_napi_label=node-v57"
```

which appears to indicate that that option was in fact passed to gyp though node-gyp.

murgatroid99: I have now tried this idea. It does not change the behavior, and it appears to remove that option from the logged node-gyp command line

mpminardi: Oh that is odd, when I make this change locally and run `npm install` from within `packages/grpc-native-core` I get to a point where I see `AssertionError: $(builddir)/base_dynamic_annotations.a != $(builddir)/libabsl/base_dynamic_annotations.a` instead.

PR #1233, opened 2020-01-09
Add absl support

veblush:
Updated `grpc-native-core` project templates to add abseil targets properly.

Related to https://github.com/grpc/grpc/pull/20184.

### DIFF
--- a/packages/grpc-native-core/package.json
+++ b/packages/grpc-native-core/package.json
@@ -68,7 +68,7 @@
     "deps/grpc/include/grpc/**/*.h",
     "deps/grpc/src/core/**/*.{c,cc,h}",
     "deps/grpc/src/boringssl/err_data.c",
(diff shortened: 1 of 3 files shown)


murgatroid99:
I have a similar change in #1232. Unfortunately there are other problems I have to fix before it will actually work.

murgatroid99:
In the PR you linked to, the Node test and build jobs display a passing result. So, what prompted you to make those changes in the `binding.gyp.template` file?

veblush:
It seems that we have a test with a head of grpc-node and changes from gRPC's PR. As you saw, I didn't have any failure with the PR but I thought that I needed some change on node.js like I did for other wrapped languages.

veblush:
When I tried to build grpc-node, however, it didn't work well so I ended up with this PR. I think there must be a subtle difference, then?

PR #1243, opened 2020-01-24
grpc-js: Add HTTP CONNECT support, i.e. egress proxy support

murgatroid99:
The implementation here is based pretty closely on the corresponding core code in https://github.com/grpc/grpc/blob/master/src/core/ext/filters/client_channel/http_proxy.cc. The code for establishing the connection is loosely based on [this section of the `tunnel-agent` package](https://github.com/request/tunnel-agent/blob/master/index.js#L117-L182), which is used by the `request` package for establishing HTTP CONNECT tunnels.

With this change the library handles the environment variables `grpc_proxy`, `https_proxy`, `http_proxy`, `no_grpc_proxy`, and `no_proxy`.

I am considering this an experimental feature for now because I am not familiar with environments with those kinds of proxies and I don't know how to test it.

### DIFF
--- a/packages/grpc-js/src/http_proxy.ts
+++ b/packages/grpc-js/src/http_proxy.ts
@@ -1,0 +1,160 @@
+/*
+ * Copyright 2019 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


murgatroid99:
This PR is currently not correct. The `socket` option that I use is actually only used by `tls` as the underlying connection. The http2 module does not use it at all, so this implementation will not use the proxy for plaintext connections.

Issue #258, opened 2018-04-05
Unable to connect to Unix socket using node-grpc client


gcarling:
I am trying to communicate with a server that is listening on a unix socket - if I use TCP I have no issues, but after much effort have been unable to get it working with the unix socket. Code is 

```
    const BIND_SOCKET = '/tmp/collector.sock';
    const BIND_TCP = ':9832'; // works fine if I use this for client + server
    const PROTO_PATH = './Collector.proto';
    const collector = grpc.load({ root: __dirname, file: PROTO_PATH }).collectorv1;

    this.client = new collector.Collector(BIND_SOCKET, grpc.credentials.createInsecure());
```

I have tried many variations on the socket, such as `unix:/tmp/collector.sock` and many others I have found online. However none of them work and my write call simply hangs forever, regardless of whether the server is active and listening on the port.

I am using `grpc` version `1.10.1`

murgatroid99:
I am pretty sure the node library does not currently support unix domain sockets.

gcarling:
When I was debugging this I got up to here - line 485 of `client.js`

`this.$channel = new grpc.Channel(address, credentials, options);`

in this context, `grpc` has been required like this - 

`var grpc = require('./grpc_extension');`

My understanding was there was some binary outside of JS that was being used to make the actual connection. This leads me to believe that any behavior supported by that library should be supported here? Maybe I am not fully understanding, but my point is that it does not appear as though the node client is actually the one making the connection. I could be wrong.

gcarling:
@murgatroid99 Can you confirm for me that `node-grpc` uses the C++ library under the hood? I saw this issue (https://github.com/grpc/grpc/issues/6504) which confirms for me that the C++ version supports unix sockets. Not sure if you are using this library or a different C++ binary as the base for connection, in any event hoping you can give me some info on that so I can debug further on my own. This is pretty critical to our performance so I am willing to put time into this, even open a PR if necessary.

nicolasnoble:
[...]

PR #1244, opened 2020-01-29
grpc-js: Support Unix domain sockets and Named pipes

raksooo:
This is an attempt at implementing support for UDS and Named pipes as discussed in #258. I'm not sure if the design of this solution is the way to go here or if there's a better way to do it. 

As I understand the http2-library, it isn't able to create a connection over a Unix domain socket or a Named pipe. To solve that I opened a connection using the net-library and supplied that connection to `http2.connect`, as in the example by @jcantwell-JC in the issue mentioned above.

I added a method to the resolver which decides if the connection should be an IPC connection, which works well, but I'm not sure if this is the best way to accomplish this.

### DIFF
--- a/packages/grpc-js/src/resolver-dns.ts
+++ b/packages/grpc-js/src/resolver-dns.ts
@@ -312,6 +312,10 @@ class DnsResolver implements Resolver {
     }
     throw new Error(`Failed to parse target ${target}`);
   }
+
+  static isIPC(): boolean {
+    return false;
+  }
 }
 
 /**
--- a/packages/grpc-js/src/resolver-named-pipe.ts
+++ b/packages/grpc-js/src/resolver-named-pipe.ts
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2019 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  Resolver,
+  ResolverListener,
+  registerResolver,
+  registerDefaultResolver,
+} from './resolver';
+
+class NamedPipeResolver implements Resolver {
+  private addresses: string[] = [];
+  constructor(target: string, private listener: ResolverListener) {
+    this.addresses = [target];
+  }
+  updateResolution(): void {
+    process.nextTick(
+      this.listener.onSuccessfulResolution,
+      this.addresses,
+      null,
+      null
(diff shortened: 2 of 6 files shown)


thelinuxfoundation:

* User @raksooo isn't covered by a CLA. They will need to complete the form at https://identity.linuxfoundation.org/projects/cncf

_Regards,
CLA GitHub bot_

murgatroid99:
I'm sorry, we're unlikely to accept this. The API of this library is supposed to match the API of the `grpc` library, and that one does not have the '//./pipe/` URI scheme. The valid URI schemes are listed in [this document](https://github.com/grpc/grpc/blob/master/doc/naming.md).

On the plus side, I have a different change in the works to fix UDS support and the `unix:` scheme, and once that works, it should be possible to pass a named pipe path using that scheme. For example, `unix:///./pipe/pipename` should work.

murgatroid99:
The alternate fix I referred to is in #1245.

faern:
The documentation you link to says

> unix:path or unix://absolute_path -- Unix domain sockets (Unix systems only)

Would it then not be strange to support Windows named pipes through the `unix:` prefix? Don't get me wrong, we do need named pipes support, but this felt like a strange way to solve it 🤔

faern:
It surprises me that gRPC itself specifies address naming schemes at all. Quite limiting to let the protocol library handle the transport layer or express opinions about it. Even more so if it imposes these limitations/requirements on other implementors.

murgatroid99:
The point about the `unix:` scheme applying specifically to UDS is fair, and I'm not actually sure if it works that way in the other libraries. But the fact is that in terms of name transformation and connection establishment, the Node APIs do not distinguish between Unix domain sockets and Windows named pipes. That means that there's no substantial difference between code that supports UDS and code that supports named pipes.

Regarding naming schemes, I want to note that that specification does not constrain the gRPC protocol, just the official gRPC libraries. The linked document describes how the gRPC libraries handle and resolve the names passed to them.

I don't see the issue with the gRPC libraries handling name resolution and the underlying transport. The Node http2 library automatically does DNS resolution and establishes the underlying TLS or TCP connection used by the http2 session. That seems like the same thing to me.

faern:
Having all IPC under the `unix:` prefix is fine. It should just have been named something else IMO. After all, UDS *is* named pipes for unix platforms. So named pipes is the broader term.

I agree it's handy that libraries do have handy helper functions for the most common use cases. But to actually be flexible they should also accept an arbitrary IO stream to operate over. So people with not-default use cases can just plug in their transports. So I agree it's nice that the official libraries can handle a number of URIs out of the box, but it should not limit the user to those IMO.

murgatroid99:
As I am discovering in my own change, even the built in http2 library doesn't actually make it that easy to just plug in a transport. If you look at the latest version of the PR I linked to, you can see that I still need to set the `createConnection` option, but only sometimes.

In the longer run it may be possible for grpc-js to eventually support plugging in transports and/or resolvers, but that's another issue of compatibility with the other implementation.

raksooo:
Thanks for the feedback! It didn't cross my mind to use "unix:" for Windows Named pipes although it's pretty clear that it works when looking at the code. I don't know where but it would be nice if it is documented somewhere.

Your PR looks really good and works well. Thanks for taking the time to implement this! Feel free to close this PR.

PR #1257, opened 2020-02-07
Add fan-out configs and scripts for windows and linux

murgatroid99:
This change actually does a few related things:

 - Move templates to the root directory and add templates for Kokoro configs
 - Modify the `generate_packages.sh` script to generate those templates, including adding a `gen_build_yaml.py` script
 - Generate and commit the results of those templates
 - Move the Alpine build Dockerfile to the tools/release directory, to be with the others, and update references to it


packages/grpc-native-core/tools/buildgen/gen_build_yaml.py
```diff
@@ -0,0 +1,115 @@
...
+# limitations under the License.
+
+from __future__ import print_function
+import re
+import os
+import sys
+import yaml
+
+node_versions = ["4", "5", "6", "7", "8", "9", "10", "11", "12", "13"]
+electron_versions = ["1.0", "1.1", "1.2", "1.3", "1.4", "1.5", "1.6", "1.7", "1.8", "2.0", "3.0", "3.1", "4.1", "4.2", "5.0", "6.0", "6.1", "7.0", "7.1"]
```
nicolasnoble: Latest electron release is 8.0, so might as well?

murgatroid99: Done. And now that commit demonstrates what a change to add a version will look like.

PR #1259, opened 2020-02-10
Electron 8 binaries

Jakeeyturner:
Signed-off-by: Jake Turner <jaketurner25@live.com>


murgatroid99:
We're moving to a new build system with a new way of handling the set of versions. So this change is likely to not be applicable by the time we do our next release.

PR #1285, opened 2020-03-03
grpc-js: Always return IPv6 addresses from DNS resolver

murgatroid99:
With the new representation of addresses from #1245, I think that we will no longer be impacted by the bug that prevented us from using IPv6 addresses on older Node versions.


murgatroid99:
Apparently Node on MacOS can't resolve `loopback6.unittest.grpc.io`, and maybe also `loopback46.unittest.grpc.io` (I can't tell what's going on with that test).

murgatroid99:
A rebuild got rid of that error, so I assume it was some kind of transient failure.

PR #1287, opened 2020-03-05
grpc-js: Use resolver to bind server ports

murgatroid99:
Also, have bindAsync report a port number of 0 on error, and in the other implementation, treat 0 as an error. This should also make UDS servers work.


nicolasnoble:
It doesn't look like the tests are happy.

PR #1291, opened 2020-03-06
grpc-js: Use dns.lookup and dns.resolve to cover problems in each other

### LINKED ISSUE #1289: grpc-js resolver test using loopback6.unittest.grpc.io is flaky
https://source.cloud.google.com/results/invocations/132ebe45-7193-4a46-9427-a97deb6c4c7d/targets

The test with `loopback46.unittest.grpc.io` also fails there, but it's not clear if those failures are independent.

murgatroid99:
Hopefully fixes #1289


murgatroid99:
I don't think this is the right way to go.

PR #1297, opened 2020-03-10
grpc-js: tests: skip checking for IPv6 addresses from DNS servers

### LINKED ISSUE #1289: grpc-js resolver test using loopback6.unittest.grpc.io is flaky

murgatroid99:
This should fix #1289. I think this is the right way to move forward until changes can be made to Node itself to allow us to get AAAA records more consistently using `dns.lookup`.


murgatroid99:
nodejs/node#32183 has some changes that should allow us to enable these tests again.

PR #1313, opened 2020-03-24
Allow SSL credential parameters to be specified using (PEM) string.

kskalski:
This enables use of ssl credentials in browser-only environment (no Buffer available), as well as widens the supported API in nodejs environment thanks to 'tls' module supporting both kinds of parameters.

### DIFF
--- a/packages/grpc-js/src/channel-credentials.ts
+++ b/packages/grpc-js/src/channel-credentials.ts
@@ -21,9 +21,9 @@ import { CallCredentials } from './call-credentials';
 import { CIPHER_SUITES, getDefaultRootsData } from './tls-helpers';
 
 // tslint:disable-next-line:no-any
-function verifyIsBufferOrNull(obj: any, friendlyName: string): void {
-  if (obj && !(obj instanceof Buffer)) {
-    throw new TypeError(`${friendlyName}, if provided, must be a Buffer.`);
+function verifyIsBufferOrStringOrNull(obj: any, friendlyName: string): void {
+  if (obj && !(obj instanceof Buffer) && typeof obj !== 'string') {
+    throw new TypeError(`${friendlyName}, if provided, must be a Buffer or string.`);
   }
 }
 
@@ -48,11 +48,12 @@ export type CheckServerIdentityCallback = (
   cert: Certificate
 ) => Error | undefined;
 
-function bufferOrNullEqual(buf1: Buffer | null, buf2: Buffer | null) {
+function bufferOrStringOrNullEqual(buf1: Buffer | string | null, buf2: Buffer | string | null) {
   if (buf1 === null && buf2 === null) {
     return true;
   } else {
-    return buf1 !== null && buf2 !== null && buf1.equals(buf2);
+    return buf1 !== null && buf2 !== null && (buf1 === buf2 ||
+        (buf1 instanceof Buffer && buf2 instanceof Buffer && (<Buffer> buf1).equals(<Buffer> buf2)));
   }
 }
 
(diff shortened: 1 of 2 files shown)


murgatroid99:
I don't think it makes sense to try to make this part of the API compatible with browsers, when browsers don't even support the full http2 functionality that we need to implement the gRPC protocol. In addition, if and when this does work on a browser, I'd prefer that users use a browser shim for `Buffer`.

kskalski:
Note that this patch provides richer functionality for nodejs context and makes API in par with 'tls' module - providing certificates by string creates lower entry point during development (e.g. when embedding the PEM in source for testing) and less surprises as to supported parameters for someone experienced with 'tls' node module.

But I also find the assertions that http/2 is not fully functional in browser context and that you do not target grpc-js in browser use-case as surprising. I thought http/2 was primarily created for the browser to communicate with backends in a faster and more flexible way and this is what technologies like websockets use extensively. In fact in my tests with this patch I can get an encrypted bi-directional gRPC stream working just fine between Electron 8 renderer (chromium) and .NET server.

Anyway, if you actually want to restrict use of grpc-js in browser, note following:
- the current API doesn't do it as I guess following your suggestion one could use Buffer shim
- it is a rather obscure way to do it here, I would expect some explicit assertions when creating channel / call, right now someone not using encryption would get a working communication, but then switching to encryption would be surprised that API intentionally makes it harder or impossible
- I know grpc-js is still experimental/beta, but apart from not being production ready I found no mention that it is designed as node-only library... actually not knowing nodejs popularity one would assume grpc-js (as purely JS) is browser-only technology ;-)

murgatroid99:
> Note that this patch provides richer functionality for nodejs context and makes API in par with 'tls' module - providing certificates by string creates lower entry point during development (e.g. when embedding the PEM in source for testing) and less surprises as to supported parameters for someone experienced with 'tls' node module.

With the existing API you can create a channel credentials object from a string by simply using `Buffer.from` to get the corresponding `Buffer`. I don't think it's necessary to make a change here so that users can avoid that small bit of code.

> But I also find the assertions that http/2 is not fully functional in browser context and that you do not target grpc-js in browser use-case as surprising. I thought http/2 was primarily created for the browser to communicate with backends in a faster and more flexible way and this is what technologies like websockets use extensively.

Browsers support enough of HTTP/2 to handle their own use cases: retrieving multiple resources from a server concurrently with a single connection, handling server push, etc. What they don't do is provide a JavaScript API that completely implements the HTTP/2 protocol. In particular, existing browser HTTP/2 JavaScript APIs do not provide access to trailing headers, which gRPC uses.

Websockets is an entirely separate protocol, created before anyone even started designing HTTP/2.

> In fact in my tests with this patch I can get an encrypted bi-directional gRPC stream working just fine between Electron 8 renderer (chromium) and .NET server.

Electron combines both Node.js and Chromium. I am sure that in that test, grpc-js was using Node's http2 APIs. This is an intended and supported usage of grpc-js.

> Anyway, if you actually want to restrict use of grpc-js in browser, note following...

It is currently not possible to use grpc-js in a browser. If you try it, it will fail to load the `http2` module, and there is no browser shim for that module because not all of the functionality is available. If and when someone creates a complete working browser shim for the `http2` module, grpc-js probably will work in the browser and that's fine.

> I know grpc-js is still experimental/beta, but apart from not being production ready I found no mention that it is designed as node-only library... actually not knowing nodejs popularity one would assume grpc-js (as purely JS) is browser-only technology

It's true that nothing explicitly states that grpc-js only works with Node, but it's in the "grpc-node" repository, and it's a reimplementation of the `grpc` library (which is very much Node-only), and it uses multiple Node-specific APIs, and it's distributed using npm. Plus, the README says "Node 10 is recommended. The exact set of compatible Node versions can be found..." so that should pretty clearly indicate that it's not browser-only.

kskalski:
Oh, thanks, that makes sense. 
Now I understand the difference between grpc-js and grpc-web and it seems for the browser client I would have to use grpc-web, even if grpc-js works fine in Electron environment. It's a pity that grpc-web doesn't support bidi streams, but I can work around that.

I still think the API is better accepting string PEM, especially at such low cost of implementation (just passing the same options to 'tls' module), but it's clearly a bit of a complication, so I can understand why you prefer it as it is now.

PR #1314, opened 2020-03-26
Use make from homebrew in Mac tests

murgatroid99:
Based on a suggestion from [this StackOverflow comment](https://stackoverflow.com/questions/60854726/how-can-i-fix-this-bizarre-no-such-file-or-directory-error-when-building-with#comment107671441_60854726), maybe this will fix our Mac issues.


murgatroid99:
First run: all tests passed

murgatroid99:
Second run: all Mac builds succeeded again

PR #1322, opened 2020-03-30
PickFirstLoadBalancer: Only change to TRANSIENT_FAILURE when all channels fail

### LINKED ISSUE #664: Failure in environments without IPv6 support
### Problem description
Tried to implement GRPC on a project using FreeBSD 11.2 and it throw a core dump, then decided to try the demo project from the grpc repository, core dump, updated the packages to the latest versions, core dump... It compiles with node-gyp without errors apparently. The same steps on Linux works.

### Reproduction steps
git clone https://github.com/grpc/grpc
cd grpc/examples/node
npm install
cd dynamic_codegen
node greeter_server.js

### Environment
 - OS name, version and architecture: FreeBSD 11.2-RELEASE-p4
 - Node version: v8.12.0
 - Node installation method: pkg install node8
 - If applicable, compiler version: clang 6.0.0
 - Package name and version: gRPC@1.17.0-pre1

### Additional context
```
user@server:/home/user/grpc/examples/node/static_codegen % node greeter_server.js 
Segmentation fault (core dumped)
ls -la
-rw-------  1 user  user  64303104 Nov 29 22:22 node.core
```

### LINKED ISSUE #692: Error: 14 UNAVAILABLE: TCP Read failed
### Problem description
Conversation Error: { Error: 14 UNAVAILABLE: TCP Read failed
    at Object.exports.createStatusError (/relay/node_modules/grpc/src/common.js:91:15)
    at ClientDuplexStream._emitStatusIfDone (/relay/node_modules/grpc/src/client.js:233:26)
    at ClientDuplexStream._receiveStatus (/relay/node_modules/grpc/src/client.js:211:8)
    at Object.onReceiveStatus (/relay/node_modules/grpc/src/client_interceptors.js:1306:15)
    at InterceptingListener._callNext (/relay/node_modules/grpc/src/client_interceptors.js:568:42)
    at InterceptingListener.onReceiveStatus (/relay/node_modules/grpc/src/client_interceptors.js:618:8)
    at /relay/node_modules/grpc/src/client_interceptors.js:1123:18
  code: 14,
  metadata: Metadata { _internal_repr: {} },
  details: 'TCP Read failed' }

### Environment
 - Raspbian GNU/Linux 9 (4.14.90-v7+ #1183 SMP Fri Dec 21 14:03:50 GMT 2018 armv7l GNU/Linux)
 - Node version v8.15.0
 - Package name and version smarthome-relay@1.3.8 /relay
├─┬ google-assistant@0.5.2
│ └── grpc@1.17.0  deduped
└── grpc@1.17.0

### Additional context
it happens randomly but it start doing that since ~1 month back

orgads:
This happens when a device doesn't have IPv6 address, but the endpoint
address is resolved to both IPv4 and IPv6 addresses.

The automatic attempt to reconnect that fails is detected as another channel
failure, which leads the load balancer to abort the connection.

Fixes #664
Fixes #692

### DIFF
--- a/packages/grpc-js/src/load-balancer-pick-first.ts
+++ b/packages/grpc-js/src/load-balancer-pick-first.ts
@@ -173,8 +173,8 @@ export class PickFirstLoadBalancer implements LoadBalancer {
             if (this.subchannelStateCounts[ConnectivityState.CONNECTING] > 0) {
               newLBState = ConnectivityState.CONNECTING;
             } else if (
-              this.subchannelStateCounts[ConnectivityState.TRANSIENT_FAILURE] >
-              0
+              this.subchannelStateCounts[ConnectivityState.TRANSIENT_FAILURE] ==
+              this.subchannels.length
             ) {
               newLBState = ConnectivityState.TRANSIENT_FAILURE;
             } else {


murgatroid99:
Both #664 and #692 are bugs on the other grpc implementation.

I think you're missing that the load balancer will only report an error if **all** of its subchannels fail to connect.

orgads:
Here is [my log](https://gist.github.com/orgads/f64abee06b64b852a20d934e0dd32224). Can be reproduced on a linux machine with `disable_ipv6=1`, and by introducing network latency using: `tc qdisc add dev wlan0 root netem delay 1000ms`.

A few observations:

1. IP4 and IP6 are attempted in parallel.
2. IP6 receives `TRANSIENT_FAILURE` immediately, IP4 is connecting.
3. IP6 has a backoff retry, so it goes back to `CONNECTING` and `TRANSIENT_FAILURE` over and over again.
4. IP4 has a 250ms timeout, after which it switches to `IDLE`. When it does, `triedAllSubchannels` in `PickFirstLoadBalancer` becomes true.
5. Now IP6 receives `TRANSIENT_FAILURE`. So the LB state becomes `TRANSIENT_FAILURE` (because its count is >0), and then `updateState(UnavailablePicker())` is called, without waiting for IP4 to complete its backoff retries.

I'm not really sure what's the right way to fix it. This PR is a semi-blind shot. Can you suggest a better solution?

murgatroid99:
The real bug in that log is the transition to IDLE in step 4. That subchannel should keep connecting there. I think I understand why it doesn't, but I'm not exactly sure how to change the logic to fix that properly.

orgads:
Pushed a different approach. Please review it.

murgatroid99:
I don't think this is the right approach either. I think #1328 is the right way to address this.

murgatroid99:
I'm going to close this and go with #1328 as the solution instead.

PR #1323, opened 2020-03-30
PickFirstLoadBalancer: Correctly initialize state counters

orgads:


murgatroid99:
This isn't exactly right. `createSubchannel` can actually get an existing subchannel. That's why there's that loop a few lines later that checks if any subchannel is already READY. The right thing to do here is to loop over the subchannels and count their states, after the READY check loop and before the loop after that.

orgads:
Good observation. Thanks, done.

PR #1328, opened 2020-03-31
grpc-js: Don't time out connection attempts

murgatroid99:
The purpose of the backoff timer is to enforce a minimum time between connection attempts. Transitioning from from CONNECTING to IDLE does not serve that purpose. I think that we can trust that Node will eventually time out connection attempts on its own, and that will result in a proper transition to TRANSIENT_FAILURE.

The reason I want to do this is to fix the problem described in https://github.com/grpc/grpc-node/pull/1322#issuecomment-606237127.


orgads:
LGTM. Thanks.

PR #1341, opened 2020-04-07
grpc-js: Make some headers conform to what the other library does

murgatroid99:
This changes a few headers:

 - `:authority` now includes the port number if it was specified in the channel target
 - `accept-encoding: identity, gzip` is now sent
 - `grpc-encoding: identity` is now omitted

I also added more detailed header tracing. For technical reasons the initial client header trace does not have a call number. We can plumb that through later but for now this is better than nothing.

I also added separate explicit tracing for receiving status code and details in the trailing metadata. Technically it's redundant with the trailing metadata tracing but I think it's easier to see when traced separately like that.

### DIFF
--- a/packages/grpc-js/src/resolver-dns.ts
+++ b/packages/grpc-js/src/resolver-dns.ts
@@ -298,10 +298,16 @@ class DnsResolver implements Resolver {
       IPV6_REGEX.exec(target) ||
       IPV6_BRACKET_REGEX.exec(target);
     if (ipMatch) {
+      if (ipMatch[2]) {
+        return ipMatch[1] + ':' + ipMatch[2];
+      }
       return ipMatch[1];
     }
     const dnsMatch = DNS_REGEX.exec(target);
     if (dnsMatch) {
+      if (dnsMatch[2]) {
+        return dnsMatch[1] + ':' + dnsMatch[2];
+      }
       return dnsMatch[1];
     }
     throw new Error(`Failed to parse target ${target}`);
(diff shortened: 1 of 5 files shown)


packages/grpc-js/src/subchannel.ts
```diff
@@ -298,8 +298,6 @@ export class Subchannel {
           return checkServerIdentity(sslTargetNameOverride, cert);
         };
         connectionOptions.servername = sslTargetNameOverride;
-      } else {
-        connectionOptions.servername = getDefaultAuthority(this.channelTarget);
```
alexander-fenster, line 302: This does not seem to be in scope of this PR?

packages/grpc-js/src/resolver-dns.ts
```diff
@@ -298,10 +298,16 @@ class DnsResolver implements Resolver {
       IPV6_REGEX.exec(target) ||
       IPV6_BRACKET_REGEX.exec(target);
     if (ipMatch) {
+      if (ipMatch[2]) {
```
alexander-fenster, line 301: Not sure if the resolver changes are related to the headers problem being fixed.

packages/grpc-js/src/subchannel.ts
murgatroid99: That change is necessary because of the `dns-resolver` change.

packages/grpc-js/src/resolver-dns.ts
murgatroid99: The output of this function determines the value of the `:authority` header.

PR #1342, opened 2020-04-08
grpc-js: Fix status check when connecting to proxy

murgatroid99:
It turns out that `http.STATUS_CODES` does not actually map status code names to status code numbers as I thought.

I also upgraded the verbosity of some logs so that if one of the environment variables is set and the proxy cannot be used, the user should always see at least one error log.


murgatroid99:
@alexander-fenster I agree with you. I don't know how to test this feature.

Issue #931, opened 2019-06-28
[question] How to generate `require('@grpc/grpc-js')` instead of `require('grpc')`


kittaakos:
I am not sure if this is the correct place to ask...

I would like to use `@grpc/grpc-js` instead of `grpc` in the generated JS code to overcome an [issue inside a bundled electron](https://github.com/grpc/grpc-node/issues/624) application.

Is it possible to staticly generate the JS code with `require('@grpc/grpc-js')` instead of `require('grpc')`. My current approach is to generate it with `require('grpc')`, then parse the generated files and replace the `require`s in a manual step 😞 Is there a flag I can tweak to adjust the generator?

I was unable to find the solution, so I set up a [bare minimum example](https://github.com/kittaakos/gen-grpc-js), perhaps someone can advise.

Thank you!

murgatroid99:
Version 1.8.0 of `grpc-tools` added the `generate_package_definition` option to the `grpc_out` argument. It can be used like this: `--grpc_out=generate_package_definition:<path>`. This will generate a package definition object that can be passed to the `loadPackageDefinition` function in both grpc libraries.

kittaakos:
Thank you, @murgatroid99. Is there an example somewhere I can check? I have [found the proposal](https://github.com/grpc/proposal/blob/master/L23-node-protobufjs-library.md) of this new feature, but not a single test or sample.

murgatroid99:
No, there is not currently an example for this.

paambaati:
@murgatroid99 I have a very similar usecase where I'm trying to statically generate pb files and TypeScript bindings, and simply replacing `grpc` with `@grpc/grpc-js` does not work for me (`ServiceDefinition`, `MethodDefinition`, etc. aren't defined/exported).

Is this possible currently?

ctrlshp:
I would really like to try this as well. I have added the `generate_package_definition` option but the command crashes with this error message "Plugin output is unparseable". I am using  grpc-tools 1.8.0 on macOS 10.13.6 and node 12.10.0. Without the `generate_package_definition` option, it works just fine, but of course it doesn't generate a static proto definition object that can be used with `@grpc/grpc-js`.

Can anybody help ?

Here is the full command :

```
npx grpc_tools_node_protoc \
    --js_out=import_style=commonjs,binary:./src/grpc/proto \
    --grpc_out=generate_package_definition:./src/grpc/proto \
[...]

PR #1343, opened 2020-04-09
grpc-js: adjust ts definitions to equal native-core

Patrick-Remy:
## Why
I tried to use the @grpc/grpc-js package with typescript definitions generated by [grpc_tools_node_protoc_ts](https://github.com/agreatfool/grpc_tools_node_protoc_ts). But the miss export of type `MethodDefinition` (which is exported in [native-core's index.d.ts](https://github.com/grpc/grpc-node/blob/master/packages/grpc-native-core/index.d.ts)) made it impossible. Also I adjusted the ServiceDefinition. 

This will also help to for #931

## Changes
Export `MethodDefinition` in index.ts and add generic `ImplementationType`
to `ServiceDefinition`.

### DIFF
--- a/packages/grpc-js/src/make-client.ts
+++ b/packages/grpc-js/src/make-client.ts
@@ -18,6 +18,7 @@
 import { ChannelCredentials } from './channel-credentials';
 import { ChannelOptions } from './channel-options';
 import { Client } from './client';
+import { UntypedServiceImplementation } from './server';
 
 export interface Serialize<T> {
   (value: T): Buffer;
@@ -49,9 +50,9 @@ export interface MethodDefinition<RequestType, ResponseType>
   extends ClientMethodDefinition<RequestType, ResponseType>,
     ServerMethodDefinition<RequestType, ResponseType> {}
 
-export interface ServiceDefinition {
+export type ServiceDefinition<ImplementationType = UntypedServiceImplementation> = {
   // tslint:disable-next-line no-any
-  [index: string]: MethodDefinition<any, any>;
+  readonly [index in keyof ImplementationType]: MethodDefinition<any, any>;
 }
 
 export interface ProtobufTypeDefinition {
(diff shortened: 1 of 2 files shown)


Patrick-Remy:
CLA is currently unavailable.

packages/grpc-js/src/make-client.ts
```diff
@@ -49,9 +50,9 @@ export interface MethodDefinition<RequestType, ResponseType>
   extends ClientMethodDefinition<RequestType, ResponseType>,
     ServerMethodDefinition<RequestType, ResponseType> {}
 
-export interface ServiceDefinition {
+export type ServiceDefinition<ImplementationType = UntypedServiceImplementation> = {
```
murgatroid99, line 53: Why change `interface` to `type` here?

Also, I don't understand how this would be used. Can you provide an example of using this type definition with and without the type parameter?

Patrick-Remy: A `type` is needed, otherwise the typescript compiler throws an error for `interface`:
> A computed property name must be of type 'string', 'number', 'symbol', or 'any'.ts(2464)

See this PR https://github.com/microsoft/TypeScript/pull/26797 

If I generate code from my proto files via  `grpc_tools_node_protoc_ts`, I get these service definitions:
```ts
interface IMyCustomService extends grpc.ServiceDefinition<grpc.UntypedServiceImplementation> {
  get: IMyCustomService_IGet;
}

interface IMyCustomService_IGet extends grpc.MethodDefinition<myservice_pb.GetRequest, google_protobuf_empty_pb.Empty> {
  prop: boolean; // false
  propDeserialize: grpc.deserialize<google_protobuf_empty_pb.Empty>;
}
```

I just looked at https://github.com/grpc/grpc-node/blob/4823d97aba704886a0eff755ff78383338cb6fe6/packages/grpc-native-core/index.d.ts#L315-L317 and added the default ImplementationType to be compatible to the interface used before.

murgatroid99: I did not remember that we had that definition in the other library. In that case, if the tests pass, I'm fine with this change.

PR #1345, opened 2020-04-09
grpc-js fix linting

Patrick-Remy:
Remove deprecated tslint config file, upgrade to gts v2 (currently alpha) and adjust config to match requirements (mostly allowing any), fix most of the linting issues.

Currently there are 69 open problems (mostly `no-unused-vars` in parameters), including rule `node/no-unpublished-require` seems to ignore the `files` property from package.json.

Should `no-unused-vars` be ignored globally or for every method itself?

After fixing all of them, linting should be added to the CI.


packages/grpc-js/.eslintrc
```diff
@@ -0,0 +1,10 @@
+{
+  "root": true,
+  "extends": "./node_modules/gts",
+  "rules": {
+    "@typescript-eslint/no-explicit-any": "off",
```
murgatroid99, line 5: I don't think this is right. `any` is used in several places, but it is used minimally and purposefully and I think it's valuable to need to add the comment to ignore the rule in every place where `any` is used.

packages/grpc-js/.eslintrc
```diff
@@ -0,0 +1,10 @@
...
+    "node/no-unpublished-import": ["error", {
+        "tryExtensions": [".ts", ".js", ".json", ".node"]
```
murgatroid99, line 7: We might want a `convertPath` option here. I'm not sure because I'm not entirely clear on why this change makes it work.

packages/grpc-js/src/resolver-dns.ts
```diff
@@ -323,17 +331,23 @@ export function setup(): void {
   registerDefaultResolver(DnsResolver);
 }
 
+// camelCase needed for compatibility to grpc-native-js package
+// eslint-disable-next-line @typescript-eslint/class-name-casing
```
murgatroid99, line 335: This should actually be changed to `DnsUrl`. It's an internal type so that won't break anything. I just didn't notice when I was first writing the code and I haven't gotten around to fixing it.

packages/grpc-js/.eslintrc
Patrick-Remy: Currently, `any` is sometimes disabled for whole files (https://github.com/grpc/grpc-node/blob/e055468c299a1a22cc36b7760175b1964a3067d7/packages/grpc-js/src/load-balancing-config.ts#L26).
I think you want ignores for every line (would be 54 times)?

Patrick-Remy:
I already signed the CLA via EasyCLA. Now if I click on the link for the form (where previously I signed the document), I am always be redirected to this issue.

packages/grpc-js/.eslintrc
Patrick-Remy: If I change it to something like
```json
        "convertPath": {
          "src/*.ts": ["^src/(.+?)\\.ts$", "build/src/$1.js"]
        }
```
The rule will ignore every import statement (can be tested by adding `import { mockFunction } from '../test/common';` to any `src/*.ts` file). Adding `".ts"` to the extensions makes it work correctly. Same behavior for `node/no-unpublished-require`.

But a path transform is needed, as `const { version: clientVersion } = require('../../package.json');` raises `error    "../../package.json" is not published                        node/no-unpublished-require`, but `import { version as clientVersion } from '../../package.json';` cannot be compiled.

packages/grpc-js/.eslintrc
murgatroid99: No, what I have there is correct. The rule should be disabled for that whole file, but not for the whole package.

Patrick-Remy:
I tried changing the commit email address, to match the one in the CLA, resulting in this..

packages/grpc-js/.eslintrc
Patrick-Remy: Okay, but isn't disabling for the whole file also inconsequent? Is there any criteria, when to disable for the whole file, and when for the line? E.g. in [service-config.ts](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/service-config.ts) are 15 times `disable-next-line no-any`,  in contrast `load-balancing-config.ts` is only 3 times `any` used, but disabled for the whole file.

murgatroid99: In service-config.ts it's also disabled for the whole file, not per line. In both files, all of the functions take an `any` as an input and then validate every field to turn it into a typechecked output. In most other cases, `any` is appropriate to use in some functions in the file, but not in all of them.

packages/grpc-js/.eslintrc
murgatroid99: I don't have strong opinions here, as long as you make it pass. I would even be OK with disabling this check if it doesn't work otherwise. I think it would be unlikely to be a problem.

packages/grpc-js/.eslintrc
Patrick-Remy: Done.

murgatroid99:
I just realized that I missed a question in the original PR comment. I would be happy with completely disabling the `no-unused-vars` rule.

packages/grpc-js/.eslintrc
Patrick-Remy: Realized, that only the flag resolveJsonModule missed for compiling with `import` statement. As the `package.json` will also be copied in the build folder, the path `../../package.json` was fine, but `../package.json` works also like a charm, including typescript support.

Patrick-Remy:
Where would be the place to add the linter to the CI build/test scripts?

PR #1348, opened 2020-04-10
grpc-js: Add more information to proxy errors

murgatroid99:
Maybe this will help some users know **why** grpc failed to connect to the proxy.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/grpc-js",
-  "version": "0.7.8",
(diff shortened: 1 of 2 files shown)


packages/grpc-js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/grpc-js",
-  "version": "0.7.8",
+  "version": "0.7.9",
```
nicolasnoble, line 3: This will conflict with #1346.

murgatroid99: Yeah, I'll fix that one up later.

PR #1355, opened 2020-04-13
Fix a few more issues with the proxy code

murgatroid99:
I manually tested these fixes against the Squid proxy with the configuration from googleapis/nodejs-secret-manager#92. I got to the point where I could see requests being made to the proxy and 503 error responses. I'm not familiar enough with Squid to know if that's a grpc problem or a proxy config problem.

The main changes:

 - Split out `host` and `port` in the `http.request` call.
 - Fix how we stringify IPv6 addresses with ports
 - Call `request.end` after making the request to the proxy.


murgatroid99:
It turns out that the current proxy handling implementation is fundamentally flawed and the changes here will not fix it.

PR #1364, opened 2020-04-16
grpc-js: Use a more structured representation of URIs internally

### LINKED ISSUE #1359: DNS_REGEX Can Hang Node Process
### Problem description
We noticed that it's possible to hang (or seemingly hang) the entire node process with some common (if invalid) hostname/endpoint values.  We discovered this while writing some validation code for our own hostname/endpoint values. This was with the `grpc@grpc-js` project. 

The problem _seems_ to be [this regular expression](https://github.com/grpc/grpc-node/blob/4d1bdc4f20f8f5673e1a5cbd2b64f3d3cd1a6c34/packages/grpc-js/src/resolver-dns.ts#L64)

    const DNS_REGEX = /^(?:dns:)?(?:\/\/(?:[a-zA-Z0-9-]+\.?)+\/)?((?:[a-zA-Z0-9-]+\.?)+)(?::(\d+))?$/;
    
and [its execution here](https://github.com/grpc/grpc-node/blob/4d1bdc4f20f8f5673e1a5cbd2b64f3d3cd1a6c34/packages/grpc-js/src/resolver-dns.ts#L141)

    const dnsMatch = DNS_REGEX.exec(target);

### Reproduction steps
We can reproduce the problem we saw with the behavior of `@grpc/grpc-js` via this small program.  Happy to put together a reproduction with `@grpc/grpc-js` if you'all think that would be helpful.

    const main = () => {
      // host name with an extra port accidentally left on 
      const target = 'foo-internal.aws-us-east-2.tracing.staging-edge.foo-data.net:443:443'
      const DNS_REGEX = /^(?:dns:)?(?:\/\/(?:[a-zA-Z0-9-]+\.?)+\/)?((?:[a-zA-Z0-9-]+\.?)+)(?::(\d+))?$/
      DNS_REGEX.exec(target)
    }
    main()

The hostname `foo-internal.aws-us-east-2.tracing.staging-edge.foo-data.net` seems pathologic, but it's based on a legitimate hostname we've been asked to  use.

Steps:

1. Run the above program with NodeJS 8, 10, or 12

**Expected Behavior**: Program exits normally

**Actual Behavior**: Program hangs for at least 2 minutes, likely longer

### Environment
 - MacOS 10.13/10.14and Ubuntu 18.04
 - Nodes 8, 10, and 12
 - Nodes 10 and 12 installed via nvm, Node 8 installed via apt-get install nodejs
 - `grpc/grpc-js@0.8.1`

### Additional context

Something something NFA backtracking.

murgatroid99:
This change is a result of a few things:

 1. When working on the proxy code, I noticed that there were a few pieces of code that performed overlapping but not identical operations related to parsing targets/addresses.
 2. #1359 was filed, and I realized that the easy fix would make the already unwieldy regexes in the DNS resolver even worse.
 3. I found [the core URI parsing code](https://github.com/grpc/grpc/blob/53657b5de385ffc54e33899b3f2a87ff78d2952b/src/core/lib/uri/uri_parser.cc) and I realized that a structured representation of URIs could solve both problems.

The new `parseUri` function is based on [`grpc_uri_parse`](https://github.com/grpc/grpc/blob/53657b5de385ffc54e33899b3f2a87ff78d2952b/src/core/lib/uri/uri_parser.cc#L188), except that I don't handle query parameters or fragments because they don't seem to be useful anywhere.

The new `splitHostPort` function is based on [`DoSplitHostPort`](https://github.com/grpc/grpc/blob/389d9baa6f5005f73abb42c4f58959736848d525/src/core/lib/gprpp/host_port.cc#L48).

One major thing to note: I realized after implementing this that for a target without a real scheme like `localhost:80`, a completely valid parsing of that target could have `scheme='localhost'` and `path='80'`, but that is not useful. So, if a target has an unknown scheme, the resolver uses the whole target as the path.

This fixes #1359.

### DIFF
--- a/packages/grpc-js/src/channel.ts
+++ b/packages/grpc-js/src/channel.ts
@@ -39,6 +39,7 @@ import { trace, log } from './logging';
 import { SubchannelAddress } from './subchannel';
 import { MaxMessageSizeFilterFactory } from './max-message-size-filter';
 import { mapProxyName } from './http_proxy';
+import { GrpcUri, parseUri, uriToString } from './uri-parser';
 
 export enum ConnectivityState {
   CONNECTING,
@@ -136,8 +137,9 @@ export class ChannelImplementation implements Channel {
   private connectivityStateWatchers: ConnectivityStateWatcher[] = [];
   private defaultAuthority: string;
   private filterStackFactory: FilterStackFactory;
+  private target: GrpcUri;
   constructor(
-    private target: string,
+    target: string,
     private readonly credentials: ChannelCredentials,
     private readonly options: ChannelOptions
   ) {
@@ -164,14 +166,24 @@ export class ChannelImplementation implements Channel {
         );
       }
     }
+    const originalTargetUri = parseUri(target);
+    if (originalTargetUri === null) {
+      throw new Error(`Could not parse target name "${target}"`);
+    }
     if (this.options['grpc.default_authority']) {
       this.defaultAuthority = this.options['grpc.default_authority'] as string;
     } else {
-      this.defaultAuthority = getDefaultAuthority(target);
+      this.defaultAuthority = getDefaultAuthority(originalTargetUri);
     }
-    const proxyMapResult = mapProxyName(target, options);
+    const proxyMapResult = mapProxyName(originalTargetUri, options);
     this.target = proxyMapResult.target;
     this.options = Object.assign({}, this.options, proxyMapResult.extraOptions);
+
+    const targetUri = parseUri(target);
+    if (targetUri === null) {
+      throw new Error(`Could not parse target name "${target}"`);
+    }
+    this.target = targetUri;
     /* The global boolean parameter to getSubchannelPool has the inverse meaning to what
      * the grpc.use_local_subchannel_pool channel option means. */
     this.subchannelPool = getSubchannelPool(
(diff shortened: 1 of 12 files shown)


ThWoywod:
I use this Library behind a HTTP Proxy and figure out that currently it does not work after this Refactoring!

With this commit (238a91c985e47ac70abf01b650797585d89f542e) everything works well.

Its probably an error with the HttpProxy request over here https://github.com/grpc/grpc-node/blob/ae61562be28495d64f6cfe820eea9d26aef7796d/packages/grpc-js/src/http_proxy.ts#L194

The request (with the master branch) contains the following options:
```
{ 
     method: 'CONNECT',
     path: '443',
     host: '2a00:1450:400e:808::200a',
     port: 443 
}
```
The request (before this refactoring (commit: 238a91c985e47ac70abf01b650797585d89f542e)) contains:
```
{ 
     method: 'CONNECT',
     host: '192.168.1.130',  // my http proxy IP
     port: 3128, // my http proxy port
     path: 'dialogflow.googleapis.com:443'
} 
```

murgatroid99:
@ThWoywod Thank you for the information. I think I fixed the problem in #1375.

murgatroid99:
I think I fixed that in version 1.0.1

ThWoywod:
@murgatroid99 Thank you for the fast fix! 👍 
Do you know when version 1.0.1 will be released?

PR #1365, opened 2020-04-16
Bump grpc-js to 1.0 and stop calling it "beta"

murgatroid99:
We're getting there.

### DIFF
--- a/packages/grpc-js/README.md
+++ b/packages/grpc-js/README.md
@@ -1,10 +1,8 @@
 # Pure JavaScript gRPC Client
 
-**Note: This is an beta-level release. Some APIs may not yet be present and there may be bugs. Please report any that you encounter**
-
 ## Installation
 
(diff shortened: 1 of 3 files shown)


packages/grpc-js/README.md
```diff
@@ -1,24 +1,26 @@
...
+ - Automatic reconnection
+ - Servers
+ - Streaming
+ - Metadata
+ - Partial compression support: clients can decompress response messages
+ - Pick first and round robin load balancing policies
+ - Client Interceptors
+ - Connection Keepalives
+ - HTTP Connect support (proxies)
+
```
nicolasnoble, line 23: Is the document https://github.com/grpc/grpc-node/blob/master/PACKAGE-COMPARISON.md also fully up to date?

murgatroid99: It looks like it to me.

PR #1368, opened 2020-04-18
grpc-js: Client waitForReady callback fix.  Fixes #1352

### LINKED ISSUE #1352: grpc-js: waitForReady called multiple times
### Problem description
When using `client.waitForReady`, the ready callback is called multiple times (3 times in this case). Is this is the expected behaviour? I expected it to be called only once. Perhaps I'm misunderstanding the API.

### Reproduction steps
 I'm constructing the gRPC client (which extends from [Client.ts](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/client.ts)).

```ts
import { GradleTasksClient as GrpcClient } from './java-gradle-tasks-grpc/src/main/proto/gradle_tasks_grpc_pb';

const grpcClient = new GrpcClient(
  'localhost:8887',
);
const deadline = new Date();
deadline.setSeconds(deadline.getSeconds() + 1);
grpcClient.waitForReady(deadline, (err: Error | undefined) => {
  console.log('is ready');
});
```

### Environment
 - OS name, version and architecture: macos Catalina 10.15.3 x86_64
 - Node version: v12.16.1
 - Package name and version: `@grpc/grpc-js@0.7.9`

### Screencast

grpc-js-waitforready

badsyntax:
Fixes #1352

### DIFF
--- a/packages/grpc-js/test/test-client.ts
+++ b/packages/grpc-js/test/test-client.ts
@@ -1,0 +1,81 @@
+/*
+ * Copyright 2019 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+// Allow `any` data type for testing runtime type checking.
+// tslint:disable no-any
+import * as assert from 'assert';
+import * as path from 'path';
+
+import * as grpc from '../src';
+import { Server, ServerCredentials } from '../src';
+import { ServiceClient, ServiceClientConstructor } from '../src/make-client';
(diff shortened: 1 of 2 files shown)


badsyntax:
Change the commit messages to follow convention of `grpc-js:`

packages/grpc-js/test/test-client.ts
```diff
@@ -0,0 +1,86 @@
...
+  let server: Server;
+  let client: ServiceClient;
+
+  before(done => {
+    const protoFile = path.join(__dirname, 'fixtures', 'echo_service.proto');
+    const echoService = loadProtoFile(protoFile)
+      .EchoService as ServiceClientConstructor;
+
+    server = new Server();
+    server.addService(echoService.service, {
```
badsyntax, line 44: Does this test rely on a service being added to the server? I don't think client.waitForReady depends on this, see https://github.com/grpc/grpc/blob/master/doc/wait-for-ready.md and https://grpc.github.io/grpc/node/grpc.Client.html#waitForReady

murgatroid99: No, `waitForReady` should not depend on the server having any services. In fact, the whole test file can be simplified significantly by omitting all of the protobuf stuff and directly using `grpc.Client`.

badsyntax: I'll simplify the tests.

badsyntax: I've made that change.

murgatroid99:

PR #1369, opened 2020-04-18
grpc-js: initiate tls connection through http proxy

mrfelton:
Trying to get ssl connection to establish correctly when connecting via proxy.

In reference to https://github.com/grpc/grpc-node/issues/992

### DIFF
--- a/packages/grpc-js/src/http_proxy.ts
+++ b/packages/grpc-js/src/http_proxy.ts
@@ -207,7 +207,6 @@ export function getProxiedConnection(
         );
         var cts = tls.connect({
             ...connectionOptions,
-            host: options.host,
             socket: socket
         }, function () {
           resolve({
--- a/packages/grpc-js/src/http_proxy.ts
+++ b/packages/grpc-js/src/http_proxy.ts
@@ -21,6 +21,7 @@ import { LogVerbosity } from './constants';
 import { parseTarget } from './resolver-dns';
 import { Socket } from 'net';
 import * as http from 'http';
+import * as tls from 'tls'
 import * as logging from './logging';
 import {
   SubchannelAddress,
@@ -202,9 +203,14 @@ export function getProxiedConnection(
             ' through proxy ' +
             proxyAddressString
         );
-        resolve({
-          socket,
-          realTarget,
+        var cts = tls.connect({
+            host: options.host,
+            socket: socket
+        }, function () {
+          resolve({
+            socket: cts,
+            realTarget,
+          });
         });
       } else {
         log(
--- a/packages/grpc-js/src/http_proxy.ts
+++ b/packages/grpc-js/src/http_proxy.ts
@@ -21,6 +21,7 @@ import { LogVerbosity } from './constants';
 import { parseTarget } from './resolver-dns';
 import { Socket } from 'net';
 import * as http from 'http';
+import * as http2 from 'http2';
 import * as tls from 'tls'
 import * as logging from './logging';
 import {
@@ -158,7 +159,8 @@ export interface ProxyConnectionResult {
 
 export function getProxiedConnection(
   address: SubchannelAddress,
-  channelOptions: ChannelOptions
+  channelOptions: ChannelOptions,
+  connectionOptions: http2.SecureClientSessionOptions
 ): Promise<ProxyConnectionResult> {
   if (!('grpc.http_connect_target' in channelOptions)) {
     return Promise.resolve<ProxyConnectionResult>({});
@@ -204,6 +206,7 @@ export function getProxiedConnection(
             proxyAddressString
         );
         var cts = tls.connect({
+            ...connectionOptions,
             host: options.host,
             socket: socket
         }, function () {
--- a/packages/grpc-js/src/http_proxy.ts
+++ b/packages/grpc-js/src/http_proxy.ts
@@ -24,7 +24,6 @@ import { Socket } from 'net';
 import * as http from 'http';
 import * as tls from 'tls';
 import * as logging from './logging';
-import { SecureClientSessionOptions } from 'http2'
 import {
   SubchannelAddress,
   isTcpSubchannelAddress,
@@ -161,7 +160,7 @@ export interface ProxyConnectionResult {
 export function getProxiedConnection(
   address: SubchannelAddress,
   channelOptions: ChannelOptions,
-  connectionOptions: SecureClientSessionOptions
+  connectionOptions: tls.ConnectionOptions
 ): Promise<ProxyConnectionResult> {
   if (!('grpc.http_connect_target' in channelOptions)) {
     return Promise.resolve<ProxyConnectionResult>({});
--- a/packages/grpc-js/src/http_proxy.ts
+++ b/packages/grpc-js/src/http_proxy.ts
@@ -22,9 +22,9 @@ import { getDefaultAuthority } from './resolver';
 import { parseTarget } from './resolver-dns';
 import { Socket } from 'net';
 import * as http from 'http';
-import * as http2 from 'http2';
 import * as tls from 'tls';
 import * as logging from './logging';
+import { SecureClientSessionOptions } from 'http2'
 import {
   SubchannelAddress,
   isTcpSubchannelAddress,
@@ -161,7 +161,7 @@ export interface ProxyConnectionResult {
 export function getProxiedConnection(
   address: SubchannelAddress,
   channelOptions: ChannelOptions,
-  connectionOptions: http2.SecureClientSessionOptions
+  connectionOptions: SecureClientSessionOptions
 ): Promise<ProxyConnectionResult> {
   if (!('grpc.http_connect_target' in channelOptions)) {
     return Promise.resolve<ProxyConnectionResult>({});
@@ -206,9 +206,11 @@ export function getProxiedConnection(
             ' through proxy ' +
             proxyAddressString
         );
-        // The proxy is connecting to a TLS server, so upgrade
-        // this socket connection to a TLS connection.
         if ('secureContext' in connectionOptions) {
+          /* The proxy is connecting to a TLS server, so upgrade this socket
+           * connection to a TLS connection.
+           * This is a workaround for https://github.com/nodejs/node/issues/32922
+           * See https://github.com/grpc/grpc-node/pull/1369 for more info. */
           const cts = tls.connect({
               ...connectionOptions,
               host: getDefaultAuthority(realTarget),
--- a/packages/grpc-js/src/http_proxy.ts
+++ b/packages/grpc-js/src/http_proxy.ts
@@ -18,11 +18,12 @@
 import { URL } from 'url';
 import { log } from './logging';
 import { LogVerbosity } from './constants';
+import { getDefaultAuthority } from './resolver';
 import { parseTarget } from './resolver-dns';
 import { Socket } from 'net';
 import * as http from 'http';
 import * as http2 from 'http2';
-import * as tls from 'tls'
+import * as tls from 'tls';
 import * as logging from './logging';
 import {
   SubchannelAddress,
@@ -205,15 +206,23 @@ export function getProxiedConnection(
             ' through proxy ' +
             proxyAddressString
         );
-        var cts = tls.connect({
-            ...connectionOptions,
-            socket: socket
-        }, function () {
+        // The proxy is connecting to a TLS server, so upgrade
+        // this socket connection to a TLS connection.
+        if ('secureContext' in connectionOptions) {
+          const cts = tls.connect({
+              ...connectionOptions,
+              host: getDefaultAuthority(realTarget),
+              socket: socket,
+            }, () => {
+              resolve({ socket: cts, realTarget });
+            }
+          );
+        } else {
           resolve({
-            socket: cts,
+            socket,
             realTarget,
           });
-        });
+        }
       } else {
         log(
           LogVerbosity.ERROR,
--- a/packages/grpc-js/src/subchannel.ts
+++ b/packages/grpc-js/src/subchannel.ts
@@ -424,6 +424,7 @@ export class Subchannel {
       this.credentials._getConnectionOptions() || {};
 
     if ('secureContext' in connectionOptions) {
+      connectionOptions.ALPNProtocols = ['h2'];
       // If provided, the value of grpc.ssl_target_name_override should be used
       // to override the target hostname when checking server identity.
       // This option is used for testing only.
--- a/packages/grpc-js/src/subchannel.ts
+++ b/packages/grpc-js/src/subchannel.ts
@@ -312,14 +312,14 @@ export class Subchannel {
       connectionOptions.createConnection = (authority, option) => {
         if (proxyConnectionResult.socket) {
           return proxyConnectionResult.socket;
+        } else {
+          /* net.NetConnectOpts is declared in a way that is more restrictive
+           * than what net.connect will actually accept, so we use the type
+           * assertion to work around that. */
+          return net.connect(this.subchannelAddress);
         }
-        /* net.NetConnectOpts is declared in a way that is more restrictive
-         * than what net.connect will actually accept, so we use the type
-         * assertion to work around that. */
-        return net.connect(this.subchannelAddress);
       };
     }
-
 
     connectionOptions = Object.assign(
       connectionOptions,
@@ -416,6 +416,10 @@ export class Subchannel {
   }
 
   private startConnectingInternal() {
+    /* Pass connection options through to the proxy so that it's able to
+     * upgrade it's connection to support tls if needed.
+     * This is a workaround for https://github.com/nodejs/node/issues/32922
+     * See https://github.com/grpc/grpc-node/pull/1369 for more info. */
     const connectionOptions: http2.SecureClientSessionOptions =
       this.credentials._getConnectionOptions() || {};
 
--- a/packages/grpc-js/src/subchannel.ts
+++ b/packages/grpc-js/src/subchannel.ts
@@ -28,7 +28,7 @@ import * as logging from './logging';
 import { LogVerbosity } from './constants';
 import { getProxiedConnection, ProxyConnectionResult } from './http_proxy';
 import * as net from 'net';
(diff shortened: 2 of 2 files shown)


packages/grpc-js/src/http_proxy.ts
```diff
@@ -202,9 +205,15 @@ export function getProxiedConnection(
             ' through proxy ' +
             proxyAddressString
         );
-        resolve({
-          socket,
-          realTarget,
+        var cts = tls.connect({
+            ...connectionOptions,
+            host: options.host,
```
murgatroid99, line 210: `options.host` here is the IP address of the proxy, and when `socket` is set, `host` is only used to validate the server certificates. Why would you want to validate the server certificates against the proxy's IP address?

mrfelton: Good question. Also works without setting `host` - have pushed update that removes.

packages/grpc-js/src/subchannel.ts
```diff
@@ -302,21 +302,20 @@ export class Subchannel {
       if (proxyConnectionResult.socket) {
         connectionOptions.socket = proxyConnectionResult.socket;
```
murgatroid99, line 303: With your changes, this line should do nothing. In fact, this whole `if ('secureContext' in connectionOptions)` section should do nothing, because you are unconditionally passing a `createConnection` option, which takes priority over all of these.

packages/grpc-js/src/subchannel.ts
```diff
@@ -302,21 +302,20 @@ export class Subchannel {
...
+     * the options when establishing plaintext connections, so we need to
+     * establish that connection explicitly. */
+    connectionOptions.createConnection = (authority, option) => {
+      if (proxyConnectionResult.socket) {
+        return proxyConnectionResult.socket;
+      } else {
+        /* net.NetConnectOpts is declared in a way that is more restrictive
+         * than what net.connect will actually accept, so we use the type
+         * assertion to work around that. */
+        return net.connect(this.subchannelAddress);
```
murgatroid99, line 316: Doing this unconditionally breaks the unproxied TLS case, because you are forcing the use of a plaintext connection here.

packages/grpc-js/src/http_proxy.ts
murgatroid99: Actually, I think you want `host: getDefaultAuthority(realTarget)` here, so that you don't have to pass the `grpc.ssl_target_name_override` option when constructing the client. In fact, TLS certificate validation is the whole reason we pass the default authority to `http2.connect` in the first place.

packages/grpc-js/src/subchannel.ts
murgatroid99: Let's not worry about this for the moment.

mrfelton: Have updated this so that we still have the plaintext conditional

packages/grpc-js/src/http_proxy.ts
mrfelton: updated

mrfelton:
Do I need to do something special to get all of the tests to pass? I get the following results:

```
  123 passing (587ms)
  2 pending
  27 failing
```

This is the same result I get in master branch.

packages/grpc-js/src/subchannel.ts
murgatroid99: You have the exact same `createConnection` function in the secure and insecure case. The problem was, and is, that you're always doing `net.connect`, which creates a plain TCP connection. In the secure case, you should only be using the `createConnection` to pass in the `socket` from the proxy.

murgatroid99:
You should undo that last commit. Using the target name override there is correct, that option just shouldn't be necessary in most cases. In general, when using the `grpc-js` API, you should not be passing the `grpc.ssl_target_name_override` option at all; it is mainly good for testing.

murgatroid99:
I've been making these small corrections, but overall I still don't understand how this change could cause any difference in behavior from what the existing code does. It looks to me like it is simply pulling out part of the behavior of `http2.connect` into the proxy handling code.

mrfelton:
To be honest, I as hoping you'd be able to explain it! Somehow this forces the tls connection to the real target to be established.

mrfelton:
I've pushed another update to this. I think this resolves the issues in your previous comments.

My basic understanding is that when you initially set up the connection to the proxy it doesn't support tls, but if the proxy is connecting to a TLS server the socket needs upgrading to a TLS connection.

Does that makes sense fro your perspective?

murgatroid99:
The thing is that `http2.connect` already calls `tls.connect` internally with the options passed to it if it's connecting to an `https` address. So the original code that passes the `socket` from the proxy in the `http2.connect` options should be doing the exact same thing that this code is doing.

mrfelton:
I know in theory that is supposed to be the case, but in practice it doesn't. Or something about the way we are calling `http2.connect` is preventing that from working as it should.

packages/grpc-js/src/subchannel.ts
```diff
@@ -299,24 +300,22 @@ export class Subchannel {
...
+    }
+    /* In all but the most recent versions of Node, http2.connect does not use
+     * the options when establishing plaintext connections, so we need to
+     * establish that connection explicitly. */
+    connectionOptions.createConnection = (authority, option) => {
       if (proxyConnectionResult.socket) {
-        connectionOptions.socket = proxyConnectionResult.socket;
+        return proxyConnectionResult.socket;
+      } else if ('secureContext' in connectionOptions) {
+        return tls.connect(this.subchannelAddress);
```
murgatroid99, line 311: This isn't better because you're not passing in the options. When connecting with TLS and without a proxy, it should simply not be passing a `createConnection` option at all.

murgatroid99:
I'm not talking about "in theory", I'm talking about [the `http2.connect` source code](https://github.com/nodejs/node/blob/v10.x/lib/internal/http2/core.js#L2937)

packages/grpc-js/src/subchannel.ts
mrfelton: My typescript is not up to par :( What I really wanted to do in the ssl lock was like this:

```
      if (proxyConnectionResult.socket) {
        connectionOptions.createConnection = (authority, option) => {
          return proxyConnectionResult.socket;
        };
      }
```

But I get typescript errors.

mrfelton:
> I'm not talking about "in theory", I'm talking about [the `http2.connect` source code](https://github.com/nodejs/node/blob/v10.x/lib/internal/http2/core.js#L2937)

Understood - but in reality it simply doesn't work the way it was.

murgatroid99:
So, one difference between what you're doing here and what `http.connect` is doing is that `initializeTLSOptions` function call before passing the options to `tls.connect`. What if you copy that function (and the `initializeOptions` function that it calls) and call it the same way in the proxy code? Does your request through TOR still work?

mrfelton:
Yes, the tor connection works correctly even after wrapping my `tls.connect` options in `initializeTLSOptions`

packages/grpc-js/src/subchannel.ts
murgatroid99: Ah, the issue is that TypeScript thinks that `proxyConnectionResult.socket` might become `null` between when you check it and when the callback is called. This should work:

```ts
if (proxyConnectionResult.socket) {
  connectionOptions.createConnection = (authority, option) => {
    return proxyConnectionResult.socket!; //Exclamation mark here
  };
}
```

mrfelton:
Actually, maybe I spoke out of turn there.

If I call like this it does not work

```
        if ('secureContext' in connectionOptions) {
          socket = tls.connect(Number(options.port), getDefaultAuthority(realTarget), initializeTLSOptions(options, getDefaultAuthority(realTarget)), () => {
            resolve({ socket, realTarget });
          });
        } else {
          resolve({
            socket,
            realTarget,
          });
        }
```

mrfelton:
Correct, this works:

```
          const cts = tls.connect(Number(options.port), getDefaultAuthority(realTarget), initializeTLSOptions({ ...connectionOptions, socket}, getDefaultAuthority(realTarget)), () => {
            resolve({ socket: cts, realTarget });
          });
```

So wrapping the options in `initializeTLSOptions` doesn't seem to make a difference here.

packages/grpc-js/src/subchannel.ts
mrfelton: Nice one! updated.

murgatroid99:
I have another idea: log calls to `tls.connect` by monkey patching it like this:

```js
const originalTlsConnect = tls.connect;
tls.connect = (...args) => {
  console.log(args);
  return originalTlsConnect(...args);
};
```

Then check what the output is for your test case, both with this change (including the version with `initializeTlsOptions`, to minimize confounding factors) and with the original code.

mrfelton:
My version (works):
```js
[
  9065,
  'zapn34qfeedw2l5y26p3hnnkusqnbhxcxw64lq5cojmvq45yw4bc3sqd.onion',
  {
    secureContext: SecureContext { context: SecureContext {} },
    socket: Socket {
      connecting: false,
      _hadError: false,
      _parent: null,
      _host: null,
      _readableState: [ReadableState],
      readable: true,
      _events: [Object: null prototype] {},
      _eventsCount: 0,
      _maxListeners: undefined,
      _writableState: [WritableState],
      writable: true,
      allowHalfOpen: false,
      _sockname: null,
      _pendingData: null,
      _pendingEncoding: '',
      server: null,
      _server: null,
      parser: null,
      _httpMessage: null,
      [Symbol(asyncId)]: 133,
      [Symbol(kHandle)]: [TCP],
      [Symbol(lastWriteQueueSize)]: 0,
      [Symbol(timeout)]: null,
      [Symbol(kBuffer)]: null,
      [Symbol(kBufferCb)]: null,
      [Symbol(kBufferGen)]: null,
      [Symbol(kBytesRead)]: 0,
      [Symbol(kBytesWritten)]: 0
    },
    settings: {},
    Http1IncomingMessage: [Function: IncomingMessage],
    Http1ServerResponse: [Function: ServerResponse],
    Http2ServerRequest: [Function: Http2ServerRequest],
    Http2ServerResponse: [Function: Http2ServerResponse],
    ALPNProtocols: [ 'h2' ],
    servername: 'zapn34qfeedw2l5y26p3hnnkusqnbhxcxw64lq5cojmvq45yw4bc3sqd.onion'
  },
  [Function]
]
```

Old version (doesn't work):
```js
[
  '443',
  'zapn34qfeedw2l5y26p3hnnkusqnbhxcxw64lq5cojmvq45yw4bc3sqd.onion',
  {
    secureContext: SecureContext { context: SecureContext {} },
    socket: Socket {
      connecting: false,
      _hadError: false,
      _parent: null,
      _host: null,
      _readableState: [ReadableState],
      readable: true,
      _events: [Object: null prototype] {},
      _eventsCount: 0,
      _maxListeners: undefined,
      _writableState: [WritableState],
      writable: true,
      allowHalfOpen: false,
      _sockname: null,
      _pendingData: null,
      _pendingEncoding: '',
      server: null,
      _server: null,
      parser: null,
      _httpMessage: null,
      [Symbol(asyncId)]: 135,
      [Symbol(kHandle)]: [TCP],
      [Symbol(lastWriteQueueSize)]: 0,
      [Symbol(timeout)]: null,
      [Symbol(kBuffer)]: null,
      [Symbol(kBufferCb)]: null,
      [Symbol(kBufferGen)]: null,
      [Symbol(kBytesRead)]: 0,
      [Symbol(kBytesWritten)]: 0
    },
    host: '127.0.0.1',
    port: 9065,
    settings: {},
    Http1IncomingMessage: [Function: IncomingMessage],
    Http1ServerResponse: [Function: ServerResponse],
    Http2ServerRequest: [Function: Http2ServerRequest],
    Http2ServerResponse: [Function: Http2ServerResponse],
    ALPNProtocols: [ 'h2' ],
    servername: 'zapn34qfeedw2l5y26p3hnnkusqnbhxcxw64lq5cojmvq45yw4bc3sqd.onion'
  }
]
```

Different host and port

murgatroid99:
OK, it should be pretty simple to change that: in the unmodified code, just put [this line](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/subchannel.ts#L320-L322) in a `if (!proxyConnectionResult.socket)` block. Does that make your code work?

mrfelton:
Unfortunately not. I have also tried other things to make the `tls.connect` from within `http2.connect` be called in exactly the same way, and even when the args are the same as in my version it still doesn't work.

murgatroid99:
Just to verify in the other direction, if you modify your version of the code to pass the same arguments that the old code passes, your version still works?

mrfelton:
Yes, it does.

i.e. I can change to this and it still works:

```
const cts = tls.connect(443, getDefaultAuthority(realTarget), initializeTLSOptions({ ...connectionOptions, socket, host: '127.0.0.1', port: 9065}, getDefaultAuthority(realTarget)), () => {
  resolve({ socket: cts, realTarget });
});
```

Which gives me call opts that look like those from the original code:

```js
[
  443,
  'zapn34qfeedw2l5y26p3hnnkusqnbhxcxw64lq5cojmvq45yw4bc3sqd.onion',
  {
    secureContext: SecureContext { context: SecureContext {} },
    socket: Socket {
      connecting: false,
      _hadError: false,
      _parent: null,
      _host: null,
      _readableState: [ReadableState],
      readable: true,
      _events: [Object: null prototype] {},
      _eventsCount: 0,
      _maxListeners: undefined,
      _writableState: [WritableState],
      writable: true,
      allowHalfOpen: false,
      _sockname: null,
      _pendingData: null,
      _pendingEncoding: '',
      server: null,
      _server: null,
      parser: null,
      _httpMessage: null,
      [Symbol(asyncId)]: 134,
      [Symbol(kHandle)]: [TCP],
      [Symbol(lastWriteQueueSize)]: 0,
      [Symbol(timeout)]: null,
      [Symbol(kBuffer)]: null,
      [Symbol(kBufferCb)]: null,
      [Symbol(kBufferGen)]: null,
      [Symbol(kBytesRead)]: 0,
      [Symbol(kBytesWritten)]: 0
    },
    host: '127.0.0.1',
    port: 9065,
    settings: {},
    Http1IncomingMessage: [Function: IncomingMessage],
    Http1ServerResponse: [Function: ServerResponse],
    Http2ServerRequest: [Function: Http2ServerRequest],
    Http2ServerResponse: [Function: Http2ServerResponse],
    ALPNProtocols: [ 'h2' ],
    servername: 'zapn34qfeedw2l5y26p3hnnkusqnbhxcxw64lq5cojmvq45yw4bc3sqd.onion'
  },
  [Function]
]
```

There is a different though, which is the callback function as the 3rd arg.

It seems to me that in my version when calling `tls.connect` from within `request.once('connect'` in the proxy module it waits for the socket to be established before resolving it from `getProxiedConnection`. Which in turn means that the socket that gets passed to `http2.connect` is already fully established and tls enabled.

In http2 core they don't call `tls.connect` with a callback.

They will however call a `listener` that you can pass to `http2.connect` as a third argument, which to me looks like it would call once the socket has been upgraded for tls. See here:

https://github.com/nodejs/node/blob/ef2df6986b6a3c34c6a0a97c120d5f8d05c4a1ce/lib/internal/http2/core.js#L2952-L2953

However we don't pass in a `listener third arg.

I may be barking up the wrong tree, but it seems that our socket is just not ready to use with tls before we start trying to use it.

murgatroid99:
It should be pretty easy to check if passing a callback argument to `http2.connect` is the issue. The `session.once('connect')` callback from a few lines down could be passed directly as the third argument to `http2.connect`.

In the other direction, what if in this PR you have the proxy code return the TLS socket immediately, instead of passing a callback?

murgatroid99:
I think I may have a smoking gun: the `Http2Session` constructor waits for a `TLSSocket`'s `secureConnect` event if `socket.connecting` is true, but continues immediately otherwise. (https://github.com/nodejs/node/blob/ef2df6986b6a3c34c6a0a97c120d5f8d05c4a1ce/lib/internal/http2/core.js#L1084-L1090). However, if the `TLSSocket` constructor is passed a socket, then it only sets its own `connecting` value to true if the underlying socket's `connecting` is also true, but it emits `connect`, not `secureConnect` at that time. (https://github.com/nodejs/node/blob/master/lib/_tls_wrap.js#L791-L800). This looks like a mismatch of expectations: `Http2Session` is expecting that `connecting` means "has not yet emitted the `secureConnect` event", but in this situation `TLSSocket` is using it to mean "has not yet emitted the `connect` event".

If this is true, passing a callback to `http.connect` will not fix the problem, and returning the result of `tls.connect` without waiting for it to finish will make the code in this PR behave like the broken version.

murgatroid99:
See nodejs/node#32922 for a demonstration of the problem I described independent of both `grpc` and HTTP CONNECT.

mrfelton:
>It should be pretty easy to check if passing a callback argument to http2.connect is the issue. The session.once('connect') callback from a few lines down could be passed directly as the third argument to http2.connect.

Does't work this way.

> In the other direction, what if in this PR you have the proxy code return the TLS socket immediately, instead of passing a callback?

Also does't work this way.

> See nodejs/node#32922 for a demonstration of the problem I described independent of both grpc and HTTP CONNECT.

Wow ok, so this really does look like a bug in node itself. With that in mind, how do you feel about the workaround proposed in this PR?

mrfelton:
Thanks. Have made those updates.

mrfelton:
Thanks for your help - solving this and making it possible to do grpc over tor is a big deal. 👍

mrfelton:
Unfortunately this no longer works after merging :(

mrfelton:
Seems the new uri parsing stuff is not properly parsing the proxy uri. For example, with
  
`grpc_proxy='http//127.0.0.1:9065'`

It seems to get parsed as:

`{ scheme: 'http', authority: undefined, path: '//127.0.0.1:9065' }`

This in turns prevents a connection to the proxy.

murgatroid99:
I think I fixed the conflict in #1375.

mrfelton:
progress, but still not working unfortunately. Now getting `Received RST_STREAM with code 2` error and the tls session again doesn't look to being getting established correctly.

PR #1370, opened 2020-04-18
Bump protobuf submodule dep of grpc-tools

rhinodavid:
This dependency hasn't been updated for almost two years and the jspbs
generated are out of sync with the docs. Specifically, none of the
setters return the message so chaining isn't possible.


murgatroid99:
Sorry for the delay. Did you bump the submodule to a specific release tag and if so, which one?

rhinodavid:
> Sorry for the delay. Did you bump the submodule to a specific release tag and if so, which one?

I'm trying to think back and I'm not sure; it's been a bit. It might have been to the top of master.

I'd like to go back and bump it to the latest stable, [3.11.4](https://github.com/protocolbuffers/protobuf/releases/tag/v3.11.4). Could we merge that?

murgatroid99:
The latest stable release sounds to me like a good point to bump to.

rhinodavid:
Awesome. Should be good to go.

PR #1375, opened 2020-04-21
grpc-js: Fix interactions between proxy code and new URI parsing

murgatroid99:
There were multiple issues here:

 - We expect a proxy URL in the form `http://host:port`, but this does not actually fit with our naming scheme, so the URI parser parses it wrong. The solution is to revert to a previous version of `getProxyInfo` that parses the proxy address using the `URL` class.

 - If the target had no scheme and was parsed weirdly into the `GrpcUri` structure, the resolver knew how to deal with that, but the proxy didn't, and it was making a bad request to the proxy. The solution is to stop using a default resolver, and instead modify the target at channel construction time so that it has a known scheme.

 - The channel constructor was just overwriting the target after it got modified by the proxy.

 - The proxy map result target didn't have an explicit scheme, even though it would definitely need to be interpreted as a DNS name. Added a scheme to fix that.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/grpc-js",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "description": "gRPC Library for Node - pure JS implementation",
   "homepage": "https://grpc.io/",
   "repository": "https://github.com/grpc/grpc-node/tree/master/packages/grpc-js",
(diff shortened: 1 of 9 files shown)


ThWoywod:
It looks like like, that this fix do not really work.

I get the following error message:
image

murgatroid99:
@ThWoywod Can you run your code with the environment variables `GRPC_TRACE=all` and `GRPC_VERBOSITY=DEBUG`, and share the log output?

murgatroid99:
Never mind on the trace information. I believe I solved the problem in #1381. This time I managed to test it and I did successfully traverse a proxy and get a response from the server.

ThWoywod:
Yeah thanks, with v 1.0.2 it works pretty well

PR #1379, opened 2020-04-22
grpc-js: do not prematurely resolve proxy socket

mrfelton:
This resolves the last remaining issue with tls proxy support mentioned here https://github.com/grpc/grpc-node/pull/1369#issuecomment-617946074

Probably the result of a merge conflict but the issue is that the proxy socket is being unconditionally resolved prematurely, essentially bypassing the creation of the tls enabled socket.


murgatroid99:
I have a more complete fix, including this change, in #1381, so I think it's best to just go with that.

Issue #528, opened 2018-09-06
Protobufjs' codegen support


MOZGIII:
# Scratch the original issue, this is now about improving TypeScript definitions in general.
Skip ahead to https://github.com/grpc/grpc-node/issues/528#issuecomment-419288616

---
### Is your feature request related to a problem? Please describe.
I can't currently pass metadata (or any per-call option) when using protobuf.js.

### Describe the solution you'd like
Add a way to pass metadata at the very least, and potentially any kind of per-call option when using protobuf.js.
This seems to be an integration issue between the two projects, and I don't know where to create an issue for it. I chose this project because it own the concept of metadata.

Technically, I'm proposing, in TypeScript/Flow terms, to change the API and make the client generic to allow customization of the "metadata/arbitrary options" type, and add an argument to every call:

```typescript
// before
class MyService extends $protobuf.rpc.Service {
  constructor(rpcImpl: $protobuf.RPCImpl, requestDelimited?: boolean, responseDelimited?: boolean);
  public static create(rpcImpl: $protobuf.RPCImpl, requestDelimited?: boolean, responseDelimited?: boolean): MyService;
  public sampleCall(request: pkg.ISampleCallRequestType, callback: pkg.MyService.SampleCallCallback): void;
  public sampleCall(request: pkg.ISampleCallRequestType): Promise<pkg.ISampleCallRsponseType>;
}

// after
class MyService<T> extends $protobuf.rpc.Service {
  constructor(rpcImpl: $protobuf.RPCImpl<T>, requestDelimited?: boolean, responseDelimited?: boolean);
  public static create<T>(rpcImpl: $protobuf.RPCImpl<T>, requestDelimited?: boolean, responseDelimited?: boolean): MyService<T>;
  public sampleCall(request: pkg.ISampleCallRequestType, opts: T, callback: pkg.MyService.SampleCallCallback): void;
  public sampleCall(request: pkg.ISampleCallRequestType, opts: T): Promise<pkg.ISampleCallRsponseType>;
}
```

The `rpcImpl` should have the `opts` argument passed along with everything else. For the purposes of passing headers, we can start with supporting `T` with type like `{ headers: Metadata }` on the `node-grpc` end, and injecting headers (if any specified) into unary/streaming call.

### Describe alternatives you've considered
[...]

PR #1380, opened 2020-04-22
grpc-tools: add grpc_js param to grpc_out

badsyntax:
This is a minor change to use the pure JavaScript gRPC Client `@grpc/grpc-js` instead of the (now deprecated) `grpc` node package when generating javascript files using the proto compiler. This change is very similar to the `generate_package_definition` param.

Reasoning: 

I understand that it's recommended to use `@grpc/proto-loader` to the load protos files. 

I have some hesitation to adopt this approach, as when using TypeScript, I have to typecast using [type assertions](https://www.typescriptlang.org/docs/handbook/basic-types.html#type-assertions) to type the grpc client as a [`ServiceClientConstructor`](https://github.com/grpc/grpc-node/blob/4ec023c271890a0f8ecb2548ba2001357ae31c79/packages/grpc-js/src/make-client.ts#L87-L94), for example:

```ts
const echoService = loadProtoFile(protoFile)
      .EchoService as ServiceClientConstructor;
```

My client is then defined as:

https://github.com/grpc/grpc-node/blob/4ec023c271890a0f8ecb2548ba2001357ae31c79/packages/grpc-js/src/make-client.ts#L83-L85

..and I loose all the benefits of (for example) IntelliSense in my IDE. I also don't know if my method calls are correct until I compile the TypeScript to JS. It's not a great developer experience.

The change i'm proposing is to support "grpc_js" within the proto compiler. This will allow me to generate static javascript files that require `@grpc/grpc-js` instead of `grpc`, and will allow existing tools to then generate types from those files. 

I also feel this a good change to support existing projects & tooling that would like to switch over to using `@grpc/grpc-js`. It will allow for better adoption as the required changes in the consuming projects are minimal.

### DIFF
--- a/packages/grpc-tools/src/node_generator.cc
+++ b/packages/grpc-tools/src/node_generator.cc
@@ -212,7 +212,8 @@ void PrintService(const ServiceDescriptor* service, Printer* out,
 void PrintImports(const FileDescriptor* file, Printer* out,
                   const Parameters& params) {
   if (!params.generate_package_definition) {
-    out->Print("var grpc = require('grpc');\n");
+    grpc::string package = params.grpc_js ? "@grpc/grpc-js" : "grpc";
+    out->Print("var grpc = require('$package$');\n", "package", package);
   }
   if (file->message_type_count() > 0) {
     grpc::string file_path =
--- a/packages/grpc-tools/src/node_generator.h
+++ b/packages/grpc-tools/src/node_generator.h
@@ -26,6 +26,8 @@ namespace grpc_node_generator {
 struct Parameters {
   // Generate a package definition object instead of Client classes
   bool generate_package_definition;
+  // Use pure JavaScript gRPC Client
+  bool grpc_js;
 };
 
 grpc::string GenerateFile(const grpc::protobuf::FileDescriptor* file,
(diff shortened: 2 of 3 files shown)


laike9m:
Hi Richard, you mentioned
> This will allow me to generate static javascript files that require @grpc/grpc-js instead of grpc, and will allow existing tools to then generate types from those files.

May I ask what tools you're using to generate TS types? I tried `grpc_tools_node_protoc_ts` but it does not seem to support grpc-js. Thanks.

laike9m:
Never mind. I saw your PR: https://github.com/improbable-eng/ts-protoc-gen/pull/236

Hope it gets merged soon.

badsyntax:
@laike9m i'm not sure if the authors of that package are keen to improve it as i've had zero feedback.

i've been meaning to have a look at the new typescript generator offered by @murgatroid99, and i would suggest to consider using that package instead.

see: https://github.com/grpc/grpc-node/issues/528#issuecomment-649819404

example of how it can be used: https://github.com/grpc/grpc-node/blob/095a3b47695e37297a8f9714120ec1d91e24811d/packages/grpc-js/generateTypes.sh#L18-L20

laike9m:
Great, let me try it out. Thanks for the pointer.

laike9m:
> @laike9m i'm not sure if the authors of that package are keen to improve it as i've had zero feedback.
> 
> i've been meaning to have a look at the new typescript generator offered by @murgatroid99, and i would suggest to consider using that package instead.
> 
> see: [#528 (comment)](https://github.com/grpc/grpc-node/issues/528#issuecomment-649819404)
> 
> example of how it can be used:
> 
> https://github.com/grpc/grpc-node/blob/095a3b47695e37297a8f9714120ec1d91e24811d/packages/grpc-js/generateTypes.sh#L18-L20

I gave it a try, but somehow it confused me. Not sure if I'm doing anything wrong:
https://github.com/grpc/grpc-node/issues/528#issuecomment-653441548

PR #1382, opened 2020-04-22
fix: make stream.write() synchronous in server-call

alexander-fenster:
`stream.write()` in server stream code is asynchronous for no apparent reason. As a result, it breaks the perfectly reasonable RPC implementation code like this one:

```js
stream.write(data);
stream.write(data);
stream.emit('error', err);
```

since `write` being `async` for no reason delays its execution and an error is emitted first within the same event loop.

Note that `grpc` server implementation works properly in this case.

We had this problem in [nodejs-spanner](https://github.com/googleapis/nodejs-spanner/blob/master/test/mockserver/mockspanner.ts#L499-L503) test code where they are using gRPC server to emulate some conditions for testing the client, and it blocks migrating this library from `grpc` to `@grpc/grpc-js`.

### DIFF
--- a/packages/grpc-js/src/server-call.ts
+++ b/packages/grpc-js/src/server-call.ts
@@ -172,14 +172,14 @@ export class ServerWritableStreamImpl<RequestType, ResponseType>
     this.call.sendMetadata(responseMetadata);
   }
 
-  async _write(
+  _write(
     chunk: ResponseType,
     encoding: string,
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     callback: (...args: any[]) => void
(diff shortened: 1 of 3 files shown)


alexander-fenster:
Also, unrelated note but still - I must say that this problem wouldn't have happened if the sources were clean from the linter point of view.

image

Oh, how wrong VSCode is here. It has a _huge_ effect :) 

Having said that - are you willing to accept contributions related to linting and possibly types? The code, when opened in VSCode, had a little bit more red fragments than it was comfortable for me. (cc @JustinBeckwith just FYI)

Issue #1389, opened 2020-04-23
Export sendUnaryData type from main export


mdouglass:
### Is your feature request related to a problem? Please describe.
The sendUnaryData type was previously exported by the native gRPC library and we were using it when implementing an UntypedServiceImplementation (it is the second parameter on those functions).

### Describe the solution you'd like
Export sendUnaryData from index.ts

PR #1390, opened 2020-04-23
Export sendUnaryData as part of the public API #1389

mdouglass:


murgatroid99:
I think you need another line to export the type.

mdouglass:
> I think you need another line to export the type.

Yep, that's what I get for being lazy and using the GH edit interface rather than actually pulling my own fork. Fixed.

PR #1398, opened 2020-04-27
grpc-js: Consistently close underlying stream on server call error

### LINKED ISSUE #1396: grpc-js: end event emitted on client streaming errors
### Problem description
If an uncaught error occurs in the data handler the end handler is still fired. I'm not sure if this is intentional but based on the standard node.js readable stream I would expect this not to be called. 

In this scenario because the callback is already handled [here](https://github.com/grpc/grpc-node/blob/%40grpc/grpc-js%401.0.3/packages/grpc-js/src/server.ts#L644) you cannot subscribe to the error event to handle the error and send a response.
```
clientStreamExample(call, callback) {
  call.on('data', () => {
    throw new Error('my error');
  });

  call.on('end', () => {
    console.log('ended');
    callback(null, {});
  });
}
```

### Environment
 - OS name, version and architecture: macOS 10.15.4
 - Node version: v12.16.2
 - Package name and version @grpc/grpc-js@1.0.3

murgatroid99:
Some other formatting stuff ended up in here too. The main changes are to `setupSurfaceCall` and its call sites. The goal here is to make sure the internal `Http2ServerCallStream` stops processing whenever the surface call object encounters an error.

I think this should fix #1396.

### DIFF
--- a/packages/grpc-js/src/server-call.ts
+++ b/packages/grpc-js/src/server-call.ts
@@ -20,7 +20,11 @@ import * as http2 from 'http2';
 import { Duplex, Readable, Writable } from 'stream';
 
 import { StatusObject } from './call-stream';
-import { Status, DEFAULT_MAX_SEND_MESSAGE_LENGTH, DEFAULT_MAX_RECEIVE_MESSAGE_LENGTH } from './constants';
+import {
+  Status,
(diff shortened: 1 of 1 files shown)


murgatroid99:
I don't think this is going in the right direction.

PR #1409, opened 2020-05-06
Add a GitHub workflow for building the grpc binary

murgatroid99:
Let's see how well this works just using the existing kokoro build scripts.


.github/workflows/build.yml
```diff
@@ -0,0 +1,134 @@
...
+  pull_request:
+    branches:
+      - 'grpc@1.24.x'
+
+jobs:
+  all_node_glibc:
+    name: All Simple Native Builds
+    strategy:
+      matrix:
+        node: ['4', '5', '6', '7', '8', '9', '10', '11', '12', '13']
```
nicolasnoble, line 16: Isn't 14 available already?

.github/workflows/build.yml
```diff
@@ -0,0 +1,134 @@
...
+      if: ${{matrix.os == 'windows-latest'}}
+    - uses: actions/upload-artifact@v2
+      with:
+        name: node_glibc_${{matrix.os}}_${{matrix.node}}_${{matrix.arch}}
+        path: artifacts/
+  all_electron:
+    name: All Electron Builds
+    strategy:
+      matrix:
+        electron: ['1.0', '1.1', '1.2', '1.3', '1.4', '1.5', '1.6', '1.7', '1.8', '2.0', '3.0', '3.1', '4.1', '4.2', '5.0', '6.0', '7.0']
```
nicolasnoble, line 49: Electron 8 is out too, isn't it?

.github/workflows/build.yml
murgatroid99: I wanted to have this PR be about getting the builds we currently have on master working on this branch with this workflow, and then another PR making updates, including building for new versions.

PR #1410, opened 2020-05-07
Update protobuf_js_6_common.js

kally788:


murgatroid99:
Thank you for your contribution, but we are no longer making or accepting significant changes to this library. I suggest instead using the `@grpc/proto-loader` library.

Issue #1412, opened 2020-05-08
grpc-js: string in metadata be split with comma


LeonsChun:
### Problem description
Set a string "1,2" in metedata and put in request. In server, the string in metadata is not "1,2" but ["1", "2"].

### Reproduction steps
![image](https://user-images.githubusercontent.com/3643627/81390107-7f271b00-914d-11ea-8f2a-604fbc6f42fd.png)

murgatroid99:
This is intentional. The [gRPC protocol definition](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) says this:

> Custom-Metadata header order is not guaranteed to be preserved except for values with duplicate header names. **Duplicate header names may have their values joined with "," as the delimiter and be considered semantically equivalent.** Implementations must split Binary-Headers on "," before decoding the Base64-encoded values.

LeonsChun:
The string is not splitted in grpc-node. It's a bug?

![image](https://user-images.githubusercontent.com/3643627/81431278-7bb68280-9193-11ea-97d9-999ea3d45cb6.png)

murgatroid99:
It turns out that I misunderstood the spec. That is a bug in grpc-js. There's a fix pending in #1414, and I can probably have that out next week.

lwmxiaobei:
> It turns out that I misunderstood the spec. That is a bug in grpc-js. There's a fix pending in #1414, and I can probably have that out next week.

hi, Have you fixed this ?

murgatroid99:
I published the fix for this in grpc-js version 1.0.4.

PR #1414, opened 2020-05-08
grpc-js: Don't split header values by commas

murgatroid99:
According to #1412 the current grpc-js behavior is different from the other implementation's behavior, so we should change it.

### DIFF
--- a/packages/grpc-js/src/metadata.ts
+++ b/packages/grpc-js/src/metadata.ts
@@ -278,11 +278,7 @@ export class Metadata {
               result.add(key, value);
             });
           } else if (values !== undefined) {
-            if (isCustomMetadata(key)) {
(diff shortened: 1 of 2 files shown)


packages/grpc-js/src/metadata.ts
```diff
@@ -278,11 +278,7 @@ export class Metadata {
               result.add(key, value);
             });
           } else if (values !== undefined) {
-            if (isCustomMetadata(key)) {
-              values.split(',').forEach((v) => result.add(key, v.trim()));
-            } else {
-              result.add(key, values);
-            }
+            result.add(key, values);
```
nicolasnoble, line 281: Maybe we should have a test for this eventually?

murgatroid99:
Added a test.

Issue #1407, opened 2020-05-06
Node 14 ArrayBuffer changes causes grpc to crash.


strigona-worksight:
### Problem description
Node 14 has some breaking changes regarding ArrayBuffers as noted here https://github.com/nodejs/node/pull/30782 and https://nodejs.org/en/blog/release/v14.0.0/

> Multiple ArrayBuffers pointing to the same base address are no longer allowed by V8. This may impact native addons.

### Reproduction steps
I don't have solid/standalone reproduction steps as of yet, but there is a relevant issue here with the same stacktrace that we received: https://github.com/pulumi/pulumi/issues/4258#issuecomment-620973527

### Environment
 - OS name, version and architecture: Ubuntu 19.10
 - Node version: 14.0.0
 - Node installation method: nvm
 - If applicable, compiler version:
 - Package name and version: gRPC@1.24.2

### Additional context
https://github.com/nodejs/node/issues/32463

murgatroid99:
Thank you for the information. I will try to fix this; any additional information will help.

I suggest switching to `@grpc/grpc-js`. We will be concentrating most of our development efforts on that library going forward.

strigona-worksight:
Thanks for your reply! We'll look into switching over to the grpc-js.

strigona-worksight:
We tried out grpc 1.24.3 with node v14.4.0 and haven't had any issues yet. The release says Node 14 support was added, though I couldn't find any commits tied to that claim.

hugebdu:
Same here
```
#
# Fatal error in , line 0
# Check failed: result.second.
#
#
#
#FailureMessage Object: 0x7ffeefbfe100
 1: 0x100112612 node::NodePlatform::GetStackTracePrinter()::$_3::__invoke() [/Users/daniels/.nvm/versions/node/v14.9.0/bin/node]
 2: 0x10100c7b2 V8_Fatal(char const*, ...) [/Users/daniels/.nvm/versions/node/v14.9.0/bin/node]
 3: 0x1004d37b4 v8::internal::GlobalBackingStoreRegistry::Register(std::__1::shared_ptr<v8::internal::BackingStore>) [/Users/daniels/.nvm/versions/node/v14.9.0/bin/node]
 4: 0x1001f9436 v8::ArrayBuffer::GetBackingStore() [/Users/daniels/.nvm/versions/node/v14.9.0/bin/node]
[...]

PR #1421, opened 2020-05-11
Test on Node 14

murgatroid99:
Let's see if any current tests reproduce #1407.

### DIFF
--- a/run-tests.bat
+++ b/run-tests.bat
@@ -38,7 +38,7 @@ call npm install || goto :error
 SET JUNIT_REPORT_STACK=1
 SET FAILED=0
 
-for %%v in (8 10 12) do (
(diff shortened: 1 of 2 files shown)


hugebdu:
@murgatroid99 did you succeed to reproduce? 
happens for me pretty constantly (gRPC server)

murgatroid99:
No, I never reproduced it because I don't know what triggers the error. If it happens consistently for you, can you share some code that demonstrates that?

hugebdu:
@murgatroid99 succeeded to recreate 
https://github.com/hugebdu/node14-grpc-core-crashes

using node v14.9

apparently the problem is with `grpc.Metadata`, see my comments in the code

Issue #1170, opened 2019-11-07
Intent to implement: use GitHub Actions for tests


merlinnot:
### Is your feature request related to a problem? Please describe.

I noticed that CI runs must be triggered manually by contributors. I was wondering if we can make this process more optimal by utilizing GitHub Actions, which support Linux, macOS and Windows too.

### Describe the solution you'd like

I'd be willing to work on a PR that introduces CI workflows using GitHub Actions, so these can be triggered automatically for every PR.

### Describe alternatives you've considered

There are many more CIs available, but nowadays GitHub Actions seems to be a to-go solution for any GitHub repository because of great functionalities and very tight integration.

### Additional context

murgatroid99:
Because of the scale of our testing and the amount of computing resources required, we feel that we're better off using our Google-maintained CI infrastructure than what GitHub provides.

In addition, we would want to restrict who can initiate CI runs in either case to prevent people from abusing our computing resources to run their own computationally expensive programs.

merlinnot:
Yes, for your own infrastructure it's definitely better to have control, that's why I was thinking about some external CIs that have protections against these kind of things built-in.

That's a shame, but I'll just try to always make sure tests are passing locally.

Do you think it would be possible to use GitHub Actions for tasks like `gts check`, so at least this part is more contributor-friendly?

PR #1425, opened 2020-05-13
Delete grpc-native-core from master, along with all references

murgatroid99:
I also copied some `.proto` files that were being referenced from the submodule into `test/proto`.

Output of `git grep -- -core`:

```
README.md:Directory: [`packages/grpc-native-core`](https://github.com/grpc/grpc-node/tree/grpc@1.24.x/packages/grpc-native-core) (lives in the `grpc@1.24.x` branch) (see here for installation information)
packages/grpc-native-core/README.md:The `grpc` npm pacakge previously lived in this package. It has been removed. It now lives in the [`grpc@1.24.x` branch](https://github.com/grpc/grpc-node/tree/grpc%401.24.x/packages/grpc-native-core) where bugfixes are still being applied.
```

In the `test` directory tests that test all combinations of client and server from both libraries, I removed only the set that uses both the client and server from the native library, and I switched to installing the native library from `npm` instead of using the local copy. This way, we can still test that `grpc-js` matches the behavior of `grpc`, and I think the interop tests should still work.

After/in parallel to this, I think I can easily migrate the `grpc-tools` build to a GitHub Workflow and remove the Kokoro build tests entirely. I would also like to migrate the regular tests to a GitHub Workflow, but there doesn't seem to currently be a good way to get the per-test reports that we currently have.


murgatroid99:
It is expected that the Electron build jobs fail, because I deleted those config files and build scripts.

murgatroid99:
The Windows Build Job has been failing due to an issue that is fixed in #1422. That is probably the reason it is failing here.

The Mac build job config files were also deleted so I expect that to fail too.

merlinnot:
My offer still stands 🙂 https://github.com/grpc/grpc-node/issues/1170

murgatroid99:
@merlinnot Do you know how to get good per-test output comparable to [what we currently get from Kokoro](https://source.cloud.google.com/results/invocations/5b5a3b15-3d6e-4b1a-97fc-345c5f37a495/targets/github%2Fgrpc-node%2Freports%2Fnode12/tests) from GitHub Actions?

merlinnot:
That looks similar to GitHub annotations. I personally only used that for ESLint before, to annotate errors and warnings, but it looks like people are also using it for failed tests:

fail

Did I understand your intention correctly, to have an output with failures only?

murgatroid99:
Yes, that looks like what I'm looking for: separate reports for each test failure, along with specific output from each test. I found [mocha-github-actions-reporter](https://github.com/daniellockyer/mocha-github-actions-reporter) which probably does that, though I'm not entirely sure how it's used.

PR #1426, opened 2020-05-14
Add grpc-tools build GitHub Workflow

murgatroid99:
This is a replacement for the current Kokoro build configs.


murgatroid99:
@merlinnot Can you help me out here? I can't tell why this workflow isn't doing anything.

merlinnot:
Hey, I see you just found the issue 🙂

murgatroid99:
I did not realize until after I wrote that comment that I got an email that said what the error in the workflow was.

PR #1435, opened 2020-05-21
grpc 1.24 updates from master

murgatroid99:
Cherry pick #1371 and #1224, add the new versions of Node and Electron to the build matrix, and bump the version.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
     name: All Simple Native Builds
     strategy:
       matrix:
-        node: ['4', '5', '6', '7', '8', '9', '10', '11', '12', '13']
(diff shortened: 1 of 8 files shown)


edgareler:
Is the pre-deprecation warning at https://github.com/grpc/grpc-node/pull/1435/commits/acffaf27eed4a5d95a4c1223f8b67ae3b249ab36 only valid to the version 1.24 or to the entire grpc-native-core lib?

murgatroid99:
It applies to the entire library. The library is only going to get 1.24.x versions in the future.

edgareler:
As this library is going to be deprecated in less then a year, what is the long term solution to use gRPC on Node? [@grpc/grpc-js](https://github.com/grpc/grpc-node/tree/master/packages/grpc-js)?

murgatroid99:
Yes, our development efforts will be focused on `@grpc/grpc-js` going forward.

PR #1436, opened 2020-05-21
grpc-js: Don't push messages after ending a call

murgatroid99:
Possible fix for googleapis/gax-nodejs#826


amiller-gh:
Can confirm that this change fixes my issue in googleapis/gax-nodejs#826

murgatroid99:
That's great. I'll try to get a patch release out soon.

PR #1443, opened 2020-05-26
Move google-auth-library to a peer dependency

### LINKED ISSUE #1442: Dependency on google-auth-library 6 breaks Node 8 support
### Problem description
grpc-js claims support for Node 8 or Node 10:
https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/package.json#L9

However a recent commit added a new dependency on `google-auth-library@6`:
https://github.com/grpc/grpc-node/commit/010ef569f7f83007642f3f69e0ce13dee615d680

This release of google-auth-library requires Node 10+ in the engines field:
https://github.com/googleapis/google-auth-library-nodejs/releases/tag/v6.0.0

### Reproduction steps
This was reported to the Firebase CLI as a breaking issue on our side:
https://github.com/firebase/firebase-tools/issues/2284

### Environment
 - OS name, version and architecture: macOS
 - Node version: 8.x
 - If applicable, compiler version: N/A
 - Package name and version: grpc-js@1.0.4

### Additional context
None

murgatroid99:
This fixes #1442. The code path that calls into `google-auth-library` can't actually be triggered from the current public APIs, so if someone doesn't have the peer dependency they'll be fine. Firebase depends on `google-gax` 1.x, which depends on `google-auth-library` 5.x, and other client libraries depend on `google-gax` 2.x, which depends on `google-auth-library` 6.x. The peer dependency will be satisfied in either case.

We also need to add the dev dependency because we need it at build time for the type definitions.

Once the xDS stuff is done (or we want to create a `credentials.createGoogleDefault` public API) and that code path can actually be triggered, we should switch back to a direct dependency on `google-auth-library` 5.x to ensure that we still support Node 8. Hopefully we can eventually drop Node 8 and switch the dependency to 6.x.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -26,6 +26,7 @@
     "@types/semver": "^6.0.1",
     "clang-format": "^1.0.55",
     "execa": "^2.0.3",
+    "google-auth-library": "^6.0.0",
     "gts": "^2.0.0",
     "gulp": "^4.0.2",
     "gulp-mocha": "^6.0.0",
@@ -56,8 +57,10 @@
     "posttest": "npm run check"
   },
   "dependencies": {
-    "google-auth-library": "^6.0.0",
     "semver": "^6.2.0"
+  },
+  "peerDependencies": {
+    "google-auth-library": "5.x || 6.x"
   },
   "files": [
     "src/*.ts",


samtstern:
@murgatroid99 really appreciate the quick fix!

AviVahl:
This change caused a `peerDependency` warning in every project using `firebase`. :/

Can someone clarify how is `createGoogleDefaultCredentials` even used? it's not referenced within repo, and is not even exported? should it even be in this package?

murgatroid99:
It's an internal part of an in-progress feature (the "xDS stuff" I mentioned in the description), and it might become a public API in the future. As mentioned, once that feature is complete, this will switch to a direct dependency on `google-auth-library`. Until then, I think having this peer dependency is probably the least-bad option.

AviVahl:
I can't say I agree with you. The current approach is adding a warning to build logs just to keep supporting an EOL version of Node. Droping Node 8 (via a major release) is probably the "least bad" option in this case.

murgatroid99:
Dropping Node 8 is not an option. We need to support some platforms that are still on Node 8.

The viable options that I can see here are:

 - Have the peer dependency as it currently exists. The cost is a one-line warning in many build logs.
 - Add a direct dependency on `google-auth-library` version `5.x`. The cost is a spurious package in most people's dependency trees.
 - Have no dependency. The cost is a disconnect between `require` calls in the code and the dependency list, though that `require` call is currently unreachable from the public API.

AviVahl:
I have ~10 warnings already in my log. Why? Because 10 different people, of very common libraries and tools out there, decided that a "one-line" warning is acceptable. For me, it feels like a failure to see the rest of the ecosystem. :(

The third option you've mentioned sounds much better than the first, considering this is an in-progress feature not part of the public API.

murgatroid99:
You convinced me. Version 1.1.2 has deleted that peer dependency.

PR #1446, opened 2020-05-27
grpc-js: transition out of TRANSIENT_FAILURE if backoff timer has ended

### LINKED ISSUE #1434: Call stuck connecting when server is initially unavailable, but becomes available later
### Problem description
I have a (C#) grpc server and (grpc-js) client, which I start simultaneously and connect them over loopback (127.0.0.1). Usually the server becomes ready *after* the client, thus client will often observe nothing listening on a port initially.

Prior to grpc-js 0.7.5 it worked all quite fine, client and some initial calls that I start waited for the server to get ready and connected silently. Apparently 0.7.5 introduced some fixes that made calls fail-fast when client is not connected, so I modified code passing waitForReady:true in metadata.

Unfortunately, this didn't bring the previous behavior. Now the client / calls are stuck and never connect successfully.

### Reproduction steps
1. clone https://github.com/kskalski/snippets/tree/master/GrpcJsElectronSharp
2. cd GrpcJsElectronSharp && ./build.sh (adjust paths as needed in the script)
3. run ./Server.exe
4. run ./Client.exe
5. observe server receiving messages whenever Client's window is minimized (handler triggering message send)
6. stop Client and Server
7. run ./Client.exe
8. run ./Server.exe
9. observe server no longer receives messages when Client's window is minimized
(the calls are stuck on connecting)

### Environment
 - OS name, version and architecture: Windows 10, x64
 - Node version: Electron 8.3.0 (Node 12)
 - Node installation method Electron
 - If applicable, compiler version .NET 4.6
 - Package name and version 0.7.5 (it happens also on 1.0.3, but I was able to backtrack that the problem appeared between 0.7.4 and 0.7.5)

### Additional context
I poked around a bit and it seems that there is TRANSIENT_FAILURE in channel.js (this piece of code executed):
```            
case picker_1.PickResultType.TRANSIENT_FAILURE:
                if (callMetadata.getOptions().waitForReady) {
                    this.pickQueue.push({ callStream, callMetadata });
```
but nothing happens afterwards, there is never completion or other result of re-connection attempt

murgatroid99:
If the backoff timer ends before the connection attempt fails, the subchannel needs to transition into TRANSIENT_FAILURE briefly to signal to the channel that the connection attempt failed, but then it needs to transition out of TRANSIENT_FAILURE to be able to try to connect again.

This fixes #1434.

### DIFF
--- a/packages/grpc-js/src/subchannel.ts
+++ b/packages/grpc-js/src/subchannel.ts
@@ -231,19 +231,23 @@ export class Subchannel {
       maxDelay: options['grpc.max_reconnect_backoff_ms'],
     };
     this.backoffTimeout = new BackoffTimeout(() => {
-      if (this.continueConnecting) {
-        this.transitionToState(
-          [ConnectivityState.TRANSIENT_FAILURE],
-          ConnectivityState.CONNECTING
-        );
-      } else {
-        this.transitionToState(
-          [ConnectivityState.TRANSIENT_FAILURE],
-          ConnectivityState.IDLE
-        );
-      }
+      this.handleBackoffTimer();
     }, backoffOptions);
     this.subchannelAddressString = subchannelAddressToString(subchannelAddress);
+  }
+
+  private handleBackoffTimer() {
+    if (this.continueConnecting) {
+      this.transitionToState(
+        [ConnectivityState.TRANSIENT_FAILURE],
+        ConnectivityState.CONNECTING
+      );
+    } else {
+      this.transitionToState(
+        [ConnectivityState.TRANSIENT_FAILURE],
+        ConnectivityState.IDLE
+      );
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


packages/grpc-js/src/subchannel.ts
```diff
@@ -505,6 +509,14 @@ export class Subchannel {
         }
         this.session = null;
         this.stopKeepalivePings();
+        /* If the backoff timer has already ended by the time we get to the
+         * TRANSIENT_FAILURE state, we want to immediately transition out of
+         * TRANSIENT_FAILURE as though the backoff timer is ending right now */
+        if (!this.backoffTimeout.isRunning()) {
+          process.nextTick(() => {
+            this.handleBackoffTimer();
```
nicolasnoble, line 517: Hm. I think I would've liked it slightly better if there was a way to restart the `backoffTimeout` with a deadline of "now", which would've essentially streamlined this.

murgatroid99: Fundamentally, what's happening here is that the backoff timer already triggered, but the result of the timer was deferred until this state transition. It's important here that the `backoffTimeout` is not reset, and that it keeps any backoff it has accumulated.

sidcool1234:
package.json:

```
"@google-cloud/pubsub": "^1.1.6",
    "@grpc/grpc-js": "^1.1.7",
package-lock.json
```

```
"@google-cloud/pubsub": {
      "version": "1.7.3",
      "resolved": "https://registry.npmjs.org/@google-cloud/pubsub/-/pubsub-1.7.3.tgz",
...
"@grpc/grpc-js": {
          "version": "1.0.5",
          "resolved": "https://registry.npmjs.org/@grpc/grpc-js/-/grpc-js-1.0.5.tgz",
          "integrity": "sha512-Hm+xOiqAhcpT9RYM8lc15dbQD7aQurM7ZU8ulmulepiPlN7iwBXXwP3vSBUimoFoApRqz7pSIisXU8pZaCB4og==",
          "requires": {
            "semver": "^6.2.0"
          }
        }
```

PR #1463, opened 2020-06-06
grpc-js: clear noop server call deadline

azban:
addresses: https://github.com/googleapis/nodejs-logging-winston/issues/414

this fixes an issue with open handles when running tests with jest. the timeout is created on import of this file and not cleaned up.

at a glance, i couldn't tell if a singleton was necessary, so this could be solved by just creating a separate timer for each stream, but figured this was closer to the current implementation.

a singleton still has issues if people are creating a server in their tests, but figure this will be rare compared to people just importing grpc-js for client usages

### DIFF
--- a/packages/grpc-js/src/server-call.ts
+++ b/packages/grpc-js/src/server-call.ts
@@ -341,15 +341,13 @@ export type Handler<RequestType, ResponseType> =
 
 export type HandlerType = 'bidi' | 'clientStream' | 'serverStream' | 'unary';
 
-const noopTimer: NodeJS.Timer = setTimeout(() => {}, 0);
-
 // Internal class that wraps the HTTP2 request.
 export class Http2ServerCallStream<
   RequestType,
   ResponseType
 > extends EventEmitter {
   cancelled = false;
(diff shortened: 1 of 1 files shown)


murgatroid99:
I don't understand how this would fix anything. Why would a timer that ends immediately be an "open handle"?

azban:
not totally sure, but potentially jest is trying to exit before a new event loop is started and timers are processed?

murgatroid99:
I guess I can understand how that could cause problems. In that case I wouldn't object to fixing this, but this singleton is too heavyweight for what we actually need to do here. The purpose of setting `deadline` to a zero-length, no-op timer is just to ensure that it has the correct type. The fix I would prefer is to do what we've done in a few other places: in the constructor, set the deadline to a timer created the same way, then immediately cancel it:

```ts
this.deadline = setTimeout(() => {}, 0);
clearTimeout(this.deadline);
```

azban:
that works for me. didn't know if there was another reason to share it across call streams or keep it around.

murgatroid99:
This change is now out in version 1.1.0.

PR #1465, opened 2020-06-08
feat(client): export `ClientOptions` type

SimenB:
We have a wrapper around the base Client, and being able to annotate with this type helps it be a bit more seamless


murgatroid99:
This change is now out in version 1.1.0.

PR #1467, opened 2020-06-08
grpc-js: server: cull closed sessions from list, check for closed in tryShutdown

### LINKED ISSUE #1458: @grpc/grpc-js: tryShutdown freeze app
### Problem description
If client crashes or exits before server calls `call.end()`, `server.tryShutdown(callback)` freezes app. Server has http2session which is closed, so `maybeCallback` doesn't decrement `pendingChecks`, and `callback` is never called (if tryShutdown was promisified it never resolves).

### Reproduction steps
Create server-side stream (hellostreamingworld.proto), end client app before server calls `call.end()`.
(client - greater_client.js without changes)  
```
const promisify = require("util").promisify;
const log = require('why-is-node-running');
var messages = require('./hellostreamingworld_pb');
var services = require('./hellostreamingworld_grpc_pb');
var grpc = require('@grpc/grpc-js');

function sayHello(call) {
  var num = call.request.getNumGreetings();
  console.log(call.request.toObject());
  console.log(`${num} greetings`);
  function greet() {
        if (num-- < 0) {
  		return;
  	}
	var reply = new messages.HelloReply();
	reply.setMessage('Hello ' + call.request.getName());
    call.write(reply);
    setTimeout(greet, 2000);
  }
  setTimeout(greet, 1000);
  call.end();
}

async function main() {
	var server = new grpc.Server();
	server.addService(services.MultiGreeterService, {sayHello: sayHello});
	await promisify(server.bindAsync.bind(server))('0.0.0.0:50051', grpc.ServerCredentials.createInsecure());
	server.start();
	setTimeout(() => {
		console.log('start server shutdown');
		setTimeout(log, 10000);
		server.tryShutdown(() => {
			console.log("server shutdown");
		});
	}, 10000);
}

main();
```
### Environment
 - OS name, version and architecture: ArchLinux x86_64
 - Node version: v12.18.0
 - Package name and version [e.g. "@grpc/grpc-js": "^0.6.16"]

### Additional information
server output:

```
{ name: 'world', numGreetings: 10 }
10 greetings
start server shutdown
There are 8 handle(s) keeping the process running

# TTYWRAP
/home/maxim/code/grpc/examples/node/node_modules/grpc/node_modules/set-blocking/index.js:2 - [process.stdout, process.stderr].forEach(function (stream) {
/home/maxim/code/grpc/examples/node/node_modules/grpc/node_modules/npmlog/log.js:11        - setBlocking(true)

# SIGNALWRAP
/home/maxim/code/grpc/examples/node/node_modules/grpc/node_modules/set-blocking/index.js:2 - [process.stdout, process.stderr].forEach(function (stream) {
/home/maxim/code/grpc/examples/node/node_modules/grpc/node_modules/npmlog/log.js:11        - setBlocking(true)

# TTYWRAP
/home/maxim/code/grpc/examples/node/node_modules/grpc/node_modules/set-blocking/index.js:2 - [process.stdout, process.stderr].forEach(function (stream) {
/home/maxim/code/grpc/examples/node/node_modules/grpc/node_modules/npmlog/log.js:11        - setBlocking(true)

# DNSCHANNEL
(unknown stack trace)

# HTTP2SESSION
(unknown stack trace)

# HTTP2SETTINGS
(unknown stack trace)

# Timeout
/home/maxim/code/grpc/examples/node/static_codegen/greeter_server.js:55 - setTimeout(log, 10000);

# Timeout
/home/maxim/code/grpc/examples/node/static_codegen/greeter_server.js:38 - setTimeout(greet, 2000);

```

murgatroid99:
As currently written, the `sessions` field tracks every HTTP/2 session that has ever been established with the server, which can grow without bound. If we remove the ones that are closed, it should be bounded by network connection limits. The check for `session.closed` in `tryShutdown` should avoid potential race conditions, where we get to that line and the `close` event is pending but has not fired yet.

I think this will fix #1458.

### DIFF
--- a/packages/grpc-js/src/server.ts
+++ b/packages/grpc-js/src/server.ts
@@ -500,7 +500,6 @@ export class Server {
       }
     }
 
-    pendingChecks += this.sessions.size;
     this.sessions.forEach((session) => {
       if (!session.closed) {
         pendingChecks += 1;
--- a/packages/grpc-js/src/server.ts
+++ b/packages/grpc-js/src/server.ts
@@ -500,10 +500,12 @@ export class Server {
       }
     }
 
-    // If any sessions are active, close them gracefully.
     pendingChecks += this.sessions.size;
     this.sessions.forEach((session) => {
-      session.close(maybeCallback);
+      if (!session.closed) {
+        pendingChecks += 1;
+        session.close(maybeCallback);
+      }
     });
     if (pendingChecks === 0) {
       callback();
@@ -608,6 +610,10 @@ export class Server {
       }
 
       this.sessions.add(session);
+
+      session.on('close', () => {
+        this.sessions.delete(session);
+      });
     });
   }
 }


packages/grpc-js/src/server.ts
```diff
@@ -500,10 +500,12 @@ export class Server {
       }
     }
 
-    // If any sessions are active, close them gracefully.
     pendingChecks += this.sessions.size;
     this.sessions.forEach((session) => {
-      session.close(maybeCallback);
+      if (!session.closed) {
+        pendingChecks += 1;
```
nicolasnoble, line 506: Took me a second to realize why this was added. I don't like the fact there's a disjoint between adding a new pending action, and incrementing this counter: this feels brittle. But I don't see a way to do this without over engineering it completely:

```javascript
  function addAction(action) {
    pendingChecks++;
    action(maybeCallback);
}

[...]

addAction(callback => session.close(callback));
```

murgatroid99: I don't know if this is worth changing here but I'm glad you brought it up because there was actually an error. See the new commit.

PR #1468, opened 2020-06-08
grpc-js: Export propagate constants for type parity with grpc

### LINKED ISSUE #1457: @grpc/grpc-js: Missing propagate constants & types
### Problem description
The `@grpc/grpc-js` package is missing `propagate` types, and thus does not export them from the main package.

Screenshot 2020-06-03 17 02 42

Their use is in the gRPC docs: https://grpc.github.io/grpc/node/grpc.html#.propagate__anchor 

These exist in the native gRPC package:
https://github.com/grpc/grpc-node/blob/grpc%401.24.x/packages/grpc-native-core/src/constants.js#L169

```
/**
 * Propagation flags: these can be bitwise or-ed to form the propagation option
 * for calls.
 *
 * Users are encouraged to write propagation masks as deltas from the default.
 * i.e. write `grpc.propagate.DEFAULTS & ~grpc.propagate.DEADLINE` to disable
 * deadline propagation.
 * @memberof grpc
 * @alias grpc.propagate
 * @enum {number}
 */
exports.propagate = {
  DEADLINE: 1,
  CENSUS_STATS_CONTEXT: 2,
  CENSUS_TRACING_CONTEXT: 4,
  CANCELLATION: 8,
  DEFAULTS: 65535
};
```

These do NOT exist in the pure JS package, and thus I ran into type errors when switching from the native package:
https://www.dropbox.com/s/yidakww7zzv2zhk/Screenshot%202020-06-03%2017.02.42.png?dl=0

### Reproduction steps
Give very precise steps you've discovered to reproduce your problem. If possible and applicable, provide us with a repository we can clone that contains a reproduction case. Also if possible and applicable, please include a Dockerfile that exhibits the problem if it's specific to a certain environment. Bug reports with no reproduction steps will be closed.

### Environment
 - OS name, version and architecture: MacOS 10.14.6
 - Node version: 12.13.1
 - Node installation method: homebrew
 - If applicable, compiler version: n/a
 - Package name and version: `@grpc/grpc-js` latest

### Additional context

murgatroid99:
This fixes #1457

### DIFF
--- a/packages/grpc-js/src/index.ts
+++ b/packages/grpc-js/src/index.ts
@@ -36,7 +36,7 @@ import {
   CallProperties,
   UnaryCallback,
 } from './client';
-import { LogVerbosity, Status } from './constants';
+import { LogVerbosity, Status, Propagate } from './constants';
 import * as logging from './logging';
 import {
   Deserialize,
@@ -127,6 +127,7 @@ export {
   LogVerbosity as logVerbosity,
   Status as status,
   ConnectivityState as connectivityState,
+  Propagate as propagate,
   // TODO: Other constants as well
 };
 
--- a/packages/grpc-js/src/constants.ts
+++ b/packages/grpc-js/src/constants.ts
@@ -41,6 +41,18 @@ export enum LogVerbosity {
   ERROR,
 }
 
+/**
+ * NOTE: This enum is not currently used in any implemented API in this
+ * library. It is included only for type parity with the other implementation.
+ */
+export enum Propagate {
+  DEADLINE = 1,
+  CENSUS_STATS_CONTEXT = 2,
+  CENSUS_TRACING_CONTEXT = 4,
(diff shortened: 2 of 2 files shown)


packages/grpc-js/src/index.ts
```diff
@@ -127,6 +127,7 @@ export {
   LogVerbosity as logVerbosity,
   Status as status,
   ConnectivityState as connectivityState,
+  Propagate as propagate,
```
vlucas, line 130: These should also be exported from the main package so they can be accessed off the `grpc` namespace like other packages (like the native node one).

murgatroid99: This is the main package.

vlucas: Ah, my bad. the way GitHub folded the code, it looked like it was part of a separate import statement.

PR #1474, opened 2020-06-12
proto-loader: Add TypeScript generator

murgatroid99:
This is an implementation of grpc/proposal#183. This implementation is quick and dirty but I want to have something here so that I can use it.


packages/proto-loader/bin/proto-loader-gen-types.ts
```diff
@@ -0,0 +1,587 @@
...
+        type = 'number | string | Long';
+        break;
+      case 'bool':
+        type = 'boolean';
+        break;
+      case 'string':
+        type = 'string';
+        break;
+      case 'bytes':
+        type = 'Buffer | UInt8Array | String';
```
alexander-fenster, line 128: maybe lowercase `string`?

alexander-fenster: also, `Uint8Array` (lowercase `i`)

alexander-fenster:
It would be very useful to have this covered by a baseline test (if a generated `.d.ts` for some test proto is checked in so that subsequent changes to a tool triggered visible changes to the `.d.ts` file in the same PR).

murgatroid99:
The new draft is out in version 0.6.0-pre2

packages/proto-loader/bin/proto-loader-gen-types.ts
```diff
@@ -0,0 +1,625 @@
...
+  const typeInterfaceName = getTypeInterfaceName(dependency);
+  const importedTypes = dependency instanceof Protobuf.Type ? `${dependency.name} as ${typeInterfaceName}, ${dependency.name}__Output as ${typeInterfaceName}__Output` : `${dependency.name} as ${typeInterfaceName}`;
+  return `import { ${importedTypes} } from '${filePath}';`
+}
+
+function generatePermissiveMessageInterface(formatter: TextFormatter, messageType: Protobuf.Type) {
+  if (messageType.fullName === '.google.protobuf.Any') {
+    /* This describes the behavior of the Protobuf.js Any wrapper fromObject
+     * replacement function */
+    formatter.writeLine('export type Any__Output = AnyExtension | {');
```
murgatroid99, line 102: Reminder to self: this needs to be `Any`, not `Any__Output`.

murgatroid99:
The last commit changes a few things:

 - When fields are added by extensions, the field names are some kind of qualified name and have periods. So, all field names are now quoted in the generated code.

 - There are some inconsistencies in how Protobuf.js refers to nested message and enum type names in different contexts. To avoid the issue, nested message and enum types are now generated in the same file as the parent message.

 - All generated files now have a comment that states what `.proto` file the generated types come from.

The new version is now available in version 0.6.0-pre4 of the library.

badsyntax:
I'm giving this a spin and noticed the following problems:

Using this proto file:

```proto
// echo_service.proto
syntax = "proto3";

message EchoMessage {
  string value = 1;
  int32 value2 = 2;
}

service EchoService {
  rpc Echo (EchoMessage) returns (EchoMessage);
  rpc EchoClientStream (stream EchoMessage) returns (EchoMessage);
  rpc EchoServerStream (EchoMessage) returns (stream EchoMessage);
  rpc EchoBidiStream (stream EchoMessage) returns (stream EchoMessage);
}
```

The generated client interface seems ok, but there's a couple of issues with the server interface. 

Here's the generated code:

```ts
export namespace ServiceHandlers {
  export namespace EchoMessage {
  }
  export interface EchoService {
    Echo(call: grpc.ServerUnaryCall<messages.EchoMessage__Output>, callback: grpc.SendUnaryData<messages.EchoMessage>): void;
    
    EchoClientStream(call: grpc.ServerReadableStream<messages.EchoMessage__Output>, callback: grpc.SendUnaryData<messages.EchoMessage>): void;
    
    EchoServerStream(call: grpc.ServerWriteableStream<messages.EchoMessage__Output, messages.EchoMessage>): void;
    
    EchoBidiStream(call: grpc.ServerDuplexStream<messages.EchoMessage__Output, messages.EchoMessage>): void;
    
  }
}
```

Issues: 

1. `grpc.ServerUnaryCall<messages.EchoMessage__Output>` is not correct, `ServerUnaryCall` is [expecting two types](https://github.com/grpc/grpc-node/blob/95289edcaf36979cccf12797cc27335da8d01f03/packages/grpc-js/src/server-call.ts#L83-L85).
2. `grpc.ServerWriteableStream` should be [`grpc.ServerWritableStream`](https://github.com/grpc/grpc-node/blob/95289edcaf36979cccf12797cc27335da8d01f03/packages/grpc-js/src/server-call.ts#L90-L94)
3. `grpc.SendUnaryData` should be [`grpc.sendUnaryData`](https://github.com/grpc/grpc-node/blob/95289edcaf36979cccf12797cc27335da8d01f03/packages/grpc-js/src/server-call.ts#L269-L274)

murgatroid99:
@badsyntax Thank you for the fixes. I published the changes to 0.6.0-pre6.

murgatroid99:
@alexander-fenster Do you have a suggestion of an existing `.proto` file that can be used for a test like that?

alexander-fenster:
@murgatroid99 We use https://github.com/googleapis/gapic-showcase/blob/master/schema/google/showcase/v1beta1/echo.proto for testing our client library generators. It covers pretty much all the use cases supported by Google APIs.

PHILLIPS71:
Giving this a go, pretty much exactly what I've been looking for. I'm running into a couple of issues where I've just been going off some documentation I've found here [https://github.com/grpc/proposal/pull/183/files/577fa399cb28dbde07b41896588f2e6998699a8f](url) so please correct me If I'm wrong.

> Conversion of type 'GrpcObject' to type 'ProtoGrpcType' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.

```javascript
const pkg = grpc.loadPackageDefinition(definition) as ProtoGrpcType;
```


> Argument of type 'UserService' is not assignable to parameter of type 'UntypedServiceImplementation'.
  Index signature is missing in type 'UserService'.ts(2345)
```javascript
 server.addService(pkg.user.UserService.service, handler);
```

Here is the handler that the above is complaining about where I try to add it as a service implementation
```javascript
const handler: ServiceHandlers.user.UserService = {
  async findById(call, callback) {
    const user = await prisma.user.findOne({
      where: {
        id: call.request?.id,
      },
    });

    callback(null, user);
  },
};
```

Here is the proto file I've been using to experiment with
```protobuf
syntax = "proto3";
package user;

service UserService {
  rpc findById (FindByIdRequest) returns (User) {}
}

message FindByIdRequest {
  string id = 1;
}

message User {
  string id = 1;
  string first_name = 2;
  string last_name = 3;
  string email = 4;
  bool is_verified = 5;
}
```

murgatroid99:
I actually saw that first warning in my own use of the generator, but I forgot to update the proposal. `as unknown as ProtoGrpcType` seems to work fine, and I don't see a way to improve that.

I didn't try out the service handler types myself so I didn't see that warning. I don't know if there's going to be a good way to avoid another typecast there.

murgatroid99:
I published another version, 0.6.0-pre10 with some significant changes based on a comment on the proposal:

 - The `ServiceNameClient` and `ServiceNameHandler` interfaces are now exported from `generated/fully/qualified/package/ServiceName.ts`, similar to the existing files generated for messages and enums.

 - The root files now only export the `ProtoGrpcType` interface. Everything else can be found in other files.

murgatroid99:
The last commit, published as 0.6.0-pre11, fixes another couple of bugs:

 - Message fields were previously incorrectly marked as non-optional in `__Output` types when the `defaults` option was set.

 - Map fields were previously generated as just being a field with their value type, instead of as a map type.

PHILLIPS71:
@murgatroid99 is there any planned support for promises in this proposal in addition to / in favor of the callback method that it already generates?

murgatroid99:
The code that this tool generates is purely descriptive: all of those APIs already exist, this just provides TypeScript typings for them. The existing APIs do not have promises, so the generated code does not have promises either.

murgatroid99:
@alexander-fenster I have added "golden" files generated from the suggested `.proto` file, along with a test that they are up to date.

adrien-f:
Greetings!

Been following the project for a while, thanks you especially for the latest pre-release with support for maps!

I've been using it to generate TS interfaces so I can create messages in JSON for a grpc-gateway. One message has a `google.protobuf.Timestamp` field which decodes to:

```
export interface Timestamp {
  'seconds'?: (number | string | Long);
  'nanos'?: (number);
}
```

Now this is technically right but the JSON deserializer is expecting an ISO string: https://github.com/golang/protobuf/blob/master/jsonpb/decode.go#L231

Would you happen to know a way to have the best of both worlds or I'm totally doing the wrong thing here? Thanks in advance!

murgatroid99:
@adrien-f The generated interfaces describe the input to `Message.fromObject` and the output from `Message#toObject` in the Protobuf.js library. Protobuf.js does not have special handling for `Timestamp`, so no special interfaces are generated for `Timestamp`. There is an open issue in that repository about supporting that behavior: protobufjs/protobuf.js#1304.

pcj:
This is awesome 💯 .  Pretty much worked first try 😲 

Here's my steps: 

### `package.json`

```json
{
	"scripts": {
		"compile": "tsc -p ./",
		"lint": "eslint src --ext ts",
		"watch": "tsc -watch -p ./",
		"pretest": "yarn run compile && yarn run lint",
		"test": "node ./out/test/runTest.js",
		"genproto-ts": "node ./node_modules/@grpc/proto-loader/build/bin/proto-loader-gen-types.js --keepCase --longs=String --enums=String --defaults --oneofs --json --includeComments -O ./src/proto --grpcLib @grpc/grpc-js src/bezel/client/application.proto"
	},
	"dependencies": {
		"@grpc/grpc-js": "1.1.1",
		"@grpc/proto-loader": "0.6.0-pre11"
	},
	"devDependencies": {
		"@types/node": "^13.11.0",
		"eslint": "^6.8.0",
		"@typescript-eslint/parser": "^2.30.0",
		"@typescript-eslint/eslint-plugin": "^2.30.0",
		"typescript": "^3.8.3"
	}
}
```

### Invoke Generator

```sh
$ npm run genproto-ts

> bazel-stack-vscode@0.3.1 generate-application-proto-ts /Users/pcj/go/src/github.com/stackb/bzl-vscode
> node ./node_modules/@grpc/proto-loader/build/bin/proto-loader-gen-types.js \
  --keepCase \
  --longs=String \
  --enums=String \
  --defaults \
  --oneofs \
  --json \
  --includeComments \
  -O ./src/generated \
  --grpcLib @grpc/grpc-js \
  src/bezel/client/application.proto
```

### Generated Outputs

```sh
$ find src/proto
src/proto
src/proto/google
src/proto/google/protobuf
src/proto/google/protobuf/Timestamp.ts
src/proto/application.ts
src/proto/build
src/proto/build/stack
src/proto/build/stack/bzl
src/proto/build/stack/bzl/v1beta1
src/proto/build/stack/bzl/v1beta1/ApplicationMetadata.ts
src/proto/build/stack/bzl/v1beta1/Application.ts
src/proto/build/stack/bzl/v1beta1/GetApplicationMetadataRequest.ts
```

There are some nits regarding the generated code; missing semicolons, unused imports, etc.

Screen Shot 2020-07-22 at 10 03 25 PM
 
### Setup

```typescript
import * as loader from '@grpc/proto-loader';
import * as path from 'path';

import { ProtoGrpcType } from '../../proto/application';
import { ApplicationClient } from '../../proto/build/stack/bzl/v1beta1/Application';

const protoFilePath = path.normalize(`${__dirname}/../../../src/bezel/client/application.proto`);

let protoPackage: any = loader.loadSync(protoFilePath, {
    keepCase: true,
    longs: String,
    enums: String,
    defaults: false,
    oneofs: true
});

/**
 * @see https://github.com/murgatroid99/proposal/blob/a872c74877b7a388320210ea6412c017e29b76eb/L70-node-proto-loader-type-generator.md#proposal
 */
const proto = grpc.loadPackageDefinition(protoPackage) as unknown as ProtoGrpcType;

/**
 * Exported type+implementations of the v1beta1 package.
 */
export const v1beta1 = proto.build.stack.bzl.v1beta1;

/**
 * Create a new client for the Application service.
 * 
 * @param address The address to connect.
 */
export function newApplicationClient(address: string): ApplicationClient {
    const creds = grpc.credentials.createInsecure();
    return new v1beta1.Application(address, creds);
}
```

### Usage

```typescript
import { newApplicationClient } from "./client/application";
import { ApplicationMetadata } from "../proto/build/stack/bzl/v1beta1/ApplicationMetadata";

...

        const client = newApplicationClient(cfg.grpcServerAddress);

        client.getApplicationMetadata({}, new grpc.Metadata(), (err?: grpc.ServiceError, resp?: ApplicationMetadata) => {
            if (err) {
                console.warn(`could not rpc application metadata: ${err}`);
                return;
            }
            console.info(`connected to bezel ${resp?.name} ${resp?.version}`);
        });

    }

```

### UX

proto-loader


And lo and behold, it worked!

Great job @murgatroid99!

adrien-f:
@murgatroid99 Thanks for the answer, this makes sense, I'll follow the issue there 👍

murgatroid99:
@pcj I want to note that in your generator invocation script, you pass the `--json` argument, but you don't set that option in your `loader.loadSync` call. The expectation is that those options will be the same, and if they are mismatched that could result in some inconsistencies between the compile time and runtime types.

I can fix the missing semicolon because that was just an oversight, but that unused import may stick around because having it simplifies the code generation logic.

pcj:
@murgatroid99 Have you considered whether the boilerplate code above (titled "Setup") should be generated as well?  Seems like a good candidate.

murgatroid99:
@pcj I hadn't considered that. I think if I do it, the API will look a bit more like this:

```ts
export function loadProtos(): Promise<ProtoGrpcType>;

export function loadProtosSync(): ProtoGrpcType;
```

The loaded object isn't limited to exactly one service, so it wouldn't make sense to try to extract it the way your setup code does. And I'd prefer that the code not be opinionated about loading synchronously vs asynchronously because I think there are valid use cases for both.

In the end, that would cut down a few imports and 3 lines of boilerplate (`path.normalize`, `loader.loadSync`, and `grpc.loadPackageDefinition`) to one import and one line of code.

murgatroid99:
There's one thing about generating the boilerplate that I'm not sure how to handle: It is possible to pass multiple filenames in to the TypeScript generator, and to the `load` and `loadSync` functions. Currently, the generator just generates a separate top-level TypeScript file for each input file, but it can be useful to load all of the files together at runtime. So the question then is: should we try to generate a function that loads all of the files, and if so, what should that file be called?

pcj:
Perhaps that can be an option to generate a function that loads all at one, faintly analogous to `option java_multiple_files = true`?

murgatroid99:
This PR has already gotten quite long. I think it would be best to keep the scope as it currently is, and consider additional enhancements like boilerplate generation in later changes.

murgatroid99:
The last commit fixes a few more issues:

 - Have the `...Handler` interfaces extend `grpc.UntypedServiceImplementation` so that they can be passed to `Server#addService` without a typecast.

 - In the hander interfaces, make the request type use the `__Output` interface, instead of the response type.

 - In the service client interfaces, replace some duplicate function signatures with some missing type signatures.

jonaskello:
I was experimenting with version `0.6.0-pre14` and had a `.proto` file that did not have any `package` specified. In this case the `import` statements were generated like `import Foo from 'Foo'`. I had to manually correct them to `import Foo from './Foo'`. Added `package` to the proto and now it works like expected. However it may be interesting to fix since I think using `package` is optional?

ryanrhee:
I noticed that the latest commit removes the pre-release version and is using `0.6.0`. Is there an ETA on when this PR will be merged?

jonaskello:
It seems like two types are generated for each message, one with required fields and one with optional fields. I think this is to reflect that grpc allows sending partial data to save wire format space, but when received all fields are specified and missing ones are filled with default values?

Today something like this is generated:

```ts
// Folder.ts

export interface Folder {
  id?: string;
  name?: string;
}

export interface Folder__Output {
  id: string;
  name: string;
}

// AllRootFoldersForUserRequest.ts

export interface AllRootFoldersForUserRequest {
  userId?: string;
}

export interface AllRootFoldersForUserRequest__Output {
  userId: string;
}

// AllRootFoldersForUserResponse.ts

import { Folder as _UnitFolders_Folder, Folder__Output as _UnitFolders_Folder__Output } from "../UnitFolders/Folder";

export interface AllRootFoldersForUserResponse {
  folders?: _UnitFolders_Folder[];
}

export interface AllRootFoldersForUserResponse__Output {
  folders: _UnitFolders_Folder__Output[];
}

// FoldersService.ts

import {
  AllRootFoldersForUserRequest as _UnitFolders_AllRootFoldersForUserRequest,
  AllRootFoldersForUserRequest__Output as _UnitFolders_AllRootFoldersForUserRequest__Output,
} from "../UnitFolders/AllRootFoldersForUserRequest";
import {
  AllRootFoldersForUserResponse as _UnitFolders_AllRootFoldersForUserResponse,
  AllRootFoldersForUserResponse__Output as _UnitFolders_AllRootFoldersForUserResponse__Output,
} from "../UnitFolders/AllRootFoldersForUserResponse";

export interface UnitFoldersServiceHandlers extends grpc.UntypedServiceImplementation {
  allRootFoldersForUser(
    call: grpc.ServerUnaryCall<
      _UnitFolders_AllRootFoldersForUserRequest__Output,
      _UnitFolders_AllRootFoldersForUserResponse
    >,
    callback: grpc.sendUnaryData<_UnitFolders_AllRootFoldersForUserResponse>
  ): void;
}
```

Typescript has a built-in utility type `Partial<T>` but we can also define `DeepPartial<T>` to add `?` to every field recursively:

```ts
type DeepPartial<T> = {
  [P in keyof T]?: DeepPartial<T[P]>;
};
```

With this defined the above generated code could become:

```ts
// Folder.ts

export interface Folder {
  id: string;
  name: string;
}

// AllRootFoldersForUserRequest.ts

export interface AllRootFoldersForUserRequest {
  userId: string;
}

// AllRootFoldersForUserResponse.ts

import { Folder as _UnitFolders_Folder } from "../UnitFolders/Folder";

export interface AllRootFoldersForUserResponse {
  folders: _UnitFolders_Folder[];
}

// FoldersService.ts

import { AllRootFoldersForUserRequest as _UnitFolders_AllRootFoldersForUserRequest } from "../UnitFolders/AllRootFoldersForUserRequest";
import { AllRootFoldersForUserResponse as _UnitFolders_AllRootFoldersForUserResponse } from "../UnitFolders/AllRootFoldersForUserResponse";

export interface UnitFoldersServiceHandlers extends grpc.UntypedServiceImplementation {
  allRootFoldersForUser(
    call: grpc.ServerUnaryCall<
      _UnitFolders_AllRootFoldersForUserRequest,
      DeepPartial<_UnitFolders_AllRootFoldersForUserResponse>
    >,
    callback: grpc.sendUnaryData<DeepPartial<_UnitFolders_AllRootFoldersForUserResponse>>
  ): void;
}
```

I becomes a little less verbose this way. Not sure if it is worth it, just a suggestion :-).

murgatroid99:
@jonaskello It's not as simple as the fields being optional in one type but not the other. The "input" types are also more permissive in what types they accept for certain fields. For example, for a 64 bit field, the input types accept `(string | number | Long)?`, but for the output type, we know in advance what type the field will have, based on the `--longs` flag passed to the code generator. So, if `--longs=String` is passed, the corresponding field in the output type will just have the type `string`.

You just happened to only use field types for which the only difference between the input and output type is whether the field is optional.

jonaskello:
@murgatroid99 Aha, now the alternate types for the same message makes more sense. Thanks for taking the time to explain it! :-)

packages/proto-loader/bin/proto-loader-gen-types.ts
```diff
@@ -0,0 +1,745 @@
...
+        // Bidi streaming
+        formatter.writeLine(`${methodName}(call: grpc.ServerDuplexStream<${requestType}, ${responseType}>): void;`);
+      } else {
+        // Client streaming
+        formatter.writeLine(`${methodName}(call: grpc.ServerReadableStream<${requestType}, ${responseType}>, callback: grpc.sendUnaryData<${responseType}>): void;`);
+      }
+    } else {
+      if (method.responseStream) {
+        // Server streaming
+        formatter.writeLine(`${methodName}(call: grpc.ServerWritableStream<${requestType}, ${responseType}>): void;`);
```
badsyntax, line 531: Due to a [mismatch of types](https://github.com/grpc/grpc-node/issues/1581), would it better to define the handlers using these types:

- grpc.handleUnaryCall
- grpc.handleServerStreamingCall
- grpc.handleClientStreamingCall
- grpc.handleBidiStreamingCall

This would provide better compatibility between between the `grpc` and `@grpc/grpc-js` packages, allowing you to use this TypeScript generator with the legacy `grpc` package. (See also https://github.com/grpc/grpc-node/pull/1587)

badsyntax: See https://github.com/murgatroid99/grpc-node/pull/2

pcj:
Using this, mostly working great!  I've noticed one issue with imports of a nested message type:  Here is the generated code:

```
// Original file: proto/build_event_stream.proto

import { NamedSetOfFilesId as _build_event_stream_BuildEventId_NamedSetOfFilesId, NamedSetOfFilesId__Output as _build_event_stream_BuildEventId_NamedSetOfFilesId__Output } from '../build_event_stream/BuildEventId/NamedSetOfFilesId';

/**
 * Collection of all output files belonging to that output group.
 */
export interface OutputGroup {
  /**
   * Name of the output group
   */
  'name'?: (string);
  /**
   * List of file sets that belong to this output group as well.
   */
  'fileSets'?: (_build_event_stream_BuildEventId_NamedSetOfFilesId)[];
}

...
```

The issue is that the file `NamedSetOfFilesId` does not exist so the import statement is incorrect.  Here's the proto: https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/buildeventstream/proto/build_event_stream.proto#L108

The dependency I'm currently using is `"@grpc/proto-loader": "0.6.0-pre11",`

murgatroid99:
@pcj I published a change that should have fixed that in version `0.6.0-pre12`, and it's actually up to `-pre14`.

packages/proto-loader/bin/proto-loader-gen-types.ts
```diff
@@ -0,0 +1,745 @@
...
+  }
+  formatter.unindent();
+  formatter.writeLine('}');
+}
+
+function generateServiceHandlerInterface(formatter: TextFormatter, serviceType: Protobuf.Service, options: GeneratorOptions) {
+  if (options.includeComments) {
+    formatComment(formatter, serviceType.comment);
+  }
+  formatter.writeLine(`export interface ${serviceType.name}Handlers extends grpc.UntypedServiceImplementation {`);
```
badsyntax, line 511: I'm wondering if we could potentially **not** extend from `grpc.UntypedServiceImplementation`?

The problem with using `grpc.UntypedServiceImplementation` is you cannot use this interface for implementation classes. 

What i'd like to do:

```ts
class Server implements ServerHandlers {
  // ... implementation
}
```

But this results in: 

```
Class 'Server' incorrectly implements interface 'ServerHandlers'.
  Index signature is missing in type 'Server'.
```

because `grpc.UntypedServiceImplementation` is defined as:

https://github.com/grpc/grpc-node/blob/21adc97c4ad458b9a09bd83c2cadc7fb2a345367/packages/grpc-js/src/server.ts#L92-L94

The fact that we're now generating types for the server handler, perhaps implies we should not be extending from `grpc.UntypedServiceImplementation`?

badsyntax: Hmm, sorry, actually that would not work, because it then would not match the types here:

https://github.com/grpc/grpc-node/blob/21adc97c4ad458b9a09bd83c2cadc7fb2a345367/packages/grpc-js/src/server.ts#L147-L150

`server.addService` requires `UntypedServiceImplementation` type.

murgatroid99: The problem that solved was doing this:

```js
const handlerImplementation: ServerHandlers = {
  method1: ...
  method2: ...
};
server.addService(serviceDefinition, handlerImplementation);
```

Without extending `UntypedServiceImplementation`, the last line was failing with a similar error to the one you quoted, because `addService` expects an `UntypedServiceImplementation`. If we made the change you suggest without any other changes, I think you would be able to implement that `Server` class, but it would not be accepted by `addService`. I don't know what else to do here, so if you have a suggestion that avoids that problem I'm all ears.

badsyntax: Is it worth exploring generic types?

```ts
export declare class Server<T = UntypedServiceImplementation> {
    // ...
    addService(service: ServiceDefinition, implementation: T): void;
    // ...
}
```

```ts
class ExampleServer implements ServerHandlers {
  // ... implementation
}

const server = new grpc.Server<ExampleServer>();
server.addService(proto.chat_package.Chat.service, new ExampleServer());
```

Or we could use generic types on the `addService` method instead.

murgatroid99: It would be better to have generic types on `addService`. The reason `addService` is a method instead of a constructor argument or something is to allow for adding multiple services to a single server. That wouldn't play nicely with class-level generics.

And now that you mention it, there is the open PR #1561, which is adding a generic type to `addService` to solve a slightly different problem, but it might help with this one too.

badsyntax: Thanks for the link to the PR. I see the difficulty now. 

If we add a generic type that defaults to `UntypedServiceImplementation`, we don't enforce any server method type checks. 

```ts
addService<T = UntypedServiceImplementation>(service: ServiceDefinition, implementation: T): void;
```

If we enforce the generic type to be of type `UntypedServiceImplementation`, we can't then implement that type with a class (as typescript will error with "Index signature is missing ").

```ts
addService<T extends UntypedServiceImplementation = UntypedServiceImplementation>(service: ServiceDefinition, implementation: T): void;
```

For the first option, although not ideal, you could argue in a lot of cases the generic type passed to `addService` would be an auto generated type (interface) with method definitions that match the expected method types (`Handler<any, any>`). I still would prefer to have those checks there. I'm not sure how best to achieve this.

badsyntax:
Added a PR here https://github.com/murgatroid99/grpc-node/pull/3 to fix a minor type mismatch for client streaming.

packages/proto-loader/bin/proto-loader-gen-types.ts
```diff
@@ -0,0 +1,745 @@
...
+      importedTypes = `${dependency.name} as ${typeInterfaceName}, ${dependency.name}__Output as ${typeInterfaceName}__Output`;
+    } else if (dependency instanceof Protobuf.Enum) {
+      importedTypes = `${dependency.name} as ${typeInterfaceName}`;
+    } else if (dependency instanceof Protobuf.Service) {
+      importedTypes = `${dependency.name}Client as ${typeInterfaceName}Client`;
+    } else {
+      throw new Error('Invalid object passed to getImportLine');
+    }
+  }
+  return `import { ${importedTypes} } from '${filePath}';`
```
SimenB, line 144: thoughts on adding an option to output [`import type`](https://devblogs.microsoft.com/typescript/announcing-typescript-3-8/#type-only-imports-exports)?

merlinnot: Would be great, I also struggled with it.

murgatroid99: That does look useful, though I am a little concerned about having an option that is essentially a TypeScript version compatibility switch. I would prefer to always output `export type`, but I am also not sure if it makes sense to assume that the user always has TypeScript 3.8 or higher.

merlinnot: We don't really have any current users of the TypeScript generator, so at least that's not a breaking change 🙂 People who use gRPC with TypeScript and would be interested in using/migrating to TypeScript generator are probably already advanced enough to keep their dependencies up to date. Stable version of TypeScript 3.8 has been available for over half a year. Just my two cents 🙂

murgatroid99: I guess in the worst case we can add a compatibility option later if we need one. OK, I am willing to change the generator to generate `import type` and `export type`.

murgatroid99:
The latest changes, including generating `import type` in a few places, are now available in version `0.6.0-pre16`.

packages/proto-loader/bin/proto-loader-gen-types.ts
```diff
@@ -598,7 +598,7 @@ function generateLoadedDefinitionTypes(formatter: TextFormatter, namespace: Prot
 
 function generateRootFile(formatter: TextFormatter, root: Protobuf.Root, options: GeneratorOptions) {
   formatter.writeLine(`import * as grpc from '${options.grpcLib}';`);
```
SimenB, line 600:
```suggestion
  formatter.writeLine(`import type * as grpc from '${options.grpcLib}';`);

```

murgatroid99: Done, I also addressed the other instance of the same import line and fixed another bug. That's all now available in 0.6.0-pre17.

pcj:
`0.6.0-pre17` appears to be working for me, no typescript compiler or runtime errors I'm observing.

entropitor:
`0.6.0-pre17` seems to fail on https://github.com/hashicorp/terraform/blob/master/docs/plugin-protocol/tfplugin5.2.proto I think it has to do with the fact that the message name and method name are the same

Small reproducable example
```proto
syntax = "proto3";

package tfplugin5;

// Schema is the configuration schema for a Resource, Provider, or Provisioner.
message Schema {
    // The version of the schema.
    // Schemas are versioned, so that providers can upgrade a saved resource
    // state when the schema is changed. 
    int64 version = 1;
}

service Provider {
    //////// Information about what a provider supports/expects
    rpc GetSchema(GetProviderSchema.Request) returns (GetProviderSchema.Response);
}

message GetProviderSchema {
    message Request {
    }
    message Response {
        Schema provider = 1;
        map<string, Schema> resource_schemas = 2;
        map<string, Schema> data_source_schemas = 3;
        Schema provider_meta = 5;
    }
}
```

entropitor:
For the same file (https://github.com/hashicorp/terraform/blob/master/docs/plugin-protocol/tfplugin5.2.proto#L155), it looks like it's also failing to compile `map<string, Schema> resource_schemas = 2;;` correctly, it translates it to `'resourceSchemas'?: (_tfplugin5_Schema);` instead of `'resourceSchemas'?: {[key: string]: (_tfplugin5_Schema)};`

murgatroid99:
@entropitor Regarding your first comment, I do not entirely follow what you are saying the problem is. You say that you think the cause is that the method and message names are the same, but in your example no two things have the same name so I don't see how it would demonstrate that specific problem.

Regarding your second comment, I thought I fixed that map bug already, but I'll take another look. Are you seeing that generation error for all map fields, or just some of them?

entropitor:
Ah @murgatroid99, for the part I highlighted, it's not the same 🙈 (for most other methods in the same linked proto file, they are), so I guess it's the fact that the message is within another message that is the problem then 😅

murgatroid99:
I just tried generating the output for your example proto file and the problem is not obvious to me. Can you be more specific about what is failing exactly?

I also tried generating the larger file you linked and the specific map field you pointed to appears to be generating correctly for me.

What command line options did you use, exactly?

entropitor:
I'm using `proto-loader-gen-types -O src/generated --grpcLib @grpc/grpc-js --keepCase --longs Number src/proto/*.proto`. it's generating everything without an error, but actually compiling the generated file with typescript throws type errors because the nested messages are not generated. See

```typescript
export namespace messages {
  export namespace tfplugin5 {
    export type ApplyResourceChange = _tfplugin5_ApplyResourceChange;
    export type ApplyResourceChange__Output = _tfplugin5_ApplyResourceChange__Output;
    export type AttributePath = _tfplugin5_AttributePath;
    export type AttributePath__Output = _tfplugin5_AttributePath__Output;
    export type Configure = _tfplugin5_Configure;
    export type Configure__Output = _tfplugin5_Configure__Output;
    // NOTICE HERE NO namepsace Configure which includes a Request and Response type!!!
    export type Diagnostic = _tfplugin5_Diagnostic;
    export type Diagnostic__Output = _tfplugin5_Diagnostic__Output;
    export type DynamicValue = _tfplugin5_DynamicValue;
    export type DynamicValue__Output = _tfplugin5_DynamicValue__Output;
    export type GetProviderSchema = _tfplugin5_GetProviderSchema;
    export type GetProviderSchema__Output = _tfplugin5_GetProviderSchema__Output;
    export type GetProvisionerSchema = _tfplugin5_GetProvisionerSchema;
    export type GetProvisionerSchema__Output = _tfplugin5_GetProvisionerSchema__Output;
    export type ImportResourceState = _tfplugin5_ImportResourceState;
    export type ImportResourceState__Output = _tfplugin5_ImportResourceState__Output;
    export type PlanResourceChange = _tfplugin5_PlanResourceChange;
    export type PlanResourceChange__Output = _tfplugin5_PlanResourceChange__Output;
    export type PrepareProviderConfig = _tfplugin5_PrepareProviderConfig;
    export type PrepareProviderConfig__Output = _tfplugin5_PrepareProviderConfig__Output;
    export namespace Provider {
    }
    export type ProvisionResource = _tfplugin5_ProvisionResource;
    export type ProvisionResource__Output = _tfplugin5_ProvisionResource__Output;
    export namespace Provisioner {
    }
    export type RawState = _tfplugin5_RawState;
    export type RawState__Output = _tfplugin5_RawState__Output;
    export type ReadDataSource = _tfplugin5_ReadDataSource;
    export type ReadDataSource__Output = _tfplugin5_ReadDataSource__Output;
    export type ReadResource = _tfplugin5_ReadResource;
    export type ReadResource__Output = _tfplugin5_ReadResource__Output;
    export type Schema = _tfplugin5_Schema;
    export type Schema__Output = _tfplugin5_Schema__Output;
    export type Stop = _tfplugin5_Stop;
    export type Stop__Output = _tfplugin5_Stop__Output;
    export type StringKind = _tfplugin5_StringKind;
    export type UpgradeResourceState = _tfplugin5_UpgradeResourceState;
    export type UpgradeResourceState__Output = _tfplugin5_UpgradeResourceState__Output;
    export type ValidateDataSourceConfig = _tfplugin5_ValidateDataSourceConfig;
    export type ValidateDataSourceConfig__Output = _tfplugin5_ValidateDataSourceConfig__Output;
    export type ValidateProvisionerConfig = _tfplugin5_ValidateProvisionerConfig;
    export type ValidateProvisionerConfig__Output = _tfplugin5_ValidateProvisionerConfig__Output;
    export type ValidateResourceTypeConfig = _tfplugin5_ValidateResourceTypeConfig;
    export type ValidateResourceTypeConfig__Output = _tfplugin5_ValidateResourceTypeConfig__Output;
  }
}
```
and
```typescript
export namespace ServiceHandlers {
  export namespace tfplugin5 {
    export namespace ApplyResourceChange {
      export namespace Request {
      }
      export namespace Response {
      }
    }
    export namespace AttributePath {
      export namespace Step {
      }
    }
    export namespace Configure {
      export namespace Request {
      }
      export namespace Response {
      }
    }
    export namespace Diagnostic {
    }
    export namespace DynamicValue {
    }
    export namespace GetProviderSchema {
      export namespace Request {
      }
      export namespace Response {
      }
    }
    export namespace GetProvisionerSchema {
      export namespace Request {
      }
      export namespace Response {
      }
    }
    export namespace ImportResourceState {
      export namespace ImportedResource {
      }
      export namespace Request {
      }
      export namespace Response {
      }
    }
    export namespace PlanResourceChange {
      export namespace Request {
      }
      export namespace Response {
      }
    }
    export namespace PrepareProviderConfig {
      export namespace Request {
      }
      export namespace Response {
      }
    }
    export interface Provider {
      ApplyResourceChange(call: grpc.ServerUnaryCall<messages.tfplugin5.ApplyResourceChange.Request__Output, messages.tfplugin5.ApplyResourceChange.Response>, callback: grpc.sendUnaryData<messages.tfplugin5.ApplyResourceChange.Response>): void;
      
      Configure(call: grpc.ServerUnaryCall<messages.tfplugin5.Configure.Request__Output, messages.tfplugin5.Configure.Response>, callback: grpc.sendUnaryData<messages.tfplugin5.Configure.Response>): void;
      
      GetSchema(call: grpc.ServerUnaryCall<messages.tfplugin5.GetProviderSchema.Request__Output, messages.tfplugin5.GetProviderSchema.Response>, callback: grpc.sendUnaryData<messages.tfplugin5.GetProviderSchema.Response>): void;
      
      ImportResourceState(call: grpc.ServerUnaryCall<messages.tfplugin5.ImportResourceState.Request__Output, messages.tfplugin5.ImportResourceState.Response>, callback: grpc.sendUnaryData<messages.tfplugin5.ImportResourceState.Response>): void;
      
      PlanResourceChange(call: grpc.ServerUnaryCall<messages.tfplugin5.PlanResourceChange.Request__Output, messages.tfplugin5.PlanResourceChange.Response>, callback: grpc.sendUnaryData<messages.tfplugin5.PlanResourceChange.Response>): void;
      
      PrepareProviderConfig(call: grpc.ServerUnaryCall<messages.tfplugin5.PrepareProviderConfig.Request__Output, messages.tfplugin5.PrepareProviderConfig.Response>, callback: grpc.sendUnaryData<messages.tfplugin5.PrepareProviderConfig.Response>): void;
      
      ReadDataSource(call: grpc.ServerUnaryCall<messages.tfplugin5.ReadDataSource.Request__Output, messages.tfplugin5.ReadDataSource.Response>, callback: grpc.sendUnaryData<messages.tfplugin5.ReadDataSource.Response>): void;
      
      ReadResource(call: grpc.ServerUnaryCall<messages.tfplugin5.ReadResource.Request__Output, messages.tfplugin5.ReadResource.Response>, callback: grpc.sendUnaryData<messages.tfplugin5.ReadResource.Response>): void;
      
      Stop(call: grpc.ServerUnaryCall<messages.tfplugin5.Stop.Request__Output, messages.tfplugin5.Stop.Response>, callback: grpc.sendUnaryData<messages.tfplugin5.Stop.Response>): void;
      
      UpgradeResourceState(call: grpc.ServerUnaryCall<messages.tfplugin5.UpgradeResourceState.Request__Output, messages.tfplugin5.UpgradeResourceState.Response>, callback: grpc.sendUnaryData<messages.tfplugin5.UpgradeResourceState.Response>): void;
      
      ValidateDataSourceConfig(call: grpc.ServerUnaryCall<messages.tfplugin5.ValidateDataSourceConfig.Request__Output, messages.tfplugin5.ValidateDataSourceConfig.Response>, callback: grpc.sendUnaryData<messages.tfplugin5.ValidateDataSourceConfig.Response>): void;
      
      ValidateResourceTypeConfig(call: grpc.ServerUnaryCall<messages.tfplugin5.ValidateResourceTypeConfig.Request__Output, messages.tfplugin5.ValidateResourceTypeConfig.Response>, callback: grpc.sendUnaryData<messages.tfplugin5.ValidateResourceTypeConfig.Response>): void;
      
    }
    export namespace ProvisionResource {
      export namespace Request {
      }
      export namespace Response {
      }
    }
    export interface Provisioner {
      GetSchema(call: grpc.ServerUnaryCall<messages.tfplugin5.GetProvisionerSchema.Request__Output, messages.tfplugin5.GetProvisionerSchema.Response>, callback: grpc.sendUnaryData<messages.tfplugin5.GetProvisionerSchema.Response>): void;
      
      ProvisionResource(call: grpc.ServerWritableStream<messages.tfplugin5.ProvisionResource.Request__Output, messages.tfplugin5.ProvisionResource.Response>): void;
      
      Stop(call: grpc.ServerUnaryCall<messages.tfplugin5.Stop.Request__Output, messages.tfplugin5.Stop.Response>, callback: grpc.sendUnaryData<messages.tfplugin5.Stop.Response>): void;
      
      ValidateProvisionerConfig(call: grpc.ServerUnaryCall<messages.tfplugin5.ValidateProvisionerConfig.Request__Output, messages.tfplugin5.ValidateProvisionerConfig.Response>, callback: grpc.sendUnaryData<messages.tfplugin5.ValidateProvisionerConfig.Response>): void;
      
    }
    export namespace RawState {
    }
    export namespace ReadDataSource {
      export namespace Request {
      }
      export namespace Response {
      }
    }
    export namespace ReadResource {
      export namespace Request {
      }
      export namespace Response {
      }
    }
    export namespace Schema {
      export namespace Attribute {
      }
      export namespace Block {
      }
      export namespace NestedBlock {
      }
    }
    export namespace Stop {
      export namespace Request {
      }
      export namespace Response {
      }
    }
    export namespace UpgradeResourceState {
      export namespace Request {
      }
      export namespace Response {
      }
    }
    export namespace ValidateDataSourceConfig {
      export namespace Request {
      }
      export namespace Response {
      }
    }
    export namespace ValidateProvisionerConfig {
      export namespace Request {
      }
      export namespace Response {
      }
    }
    export namespace ValidateResourceTypeConfig {
      export namespace Request {
      }
      export namespace Response {
      }
    }
  }
}
```

Notice how it's refering to e.g. `messages.tfplugin5.Configure.Request__Output` which is not appearing in the messages namespace

murgatroid99:
That's not the output of version 0.6.0-pre17 of the generator. The latest version does not generate the `messages` namespace, and includes the fix for generating nested messages.

entropitor:
Oh wow, thanks @murgatroid99, I used a `^` and it picked up an older version 😱 From first looks, this seems to work

angaaron:
Working really well for us thanks! is there a expected release for 0.6.0?

murgatroid99:
I would like to publish this, but it has been stuck in a kind of code review limbo. Actually, if anyone reading this would like to do a code review, that might help expedite the process.

SyedAsimAliSE:
Hi, there ; is there a way i can use wild card expansions in windows 10 ```proto/*.proto``` for some reason i am unable to generate all the files in my proto folder but if i am giving it a single file it is working.

badsyntax:
> I would like to publish this, but it has been stuck in a kind of code review limbo. Actually, if anyone reading this would like to do a code review, that might help expedite the process.

If it helps at all, I've been [comparing the types](https://github.com/badsyntax/grpc-js-types) generated by this package with other type generators, and well as comparing the types across `grpc` and `@grpc/grpc-js` packages, and there appears to be type parity. So while i'm not personally confident in doing a code review, from my observations and testing I feel this package is stable enough for an initial release.

jcready:
Would it be reasonable to re-export the request/response types from the either the top-level generated files or from the nested service interface files?

As it exists right now I have something like

```ts
import { ProtoGrpcType } from '../../__generated__/some_service_api';
import type { MyServiceAPIClient } from '../../__generated__/some/deeply/nested/directory/v1beta1/MyServiceAPI';

const protoFilePath = '../../protos/some/deeply/nested/directory/v1beta1/some_service_api.proto';

let protoPackage = loader.loadSync(protoFilePath, {
  keepCase: false,
  longs: String,
  enums: String,
  defaults: true,
  oneofs: true
});

const proto = grpc.loadPackageDefinition(protoPackage) as unknown as ProtoGrpcType;
const { MyServiceAPI } = proto.some.deeply.nested.directory.v1beta1;
```

But if I wanted to get a reference to the request/response types for each service method there is no straightforward way to do so without having to manually import each pair of types from each generated method file. Something like

```ts
import type { GetSomeMethodRequest } from '../../__generated__/some/deeply/nested/directory/v1beta1/GetSomeMethod';
import type { GetSomeMethodResponse__Output as GetSomeMethodResponse } from '../../__generated__/some/deeply/nested/directory/v1beta1/GetSomeMethodResponse';
// repeat for every method
```

It would be incredibly helpful if I could somehow reference the generated request/response types without having to import from a ton of different generated files. An incredibly simple way would just be to re-export all the imported types. Something like:

```ts
import type * as grpc from '@grpc/grpc-js'
import type { GetSomeMethodRequest as _some_deeply_nested_directory_v1beta1_GetSomeMethodRequest, GetSomeMethodRequest__Output as _some_deeply_nested_directory_v1beta1_GetSomeMethodRequest__Output } from '../../../../../some/deeply/nested/directory/v1beta1/GetSomeMethodRequest';
import type { GetSomeMethodResponse as _some_deeply_nested_directory_v1beta1_GetSomeMethodResponse, GetSomeMethodResponse__Output as _some_deeply_nested_directory_v1beta1_GetSomeMethodResponse__Output } from '../../../../../some/deeply/nested/directory/v1beta1/GetSomeMethodResponse';

export { _some_deeply_nested_directory_v1beta1_GetSomeMethodRequest, _some_deeply_nested_directory_v1beta1_GetSomeMethodRequest__Output };
export { _some_deeply_nested_directory_v1beta1_GetSomeMethodResponse, _some_deeply_nested_directory_v1beta1_GetSomeMethodResponse__Output };

// even better

export interface SomeServiceAPIClientIO {
  GetSomeMethod: {
    Request: _some_deeply_nested_directory_v1beta1_GetSomeMethodRequest,
    Response: _some_deeply_nested_directory_v1beta1_GetSomeMethodResponse__Output
  }
}

export interface SomeServiceAPIHandlersIO {
  GetSomeMethod: {
    Request: _some_deeply_nested_directory_v1beta1_GetSomeMethodRequest__Output,
    Response: _some_deeply_nested_directory_v1beta1_GetSomeMethodResponse
  }
}
```

This would provide a much easier method of referencing the request/response types for service methods either with named imports (from single file):

```ts
import type {
  MyServiceAPIClient,
  _some_deeply_nested_directory_v1beta1_GetSomeMethodRequest as GetSomeMethodRequest,
  _some_deeply_nested_directory_v1beta1_GetSomeMethodResponse__Output as GetSomeMethodResponse
} from '../../__generated__/some/deeply/nested/directory/v1beta1/MyServiceAPI';
```

Or by importing the whole interface (under the `// even better` comment):

```ts
import type {
  MyServiceAPIClient,
  MyServiceAPIClientIO
} from '../../__generated__/some/deeply/nested/directory/v1beta1/MyServiceAPI';

type GetSomeMethodRequest = SomeServiceAPIClientIO['GetSomeMethod']['Request'];
type GetSomeMethodResponse = SomeServiceAPIClientIO['GetSomeMethod']['Response'];
```

murgatroid99:
I had something similar, but it re-exported everything from the top level file, and that added a lot of complexity so I removed it. Re-exporting just the request and response types along with the corresponding service is an interesting idea, but I am concerned about exporting the same type in multiple places with different names, and only doing that with some message types. In particular, that leads to a situation where you can reference the request or response type of a method using one naming convention (e.g. `ServiceName.MethodName.Request`), but if you want to reference a message type that describes one of the fields of those messages, you need to use a different naming convention (e.g. `MessageName` or `MessageName__Output`). 

My own experience with using the generated code has been that the VSCode automatic import generation removes a lot of the friction of the current setup. Of course, not everyone uses VSCode, but I expect that other IDEs have similar functionality. In addition, in some cases TypeScript's type inference can allow you to avoid explicitly referencing the types at all.

jcready:
Totally understand. I was able to convince TypeScript to provide this information to me. In case anyone else is interested this is how I extract the request/response types from a service client type:

```ts
type Callback<T> = (err?: grpc.ServiceError, res?: T) => void;
type ServiceClientIO<C> = C extends () => grpc.ClientDuplexStream<infer I, infer O>
  ? [I, O]
  : C extends (args: infer I) => grpc.ClientReadableStream<infer O>
  ? [I, O]
  : C extends (cb: Callback<infer O>) => grpc.ClientWritableStream<infer I>
  ? [I, O]
  : C extends (args: infer I, cb: Callback<infer O>) => grpc.ClientUnaryCall
  ? [I, O]
  : never;

type API = MyServiceAPIClient;

type Req = {[P in keyof API]: ServiceClientIO<API[P]>[0]};
type Res = {[P in keyof API]: ServiceClientIO<API[P]>[1]};
```

Then I can use it like so to get the request/response types

```ts
type GetSomeMethodReqest = Req['GetSomeMethod'];
type GetSomeMethodResponse = Res['GetSomeMethod'];
```

The nice part about using the `Req` and `Res` types is that you get type-ahead provided **AND** when you hover over the `'GetSomeMethod'` part of `Req['GetSomeMethod']` it will actually show the generated comments from the proto file (assuming you used the `--includeComments` flag).

murgatroid99:
I think you can replace your `Callback` type in that code with `grpc.requestCallback`.

jcready:
It looks like that interface should work, but for some reason changing out my `Callback<T>` type with `requestCallback<T>` results in all the `Req` and `Res` values being `never`. My TS knowledge is pretty limited and so I'm not clear as to why replacing it doesn't work.

IanEdington:
Hey @murgatroid99 Do you need support on this PR? I'd be willing to do a code review. I'm very familiar with code generation + typescript but not with protobuf js specifically.

We're using gRPC for a new node service and would much prefer using this over grpc-tools protobuf lib

murgatroid99:
@IanEdington I would appreciate it if you reviewed this code.

packages/proto-loader/README.md
```diff
@@ -51,3 +52,66 @@ const options = {
...
+
+The `proto-loader-gen-types` script distributed with this package can be used to generate TypeScript type information for the objects loaded at runtime. More information about how to use it can be found in [the *@grpc/proto-loader TypeScript Type Generator CLI Tool* proposal document](https://github.com/grpc/proposal/blob/master/L70-node-proto-loader-type-generator.md). The arguments mostly match the `load` function's options; the full usage information is as follows:
+
+```console
+proto-loader-gen-types.js [options] filenames...
+
+Options:
+  --help             Show help                                         [boolean]
+  --version          Show version number                               [boolean]
+  --keepCase         Preserve the case of field names                  [boolean]
```
SimenB, line 66: does this work? I've tried passing `--keepCase false, `--keepCase true` and `--no-keepCase` without it affecting output

(what I want is to _just_ generate the camel cased ones)

murgatroid99: It's just a flag. You pass the option to get that behavior, and omit it to get the other behavior.

SimenB: Right, but whatever value I pass as this flag the generated code is identical. If it doesn't change anything it should be removed, no?

murgatroid99: It doesn't have a value. You either pass it or you don't. So, if you want the original field name case from the proto files, you pass `--keepCase`. And if you want it to force camel case names, you don't pass that flag at all.

SimenB: `yargs` should allow you to pass `false` to negate boolean flags.

However, not passing it at all also makes no difference in the generated code.

murgatroid99: It works for me. I tried modifying the `generate-golden` script introduced in this PR to remove that argument, and the field names generated in camel case instead of the original casing, as expected. I want to note that this only affects field names that are written in a format other than camel case, such as snake case with underscores. It won't affect message or service names, or field names that are already just one word.

SimenB: Aha, gotcha! Yeah, that's it then, I was looking at message names, all field names are already came case in our case 👍

merlinnot:
@murgatroid99 I found a potential issue while working on a service with multiple versions, or maybe I'm simply doing something wrong, but I looked at the source code and I have an impression that it might be a bug.

Given these two proto files:

```protobuf
syntax = "proto3";

package test.name.v1alpha1;

message Request {}

message Response {}

service MyService { rpc GetResponse(Request) returns (Response); }
```

```protobuf
syntax = "proto3";

package test.name.v1alpha2;

message Request {}

message Response {}

service MyService { rpc GetResponse(Request) returns (Response); }
```

(note the version change)

And the following command:
```bash
proto-loader-gen-types \
  --bytes array \
  --defaults false \
  --grpcLib @grpc/grpc-js \
  --includeComments \
  --includeDirs ./protos \
  --keepCase false \
  --longs number \
  --oneofs false \
  --outDir ./src/protos \
  test/name/v1alpha1/my_service.proto \
  test/name/v1alpha2/my_service.proto
```

The root `ProtoGrpcType` looks like this:
```
export interface ProtoGrpcType {
  test: {
    name: {
      v1alpha2: {
        Request: MessageTypeDefinition
        Response: MessageTypeDefinition
        MyService: SubtypeConstructor<typeof grpc.Client, _test_name_v1alpha2_MyServiceClient> & { service: ServiceDefinition }
      }
    }
  }
}
```

There's not `v1alpha1`. Additionally, changing the order of files in the command actually changes the output - seems like the last definition wins.

murgatroid99:
@merlinnot The code generator outputs files into the output directory with a name based on the base filename of the input files. It is not set up to handle multiple input files with the same base filename, as you have there. I'm not sure what would be the right way to handle that, but you can easily work around the problem by running two separate generator commands with different output directories.

bbeausej:
HI @murgatroid99 , our team has been using this MR/pre-release on services we are currently building. TL;DR it's been great and I think this MR adds tremendous value.

Here's a couple of feedback points and things to possible consider (or ignore, that's fine too!) :

**Index / Barrels**

barrels / index files would be extremely useful to simply type imports and avoid multiple import lines. What's the intent in generating a ts file per message instead of a file per package?

```
import {
  EchoServiceHandlers,
  EchoRequest,
  EchoResponse,
  EchoClientStreamRequest,
  EchoClientStreamResponse,
  EchoServerStreamRequest,
  EchoServerStreamResponse,
  EchoDuplexStreamRequest,
  EchoDuplexStreamResponse,
  CloseAllStreamsRequest,
  CloseAllStreamsResponse,
} from 'gen-types/mypackage/services/echo/v1';
```

instead of

```
import { EchoServiceHandlers } from 'gen-types/mypackage/services/echo/v1/EchoService';
import { EchoRequest } from 'gen-types/mypackage/services/echo/v1/EchoRequest';
import { EchoResponse } from 'gen-types/mypackage/services/echo/v1/EchoRequest';
import { EchoClientStreamRequest } from 'gen-types/mypackage/services/echo/v1/EchoClientStreamRequest';
import { EchoClientStreamResponse } from 'gen-types/mypackage/services/echo/v1/EchoClientStreamResponse';
import { EchoServerStreamRequest } from 'gen-types/mypackage/services/echo/v1/EchoServerStreamRequest';
import { EchoServerStreamResponse } from 'gen-types/mypackage/services/echo/v1/EchoServerStreamResponse';
import { EchoDuplexStreamRequest } from 'gen-types/mypackage/services/echo/v1/EchoDuplexStreamResponse';
import { EchoDuplexStreamResponse } from 'gen-types/mypackage/services/echo/v1/EchoDuplexStreamResponse';
import { CloseAllStreamsRequest } from 'gen-types/mypackage/services/echo/v1/CloseAllStreamsRequest';
import { CloseAllStreamsResponse } from 'gen-types/mypackage/services/echo/v1/CloseAllStreamsResponse';
```

We currently use [barrelsby](https://github.com/bencoveney/barrelsby#readme) as a build step to make those index files, a process that can be clunky and error prone , especially if we want to make a top level index files, where now you expose yourself to name clashes between your own protos and vendored ones like the google common apis. Possibly this workaround is enough.

**Helpful loader typing**

The grpc-loader types themselves could potentially benefit from small utilities to perform the service typecast for the loaded package. We ended up doing this to load our service protos to make it more dev friendly:

```
export const serviceLoader = <ServiceType>(
  entryProtoFile: string | string[],
  options: protoLoader.Options = {},
): ServiceType => {
  return (grpc.loadPackageDefinition(
    protoLoader.loadSync(entryProtoFile, options),
  ) as unknown) as ServiceType;
};
```

**Access to the pbjs Root**

Currently when using the grpc-loader, the loaded pbjs Root object is not easily retrievable. This could be useful in cases where you need the pbjs definitions for message in the root for things like `.verify(...)` or `.create(...)` or special cases of serialization. We ended up writing an equivalent loader using protobuf js that mimics the include path resolution of the protoLoader for the case where we needed it.

**\*\_\_Output types**

Wasn't very clear to team members what the purpose/use of the \_\_Output types was by default. I'm not certain `__Output` is the proper naming to make this clear for users of the generated types without requiring an explanation.

**EnumTypeDefinition included even when unused**

In the cases where you compile your Typescript with `noUnusedLocals` and your proto file does not use any enumerations, the generated types will not properly compile because the imported
`EnumTypeDefinition` local is unused. It unfortunately doesn't seem possible to turn off those compile option only for a subpath of the project forcing you to turn off this check. It's arguable this should be taken care of by linting tools (like eslint) instead of the compiler.


Thanks for the great pr!

-b

merlinnot:
> The code generator outputs files into the output directory with a name based on the base filename of the input files.

It seems like most of the files is named after a package name? All request/response/service interfaces seem to behave like this.

> I'm not sure what would be the right way to handle that,

It seems to me that we should simply merge these objects instead of overwriting, since we already have objects structured after package names, this would work great:

```ts
export interface ProtoGrpcType {
  test: {
    name: {
      v1alpha1: {
        Request: MessageTypeDefinition
        Response: MessageTypeDefinition
        MyService: SubtypeConstructor<typeof grpc.Client, _test_name_v1alpha1_MyServiceClient> & { service: ServiceDefinition }
      }
      v1alpha2: {
        Request: MessageTypeDefinition
        Response: MessageTypeDefinition
        MyService: SubtypeConstructor<typeof grpc.Client, _test_name_v1alpha2_MyServiceClient> & { service: ServiceDefinition }
      }
    }
  }
}
```

Even the imports (`_test_name_v1alpha2_MyServiceClient`) are already versioned, so there would be no conflicts here as well.

> but you can easily work around the problem by running two separate generator commands with different output directories.

That's what I'm currently doing, however it leads to a very unnatural structure of the files (not similar to any other gRPC generator).

To clarify further:
- All of the files other than top-level definitions (these which define `ProtoGrpcType`) are generated correctly.
- Imports in these top-level files are generated correctly (there would be no clashes).
- The `ProtoGrpcType` structure allows for inclusion of multiple versions naturally (it has properties named after versions, the same way as other files have it in paths).

The proposal is to merge `ProtoGrpcType` (as in object deep merge) so that it exposes all components.

This issue is easily reproducible with many Google libraries, which use more than one version (e.g. `v1beta1` and `v1`). It is not something specific to my setup, it's very much canonical, so I expect anyone who exposes more than one version of the service to run into this problem.

IanEdington (commented):
Hey sorry for saying I would do something then not doing it.
We've moved away from using the protobufjs loader in favour of the grpc proto tools

Here's a partial review from when I was looking at it earlier. Also based on this conversation it looks like some of these will be out of date.

packages/proto-loader/golden-generated/google/api/Http.ts
```diff
@@ -0,0 +1,49 @@
...
+   */
+  'fully_decode_reserved_expansion'?: (boolean);
+}
+
+/**
+ * Defines the HTTP configuration for an API service. It contains a list of
+ * [HttpRule][google.api.HttpRule], each specifying the mapping of an RPC method
+ * to one or more HTTP REST API methods.
+ */
+export interface Http__Output {
```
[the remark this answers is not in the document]
jcready: There is a difference. The fields in the `__Output` version aren't optional. I believe the reason is that protobufs have default values which don't get encoded in the buffer if the input value either isn't provided or matches the default value. Then when decoded the default values are provided. This is why input types are different from the `__Output` types. The normal/input types are what you need to provide to encode/serialize a message and the `__Output` version is what you would expect when decoding/deserializing a message.

packages/proto-loader/src/index.ts
```diff
@@ -15,12 +15,47 @@
...
-import * as path from 'path';
+
+import camelCase = require('lodash.camelcase');
 import * as Protobuf from 'protobufjs';
 import * as descriptor from 'protobufjs/ext/descriptor';
 
-import camelCase = require('lodash.camelcase');
+import { loadProtosWithOptionsSync, loadProtosWithOptions, Options, addCommonProtos } from './util';
+
+export { Long } from 'long';
```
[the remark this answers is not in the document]
murgatroid99: Some of the generated types use `Long`, and they need to access it without referring to `protobufjs`.

packages/proto-loader/README.md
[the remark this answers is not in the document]
murgatroid99: It should just be `false` for all of them. I copied this section of the README directly from the output of `proto-loader-gen-types --help`, which I just assumed would communicate the relevant information. I could change the yargs declaration section to declare the defaults explicitly if that would improve clarity.

packages/proto-loader/package.json
```diff
@@ -36,20 +38,29 @@
...
+  "bin": {
+    "proto-loader-gen-types": "./build/bin/proto-loader-gen-types.js"
+  },
   "dependencies": {
+    "@types/long": "^4.0.1",
     "lodash.camelcase": "^4.3.0",
-    "protobufjs": "^6.8.6"
+    "long": "^4.0.0",
+    "protobufjs": "^6.10.0",
+    "yargs": "^16.1.1"
```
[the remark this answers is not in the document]
murgatroid99: What would be the benefit of making these changes?

packages/proto-loader/package.json
```diff
@@ -23,7 +23,9 @@
     "check": "gts check",
     "fix": "gts fix",
     "pretest": "npm run compile",
-    "posttest": "npm run check"
+    "posttest": "npm run check",
+    "generate-golden": "node ./build/bin/proto-loader-gen-types.js --keepCase --longs=String --enums=String --defaults --oneofs --json --includeComments -I deps/gapic-showcase/schema/ deps/googleapis/ -O ./golden-generated --grpcLib @grpc/grpc-js google/showcase/v1beta1/echo.proto",
+    "validate-golden": "rm -rf ./golden-generated-old && mv ./golden-generated/ ./golden-generated-old && npm run generate-golden && diff -r ./golden-generated ./golden-generated-old"
```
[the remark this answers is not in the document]
murgatroid99: Part of the purpose of this test is to be able to examine a complex output of the generator, and to be able to see how changes to the generator affect the generated code. I agree that it's not great for testing in general, but with all of the people trying it and using it mostly successfully, I am fairly confident in the correctness, at least as far as generating compilable code is concerned.

packages/proto-loader/golden-generated/google/api/Http.ts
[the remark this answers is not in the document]
murgatroid99: That's right. The purpose of having two types is to give the user flexibility in what objects they can pass in as arguments, but give stronger type information about the objects that will be passed to them as output.

packages/proto-loader/package.json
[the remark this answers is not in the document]
IanEdington: It would make it easier to manage dependencies for the users of this library.

If a user has version 15 of yargs defined in another package "*" would use that version instead of forcing 2 versions of yargs to exist, potentially causing conflicts.

You almost definitely don't want two versions of `protobufjs` to exist in `node_modules` since they would generate slightly different `.js` / `.d.ts` files. Using a peer dependency would force this dependency to use whatever package was already defined in the package. If you have a 'peer' dependency it doesn't automatically install when doing `yarn install` or the npm equivalent. Adding a `protobufjs` `dev` dependency alongside the `peer` will make the local development experience smoother.

In general it's nice to work with libraries that are permissive about the packages they depend on, unless they need something specific for a version.

packages/proto-loader/README.md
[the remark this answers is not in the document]
IanEdington: As a user of libraries like this I would greatly appreciate that!

packages/proto-loader/golden-generated/google/api/Http.ts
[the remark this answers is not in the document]
IanEdington: ah thanks. didn't notice that 🙏

murgatroid99:
@bbeausej

### Index / Barrels
Many of the design choices in this PR, including that one, had the goal of minimizing the chances of any name conflicts, especially between files. If we wanted to add these files, we would need to choose a name that wouldn't conflict with anything else that is getting generated. For example, I think there are potential issues with packages that contain both concrete types and other packages. One other thing to consider is that types for messages or enums defined within other messages don't follow the same naming convention, but you can still access those types if you have to.

In my own experience, automatic import generation in an IDE like VSCode makes the scattering of types into different files much less of an issue.

### Helpful loader typing
A similar idea was suggested in a previous comment. I agree that this would be useful but I would prefer to not add any more complexity to this PR, so this change would belong in a future PR.

### Access to the pbjs Root
This is a strong non-goal of this library. The point is to provide an abstraction over Protobuf.js, so we do not want to expose any Protobuf.js APIs directly.

### *__Output types
If you have a suggestion for a better naming convention or a good way to explain them, please share. That's what I came up with and I don't have any better ideas. Comments in the generated code itself would be tricky because they would conflict with the generated comments from the `.proto` file.

### EnumTypeDefinition included even when unused
Doing it that way makes the code generation simpler, but I can change it if it's causing problems.

packages/proto-loader/README.md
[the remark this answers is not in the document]
murgatroid99: I also want to mention that the `longs` and `bytes` arguments are weird here, because the default behavior doesn't correspond to any allowed value, and can only be seen if the argument is omitted.

murgatroid99:
@merlinnot I think you are probably right that merging those files is the right solution, but that change is a lot trickier than it appears. First, the current code essentially runs the entire code generation algorithm separately for each input file. There would need to be changes at multiple levels to accommodate merging those files. And second, merging those objects is actually complex; those objects never actually exist in memory: it translates directly from Protobuf.js Namespace objects to generated code, and those Namespace objects wouldn't merge correctly.

merlinnot:
Thanks for the explanation. That's a shame we can't implement it easily, I guess I'll have to live with workarounds for now.

packages/proto-loader/package.json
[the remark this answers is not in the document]
murgatroid99: The problem is that I know that the version of yargs I am currently using works as I am using it, but I can't guarantee that every possible version of yargs that might ever be published will work the same way. And the NPM package structure should prevent version conflicts, so I don't see that as a relevant concern.

As for `protobufjs`, a peer dependency doesn't make sense. The purpose of a peer dependency is for when the user is expected to use that library directly, and the library with a peer dependency is a plugin or similar. But in this case, the proto-loader library is an abstraction layer over `protobufjs`, so it needs to use it directly, and it expects specific features from specific versions of `protobufjs`. I also don't see how different versions of `protobufjs` generating different output is a point in favor of making this change. Keeping the dependency as it is ensures that a specific version of `protobufjs` will be used when doing this code generation. Any other use of `protobufjs` outside of the context of this library is completely irrelevant to the function of this library.

merlinnot:
So happy to hear that, I'll test it today, tomorrow the latest!

merlinnot:
Ok, it took way less time than I expected. I tested it with moderately complex services (multiple overlapping and non-overlapping services and RPCs) and it worked out of the box. Many thanks, my codebase looks much, much more tidy now!

nicolasnoble (approved):

murgatroid99:
I have published this PR in version 0.6.0 of the `@grpc/proto-loader` library. Thank you everyone with your help and feedback for this feature.

Issue #1362, opened 2020-04-15
grpc-tools install fails roughly 50% of the time with "bad download"


michaelbeaumont:
### Problem description
When installing, `node-pre-gyp` throws a `bad download` error.

### Reproduction steps
`git clone https://github.com/grpc/grpc-node && cd grpc-node/packages/grpc-tools`
`npm i` until it fails

Dockerfile:
```
FROM node:13.13.0
RUN npm install --unsafe-perm -g grpc-tools
```

### Environment
 - Linux Arch amd64
 - 13.12 & 13.13
 - nvm
 - grpc-tools@1.8.1

but we're also seeing this on Google Cloud Build and Github Actions

### Additional context

```
$ npm install
> grpc-tools@1.8.1 install /home/mike/projects/gastromatic/grpc-node/packages/grpc-tools
> node-pre-gyp install

node-pre-gyp WARN Using needle for node-pre-gyp https download 
node-pre-gyp ERR! install error 
node-pre-gyp ERR! stack Error: bad download
node-pre-gyp ERR! stack     at afterExtract (/home/mike/projects/gastromatic/grpc-node/packages/grpc-tools/node_modules/node-pre-gyp/lib/install.js:98:46)
node-pre-gyp ERR! stack     at Unpack.afterTarball (/home/mike/projects/gastromatic/grpc-node/packages/grpc-tools/node_modules/node-pre-gyp/lib/install.js:178:9)
node-pre-gyp ERR! stack     at Unpack.emit (events.js:315:20)
node-pre-gyp ERR! stack     at Unpack.[maybeClose] (/home/mike/projects/gastromatic/grpc-node/packages/grpc-tools/node_modules/tar/lib/unpack.js:171:12)
node-pre-gyp ERR! stack     at Unpack.[unpend] (/home/mike/projects/gastromatic/grpc-node/packages/grpc-tools/node_modules/tar/lib/unpack.js:394:21)
node-pre-gyp ERR! stack     at /home/mike/projects/gastromatic/grpc-node/packages/grpc-tools/node_modules/tar/lib/unpack.js:308:46
node-pre-gyp ERR! stack     at FSReqCallback.oncomplete (fs.js:159:23)
node-pre-gyp ERR! System Linux 5.6.4-arch1-1
node-pre-gyp ERR! command "/usr/bin/node" "/home/mike/projects/gastromatic/grpc-node/packages/grpc-tools/node_modules/.bin/node-pre-gyp" "install"
[...]

PR #1476, opened 2020-06-15
Update node-pre-gyp to pickup fix for #1362

bsyk:
node-pre-gyp 0.12.0 uses needle 2.4.1 which has the bug in it.
Even with grpc 1.24.3, which refers to the updated version, it seems npm can decide to use the older version referenced by this package.

### DIFF
--- a/packages/grpc-tools/package.json
+++ b/packages/grpc-tools/package.json
@@ -24,7 +24,7 @@
     "prepublishOnly": "git submodule update --init --recursive && node copy_well_known_protos.js"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


murgatroid99:
I just checked, and `node-pre-gyp` 0.12.x definitely picks up `needle` 2.5.0, including specifically when installing `grpc-tools`. Do you have a `package-lock.json` that is locking `needle` to an earlier version?

bsyk:
Yeah, sorry. Somehow overlooked that.

murgatroid99:
To be fair, we might as well update the dependency for the next time we do a release. I think it's just not urgent.

PR #1479, opened 2020-06-16
grpc-js: Don't initiate a read after receiving a message

### LINKED ISSUE #1473: @grpc/grpc-js: Data loss on server-side streaming call (and a possible solution)
### Problem description
(This bug applies to @grpc/grpc-js.  C-based npm grpc does not have this problem.)
If a server-side streaming call that produces a lot of messages is pipelined with a slow write stream (for example, file write stream), a few of last messages can be lost.

### Reproduction steps
1. Unzip attached `TestGrpc.zip`.
2. run `npm install`
3. run `node TestGrpcServer.js` on one command window.
4. run `node TestGrpcClient.js` on the other command window.
5. On the client command window, `Total written lines: 100000` should be shown. If the number is lower than 100000, it's the bug this issue describes.

And here is the location and the cause of the bug I found:

The source file that has the bug is call-stream.ts.  
The buggy functions are `maybeOutputStatus()` (line 268), `push()` (line 285) and `startRead()` (line 635).  
The problem is, when a `push()` waits on `process.nextTick()`, other `push()`s are additionally called and call `process.nextTick()`, thus build a 'push queue' on the Node.js event queue. And when one of the `push()`s causes `this.outputStatus()` while others are on the 'push queue', the other `push()`s are skipped since `this.statusOutput` is already set (by previous `push()`). Though `push()` has `this.canPush` guard, it is set to true by `startRead()` which calls `this.push()` when `this.unpushedReadMessages.length > 0`.

And here is a possible solution to fix the bug:

I modified `call-stream.js` to fix the bug. (Currently I have no npm development environment, so I just modified the JavaScript file instead of the TypeScript file. The modified `call-stream.js` file is included in the attached `TestGrpc.zip` file. Search for `IWFIX` string in the file. It's the locations I modified. Most importantly, I introduced a member property named `numNextTickPushes` to guard 'queued' pushes. I don't know whether it's the best fix since I don't know much of the source code. But anyway it seems to fix the bug.

### Environment
 - OS: Windows 10 x64 (real machine), Linux CentOS 7.7 x64 (VirtualBox)
 - Node version: 12.18.0 (Windows 10, CentOS), 10.21.0 (Windows 10)
 - Node installation method: nvm (Windows 10), download & tar xvfz (CentOS)
 - If applicable, compiler version: not applicable
 - Package name and version: @grpc/grpc-js 1.0.5

### Additional context
Attachment: [TestGrpc.zip](https://github.com/grpc/grpc-node/files/4769104/TestGrpc.zip)

murgatroid99:
It turns out that the `Readable` stream class will aggressively call `_read` even if `push` returns `true`, so we don't also need to start another read after pushing. This fixes #1473, at least according to the test case provided in that issue.

### DIFF
--- a/test/api/interop_extra_test.js
+++ b/test/api/interop_extra_test.js
@@ -158,12 +158,13 @@ describe(`${anyGrpc.clientName} client -> ${anyGrpc.serverName} server`, functio
       call.on('data', (value) => {
         responseCount++;
       });
-      call.on('end', () => {
+      call.on('status', (status) => {
+        assert.strictEqual(status.code, grpc.status.OK);
         assert.strictEqual(responseCount, arg.response_parameters.length);
         done();
       });
-      call.on('status', function(status) {
-        assert.strictEqual(status.code, grpc.status.OK);
+      call.on('error', (error) => {
+        assert.ifError(error);
       });
     });
     describe('max message size', function() {
--- a/test/api/interop_extra_test.js
+++ b/test/api/interop_extra_test.js
@@ -144,6 +144,28 @@ describe(`${anyGrpc.clientName} client -> ${anyGrpc.serverName} server`, functio
         done();
       });
     });
+    it('should receive all messages in a long stream', function() {
+      var arg = {
+        response_type: 'COMPRESSABLE',
+        response_parameters: [
+        ]
+      };
+      for (let i = 0; i < 100_000; i++) {
+        arg.response_parameters.push({size: 0});
+      }
+      var call = client.streamingOutputCall(arg);
+      let responseCount = 0;
+      call.on('data', (value) => {
+        responseCount++;
+      });
+      call.on('end', () => {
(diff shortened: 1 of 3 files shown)


test/api/interop_extra_test.js
```diff
@@ -51,7 +51,7 @@ function echoMetadataGenerator(options, callback) {
 
 const credentials = grpc.credentials.createFromMetadataGenerator(echoMetadataGenerator);
 
-describe(`${anyGrpc.clientName} client -> ${anyGrpc.serverName} server`, function() {
+describe.only(`${anyGrpc.clientName} client -> ${anyGrpc.serverName} server`, function() {
```
badsyntax, line 54: Is this `only` intentional?

murgatroid99: No it is not. I left that in there from when I was fixing the test. Thanks for catching it.

PR #1494, opened 2020-07-07
grpc-js: Improve error handling in a few places

murgatroid99:
There are a few different error handling-related changes here:

 1. Update the error message "Received RST_STREAM with code 2" to add the string "Internal server error". We have seen several reports of people seeing the original error message and not understanding it.

 2. Add trace logs for GOAWAYs. Node's http2 internals seem to automatically translate GOAWAYs to RST_STREAMs on streams still open in that session. Logging this will help distinguish between real RST_STREAM errors and GOAWAYs.

 3. Transition a subchannel to TRANSIENT_FAILURE whenever its session fails to create a stream. This addresses a failure reported at https://github.com/googleapis/nodejs-firestore/issues/1023#issuecomment-653204096 where an error is thrown from `session.request` with the code `ERR_HTTP2_GOAWAY_SESSION`, but we never see a 'goaway' or 'close' event, so we just keep using that session forever and it never works.

I also bumped the version to 1.1.2 because I would like to publish that third change.

### DIFF
--- a/packages/grpc-js/src/subchannel.ts
+++ b/packages/grpc-js/src/subchannel.ts
@@ -309,7 +309,8 @@ export class Subchannel {
         };
         connectionOptions.servername = sslTargetNameOverride;
       } else {
-        const authorityHostname = splitHostPort(targetAuthority)?.host ?? 'localhost';
+        const authorityHostname =
+          splitHostPort(targetAuthority)?.host ?? 'localhost';
         // We want to always set servername to support SNI
         connectionOptions.servername = authorityHostname;
       }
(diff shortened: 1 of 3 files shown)


packages/grpc-js/src/subchannel.ts
```diff
@@ -309,7 +309,8 @@ export class Subchannel {
         };
         connectionOptions.servername = sslTargetNameOverride;
       } else {
-        const authorityHostname = splitHostPort(targetAuthority)?.host ?? 'localhost';
+        const authorityHostname =
+          splitHostPort(targetAuthority)?.host ?? 'localhost';
```
murgatroid99, line 313: This is just an incidental formatting change that came from formatting the rest of the file.

PR #1500, opened 2020-07-15
Fix interop test breakage

stanley-cheung:
Fixes https://github.com/grpc/grpc/issues/23466

To reproduce locally:

```
$ git clone https://github.com/grpc/grpc-node
$ cd grpc-node
$ git submodule update --init --recursive 
$ npm install gulp
$ npm install
$ cd packages/proto-loader
$ npm install
```

Got
```
> protobufjs@6.10.0 postinstall /usr/local/google/home/stanleycheung/grpc-node-test2/packages/proto-loader/node_modules/protobufjs
> node scripts/postinstall

> @grpc/proto-loader@0.5.5 prepare /usr/local/google/home/stanleycheung/grpc-node-test2/packages/proto-loader
> npm run compile

> @grpc/proto-loader@0.5.5 compile /usr/local/google/home/stanleycheung/grpc-node-test2/packages/proto-loader
> tsc -p .

../../node_modules/@types/gulp-mocha/index.d.ts:10:39 - error TS2304: Cannot find name 'MochaSetupOptions'.

10 declare function mocha(setupOptions?: MochaSetupOptions): NodeJS.ReadWriteStream;
                                         ~~~~~~~~~~~~~~~~~
Found 2 errors.
```

The fix is basically copying this line: https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/package.json#L22

PR #1502, opened 2020-07-15
Update @types/gulp-mocha dependency

murgatroid99:
This fixes a breakage caused by an `@types/mocha` update.


murgatroid99:
OK, the change in here caused other problems that I didn't anticipate, so I made #1504 with the same change as in #1500.

PR #1507, opened 2020-07-17
grpc-js: Implement EDS load balancer

murgatroid99:
This is implemented on top of #1489. The changes for this PR are the last 2 commits, and they affect the files `load-balancer-eds.ts` and `load-balancing-config.ts`.

### DIFF
--- a/packages/grpc-js/src/load-balancer-eds.ts
+++ b/packages/grpc-js/src/load-balancer-eds.ts
@@ -15,18 +15,31 @@
  *
  */
 
-import { LoadBalancer, ChannelControlHelper, registerLoadBalancerType, getFirstUsableConfig } from "./load-balancer";
-import { SubchannelAddress } from "./subchannel";
-import { LoadBalancingConfig, isEdsLoadBalancingConfig, EdsLoadBalancingConfig, PriorityLbConfig, PriorityChild, WeightedTarget, PriorityLoadBalancingConfig } from "./load-balancing-config";
-import { ChildLoadBalancerHandler } from "./load-balancer-child-handler";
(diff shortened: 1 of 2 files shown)


packages/grpc-js/src/load-balancer-eds.ts
```diff
@@ -0,0 +1,313 @@
...
+    }
+    if (!(attributes.xdsClient instanceof XdsClient)) {
+      return;
+    }
+    this.lastestConfig = lbConfig;
+    this.latestAttributes = attributes;
+    this.xdsClient = attributes.xdsClient;
+    this.edsServiceName = lbConfig.eds.edsServiceName ?? lbConfig.eds.cluster;
+
+    if (!this.isWatcherActive) {
```
stanley-cheung, line 287: As discussed offline, what if this function `updateAddressList` is called with a different `lbConfig.eds.edsServiceName`? After the first call, `this.isWatcherActive` is set to true. Will the 2nd call not get `this.xdsClient.addEndpointWatcher()` called?

Or maybe I have misunderstood the semantics - maybe `edsServiceName` won't change?

murgatroid99: I have changed this behavior to handle the name changing.

packages/grpc-js/src/load-balancer-eds.ts
```diff
@@ -0,0 +1,313 @@
...
+      this.xdsClient.addEndpointWatcher(this.edsServiceName, this.watcher);
+      this.isWatcherActive = true;
+    }
+
+    this.updateChild();
```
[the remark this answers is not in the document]
murgatroid99: I added a comment.

PR #1512, opened 2020-07-23
grpc-js: Fix handling of unsuccessful TXT record lookups

### LINKED ISSUE #1510: The round_robin loadbalancer is not working if DNS TXT record resolves failed with an serviceConfigError
When I use round_robin load balancer, if DNS TXT record resolves failed with an serviceConfigError then the service config I configured was override by the default pick_first policy config.
My grpc channel config is the following:
```
{
'grpc.service_config': JSON.stringify({
    loadBalancingPolicy: 'round_robin',
    loadBalancingConfig: [{ name: 'round_robin', round_robin: {} }],
  })
}
```
From the source:
https://github.com/grpc/grpc-node/blob/301b695d1552918629d50ee434a9e3269fac607d/packages/grpc-js/src/resolving-load-balancer.ts#L119
The first time the `serviceConfig` and `serviceConfigError` is null and the `this.previousServiceConfig` was set to null. The second time the `serviceConfigError` is not null because of failling TXT records resolving then the working service config was replaced by the `this.previousServiceConfig` which is null now so working config was fallback to pick_first config.
https://github.com/grpc/grpc-node/blob/301b695d1552918629d50ee434a9e3269fac607d/packages/grpc-js/src/resolving-load-balancer.ts#L140
should be replaced by `if (!this.defaultServiceConfig)` here IMO.

murgatroid99:
This should fix #1510 by changing how TXT record lookup failures are handled. Specifically, now no action is taken when that happens.

I also updated the config handling codepath to address updates that have been made to [the spec](https://github.com/grpc/proposal/blob/master/A21-service-config-error-handling.md) since this implementation was originally written. In addition to that, I simplified the logic somewhat to more consistently fall back to using the default service config when we do not retrieve one.

### DIFF
--- a/packages/grpc-js/src/resolver-dns.ts
+++ b/packages/grpc-js/src/resolver-dns.ts
@@ -237,19 +237,13 @@ class DnsResolver implements Resolver {
             }
           },
           (err) => {
-            this.latestServiceConfigError = {
-              code: Status.UNAVAILABLE,
-              details: 'TXT query failed',
-              metadata: new Metadata(),
-            };
-            if (this.latestLookupResult !== null) {
-              this.listener.onSuccessfulResolution(
-                this.latestLookupResult,
-                this.latestServiceConfig,
-                this.latestServiceConfigError,
-                {}
-              );
-            }
+            /* If TXT lookup fails we should do nothing, which means that we
+             * continue to use the result of the most recent successful lookup,
+             * or the default null config object if there has never been a
+             * successful lookup. We do not set the latestServiceConfigError
+             * here because that is specifically used for response validation
+             * errors. We still need to handle this error so that it does not
+             * bubble up as an unhandled promise rejection. */
           }
         );
       }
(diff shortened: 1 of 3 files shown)


packages/grpc-js/src/resolver-dns.ts
```diff
@@ -237,19 +237,13 @@ class DnsResolver implements Resolver {
...
-                {}
-              );
-            }
+            /* If TXT lookup fails we should do nothing, which means that we
+             * continue to use the result of the most recent successful lookup,
+             * or the default null config object if there has never been a
+             * successful lookup. We do not set the latestServiceConfigError
+             * here because that is specifically used for response validation
+             * errors. We still need to handle this error so that it does not
+             * bubble up as an unhandled promise rejection. */
```
stanley-cheung, line 246: Thanks for having this comment here. This makes a lot of sense and explains why we need an empty handler here. Thanks.

murgatroid99:
I changed the config handling part again.

stanley-cheung:
acknowledged.

PR #1513, opened 2020-07-23
grpc-js: CDS load balancer

murgatroid99:
This is the next step in the xDS implementation. The CDS balancer sits one level above the EDS balancer in the load balancer hierarchy.

### DIFF
--- a/packages/grpc-js/src/load-balancer-cds.ts
+++ b/packages/grpc-js/src/load-balancer-cds.ts
@@ -21,7 +21,12 @@ import {
   registerLoadBalancerType,
 } from './load-balancer';
 import { SubchannelAddress } from './subchannel';
-import { LoadBalancingConfig, isCdsLoadBalancingConfig, EdsLbConfig, CdsLoadBalancingConfig } from './load-balancing-config';
+import {
+  LoadBalancingConfig,
+  isCdsLoadBalancingConfig,
+  EdsLbConfig,
+  CdsLoadBalancingConfig,
+} from './load-balancing-config';
 import { XdsClient, Watcher } from './xds-client';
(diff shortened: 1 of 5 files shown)


packages/grpc-js/src/load-balancer-cds.ts
```diff
@@ -0,0 +1,162 @@
...
+              code: Status.UNAVAILABLE,
+              details: `xDS request failed with error ${status.details}`,
+              metadata: new Metadata(),
+            })
+          );
+        }
+      },
+    };
+  }
+
```
stanley-cheung, line 100: So there is no need for a `updateChild()` function here?

packages/grpc-js/src/load-balancer-cds.ts
```diff
@@ -0,0 +1,162 @@
...
+        if (update.lrs_server?.self) {
+          /* the lrs_server.self field indicates that the same server should be
+           * used for load reporting as for other xDS operations. Setting
+           * lrsLoadReportingServerName to the empty string sets that behavior.
+           * Otherwise, if the field is omitted, load reporting is disabled. */
+          edsConfig.lrsLoadReportingServerName = '';
+        }
+        this.childBalancer.updateAddressList(
+          [],
+          { name: 'eds', eds: edsConfig },
```
stanley-cheung, line 75: Ah, so this `edsConfig` is the end goal of this update. The logic is much simpler than `EdsLoadBalancer`. so there's no need for a separate `updateChild()` method.

packages/grpc-js/src/load-balancer-cds.ts
murgatroid99: Since the only thing in the CDS config is the name, a config change always invalidates the previous update, so that update would never do anything.

PR #1522, opened 2020-07-31
grpc-js: Add XdsResolver and corresponding XdsClient behavior

murgatroid99:
This is the final major PR in the implementation of [gRFC A27: xDS-Based Global Load Balancing](https://github.com/grpc/proposal/blob/master/A27-xds-global-load-balancing.md). This PR adds the `XdsResolver` class that handles `xds://address` target names using an `XdsClient` and emits service config objects that configure the load balancer classes created in the other classes.

This PR purposely does not call the `setup` function in `xds-resolver.ts`. That function call will enable the global load balancing feature; it will be added after the feature is sufficiently tested.

All of the new code in `src/generated` is there to handle deserializing the `HttpConnectionManager` message from a `google.protobuf.Any` field on lines 720 and 755 of `xds-client.ts`.

### DIFF
--- a/packages/grpc-js/src/channel.ts
+++ b/packages/grpc-js/src/channel.ts
@@ -34,7 +34,6 @@ import { CallCredentialsFilterFactory } from './call-credentials-filter';
 import { DeadlineFilterFactory } from './deadline-filter';
 import { CompressionFilterFactory } from './compression-filter';
 import { getDefaultAuthority, mapUriDefaultScheme } from './resolver';
-import { ServiceConfig, validateServiceConfig } from './service-config';
 import { trace, log } from './logging';
 import { SubchannelAddress } from './subchannel';
 import { MaxMessageSizeFilterFactory } from './max-message-size-filter';
--- a/packages/grpc-js/src/channel.ts
+++ b/packages/grpc-js/src/channel.ts
@@ -220,20 +220,10 @@ export class ChannelImplementation implements Channel {
         );
       },
     };
-    // TODO(murgatroid99): check channel arg for default service config
-    let defaultServiceConfig: ServiceConfig = {
-      loadBalancingConfig: [],
-      methodConfig: [],
-    };
-    if (options['grpc.service_config']) {
-      defaultServiceConfig = validateServiceConfig(
-        JSON.parse(options['grpc.service_config']!)
-      );
-    }
     this.resolvingLoadBalancer = new ResolvingLoadBalancer(
       this.target,
       channelControlHelper,
-      defaultServiceConfig
+      options
     );
     this.filterStackFactory = new FilterStackFactory([
       new CallCredentialsFilterFactory(this),
--- a/packages/grpc-js/src/xds-client.ts
+++ b/packages/grpc-js/src/xds-client.ts
@@ -452,6 +452,7 @@ export class XdsClient {
                 } else {
                   nackError = 'RouteConfiguration validation failed';
                 }
+                break;
               }
             } else {
               nackError = `Invalid resource type ${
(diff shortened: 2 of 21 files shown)


packages/grpc-js/src/xds-client.ts
```diff
@@ -308,97 +327,171 @@ export class XdsClient {
...
+      case RDS_TYPE_URL: {
+        let nackError: string | null = null;
+        if (this.routeConfigName === null) {
+          nackError = 'Unexpected RouteConfiguration response';
+        } else {
           for (const resource of message.resources) {
             if (
               protoLoader.isAnyExtension(resource) &&
-              resource['@type'] === CDS_TYPE_URL
+              resource['@type'] === LDS_TYPE_URL
```
stanley-cheung, line 442: Should this be `RDS_TYPE_URL`?

packages/grpc-js/src/xds-client.ts
```diff
@@ -543,6 +706,36 @@ export class XdsClient {
...
+        }
+        break;
+      case 'route_config':
+        return this.validateRdsResponse(httpConnectionManager.route_config!);
+    }
+    return false;
+  }
+
+  private validateRdsResponse(message: RouteConfiguration__Output): boolean {
+    return true;
```
stanley-cheung, line 736: Anything more to be done here in the future?

packages/grpc-js/src/xds-client.ts
```diff
@@ -581,16 +820,26 @@ export class XdsClient {
     }
   }
 
+  private updateRdsNames() {
+    this.adsCall?.write({
+      node: this.adsNode!,
+      type_url: CDS_TYPE_URL,
```
stanley-cheung, line 826: Perhaps you mean `RDS_TYPE_URL` here?

packages/grpc-js/src/xds-client.ts
```diff
@@ -581,16 +820,26 @@ export class XdsClient {
     }
   }
 
+  private updateRdsNames() {
+    this.adsCall?.write({
```
stanley-cheung, line 824: I noticed that some other occurrences above uses `if (this.adsCall) { this.adsCall.write({ ...`. So perhaps those can be replaced with `this.adsCall?.write({ ...`?

packages/grpc-js/src/channel.ts
```diff
@@ -220,20 +220,10 @@ export class ChannelImplementation implements Channel {
         );
       },
     };
-    // TODO(murgatroid99): check channel arg for default service config
-    let defaultServiceConfig: ServiceConfig = {
-      loadBalancingConfig: [],
-      methodConfig: [],
-    };
-    if (options['grpc.service_config']) {
-      defaultServiceConfig = validateServiceConfig(
```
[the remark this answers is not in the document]
murgatroid99: Done

packages/grpc-js/src/xds-client.ts
murgatroid99: Yes, fixed.

packages/grpc-js/src/xds-client.ts
murgatroid99: Yes, fixed.

packages/grpc-js/src/xds-client.ts
murgatroid99: I honestly don't know. But I want to have this here for structural consistency with the other types of responses, and so that I have a clear place to update if I do need validation logic.

packages/grpc-js/src/xds-client.ts
murgatroid99: I think that replacement would be good, but now that I have the whole picture of the `XdsClient` functionality, I want to do a separate PR to refactor the class more broadly. I can make that change there.

PR #1525, opened 2020-08-04
grpc-js: Add end(md?: Metadata) method to streaming server calls

### LINKED ISSUE #1520: grpc-js: ServerWritableStream.end(metadata?: Metadata) is missing from type definition
### Problem description
`ServerWritableStream.end(metadata?: Metadata)` is missing from type definition. This method is overriding the streams `end` method with different arguments. It works out if you don't include an argument because the stream has a variant like that but you can't send metadata without lying about the type.

### Reproduction steps
```typescript
const metadata = new Metadata();
stream.end(metadata as any);
```

### Environment
 - @grpc/grpc-js@1.1.1

murgatroid99:
This fixes #1520.

### DIFF
--- a/packages/grpc-js/src/server-call.ts
+++ b/packages/grpc-js/src/server-call.ts
@@ -91,10 +91,13 @@ export type ServerWritableStream<
   RequestType,
   ResponseType
 > = ServerSurfaceCall &
-  ObjectWritable<ResponseType> & { request: RequestType | null };
+  ObjectWritable<ResponseType> & {
+    request: RequestType | null;
+    end: (metadata?: Metadata) => void;
+  };
 export type ServerDuplexStream<RequestType, ResponseType> = ServerSurfaceCall &
   ObjectReadable<RequestType> &
-  ObjectWritable<ResponseType>;
+  ObjectWritable<ResponseType> & { end: (metadata?: Metadata) => void };
 
 export class ServerUnaryCallImpl<RequestType, ResponseType> extends EventEmitter
   implements ServerUnaryCall<RequestType, ResponseType> {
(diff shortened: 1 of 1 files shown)


packages/grpc-js/src/server-call.ts
```diff
@@ -255,6 +258,15 @@ export class ServerDuplexStreamImpl<RequestType, ResponseType> extends Duplex
   sendMetadata(responseMetadata: Metadata): void {
     this.call.sendMetadata(responseMetadata);
   }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  end(metadata?: any) {
```
thomassuckow, line 263: I know this is a copy from below but shouldn't this be `metadata?: Metadata` ?
`this.trailingMetadata` is of type Metadata so I can't imagine what other type it could be.

murgatroid99: The problem is that TypeScript doesn't like having a class that extends `Duplex` but implements `end` with a type that is incompatible with the `end` method declared in the `Duplex` class. Declaring the argument as `any` smooths that out.

There's a more complicated way of doing this that I use in [object-stream.ts](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/object-stream.ts) where I declare an intermediate type that uses `any & T` as the type, and then I can subtype that using `T`. I don't think that's valuable here though because this type is only in the implementation. The interface that actually gets provided to the user has the stricter `Metadata` type, and that's what matters here.

thomassuckow: @murgatroid99 The other day I came across a similar situation and was able to mitigate it with `Omit<Type, 'end'>`. Might be useful in the future.

PR #1534, opened 2020-08-12
Fix incorrectly named grpc-tools flag

mavaa:
Was  "--generate_package_definitions" (with an s) but should be "generate_package_definition" as in the documentation here: https://github.com/grpc/grpc-node/tree/master/packages/grpc-tools


mavaa:
Lost a few hours on this, hopefully it won't happen to anyone else :smile:

PR #1548, opened 2020-08-19
Prevent mutation of default headers

richardpringle:
Right now, the `defaultResponseHeaders` global is mutated on every request. This is a **huge** security vulnerability if there's any auth data in the metadata.


richardpringle:
@murgatroid99, any idea of quickly this will get released?

murgatroid99:
Conveniently, I was already working on a new release, so this is out in 1.1.4.

PR #1552, opened 2020-08-20
grpc-js: Enable the xds URL scheme and add an interop test for it

murgatroid99:
The interop test client is specified in [this document](https://github.com/grpc/grpc/blob/master/doc/xds-test-descriptions.md). The script for running it is modified from [this file in the grpc-go repository](https://github.com/grpc/grpc-go/blob/master/test/kokoro/xds.sh).

This also includes the generated typescript code for the interop client, and a couple of fixes for the xds code and some related code.


packages/grpc-js/package.json
```diff
@@ -57,7 +60,7 @@
   "dependencies": {
     "@grpc/proto-loader": "^0.6.0-pre14",
     "@types/node": "^12.12.47",
-    "google-auth-library": "^6.0.0",
+    "google-auth-library": "^5.10.1",
```
nicolasnoble, line 63: Are we downgrading this? Wasn't it bumped recently?

packages/grpc-js/scripts/xds.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
nicolasnoble, line 1: Copyright notice

packages/grpc-js/src/subchannel.ts
```diff
@@ -405,14 +411,14 @@ export class Subchannel {
...
-              LogVerbosity.ERROR,
-              `Connection to ${this.channelTarget} rejected by server because of excess pings`
-            );
             this.keepaliveTimeMs = Math.min(
               2 * this.keepaliveTimeMs,
               KEEPALIVE_MAX_TIME_MS
             );
+            logging.log(
+              LogVerbosity.ERROR,
+              `Connection to ${uriToString(this.channelTarget)} at ${this.subchannelAddressString} rejected by server because of excess pings. Increasing ping interval to ${this.keepaliveTimeMs} ms`
```
nicolasnoble, line 420: Could technically be useful to have a different message if we reached the max, but, not a big deal.

nicolasnoble (approved):

packages/grpc-js/package.json
murgatroid99: Having it at 6 was an error, because 6 requires Node 10 and above (#1442). I mistakenly bumped it up to 6 while moving it around between prod dependencies and peer dependencies and dev dependencies.

packages/grpc-js/scripts/xds.sh
murgatroid99: Fixed

packages/grpc-js/src/subchannel.ts
murgatroid99: If this ever actually comes up, then I'll consider changing it.

PR #1555, opened 2020-08-29
Fixed connectivity to Google PubSub over proxy

wkchee:
Latest version has caused @google-cloud/pubsub fail to connect over a proxy connection.

Snapshot of error:

2020-08-29T10:52:45.340Z | proxy | Successfully connected to pubsub.googleapis.c
om:443 through proxy 172.16.52.252:443
2020-08-29T10:52:45.370Z | subchannel | 172.16.52.252:443 CONNECTING -> TRANSIEN
T_FAILURE
2020-08-29T10:52:45.372Z | pick_first | CONNECTING -> TRANSIENT_FAILURE
2020-08-29T10:52:45.373Z | resolving_load_balancer | dns:172.16.52.252:443 CONNE
CTING -> TRANSIENT_FAILURE
2020-08-29T10:52:45.375Z | channel | Pick result: TRANSIENT_FAILURE subchannel:
undefined status: 14 No connection established
2020-08-29T10:52:45.377Z | call_stream | [11] cancelWithStatus code: 14 details:
 "No connection established"
2020-08-29T10:52:45.379Z | call_stream | [11] ended with status: code=14 details
="No connection established"
2020-08-29T10:52:45.381Z | connectivity_state | dns:172.16.52.252:443 CONNECTING
 -> TRANSIENT_FAILURE

Before proposed fix:

    static getDefaultAuthority(target) {
        return target.path;  // this returns "pubsub.googleapis.com:443"
    }

After proposed fix:

    static getDefaultAuthority(target) {
        const hostPort = uri_parser_1.splitHostPort(target.path);  // target.path is "pubsub.googleapis.com:443"
        if (hostPort !== null) {
            return hostPort.host; // this returns "pubsub.googleapis.com"
        }
        else {
            throw new Error(`Failed to parse target ${uri_parser_1.uriToString(target)}`);
        }
    }

### DIFF
--- a/packages/grpc-js/src/subchannel.ts
+++ b/packages/grpc-js/src/subchannel.ts
@@ -467,11 +467,13 @@ export class Subchannel {
            * if a connection is successfully established through the proxy.
            * If the proxy is not used, these connectionOptions are discarded
            * anyway */
-          connectionOptions.servername = getDefaultAuthority(
+          const targetPath = getDefaultAuthority(
             parseUri(this.options['grpc.http_connect_target'] as string) ?? {
               path: 'localhost',
             }
           );
+          const hostPort = splitHostPort(targetPath);
+          connectionOptions.servername = (hostPort !== null) ? hostPort.host : targetPath;
         }
       }
     }
--- a/packages/grpc-js/src/subchannel.ts
+++ b/packages/grpc-js/src/subchannel.ts
@@ -473,7 +473,7 @@ export class Subchannel {
             }
           );
           const hostPort = splitHostPort(targetPath);
-          connectionOptions.servername = (hostPort !== null) ? hostPort.host : targetPath;
+          connectionOptions.servername = hostPort?.host ?? : targetPath;
         }
       }
     }
--- a/packages/grpc-js/src/subchannel.ts
+++ b/packages/grpc-js/src/subchannel.ts
@@ -473,7 +473,7 @@ export class Subchannel {
             }
           );
           const hostPort = splitHostPort(targetPath);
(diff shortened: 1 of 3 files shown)


murgatroid99:
I see the problem, but the fix belongs in the proxy code (`http_proxy.ts`), not here. The `getDefaultAuthority` function is used in other places, and it needs to have the port. You should be able to directly call `splitHostPort(parsedTarget.path)` where it is currently calling `getDefaultAuthority`. Also, I don't know where you got `uri_parser_1` from, but it's not a valid reference to anything. `splitHostPort` is already imported, so you can just call it directly.

wkchee:
Thanks for the guidance. Please can help to review the changes again.

packages/grpc-js/src/subchannel.ts
```diff
@@ -467,11 +467,13 @@ export class Subchannel {
...
            * If the proxy is not used, these connectionOptions are discarded
            * anyway */
-          connectionOptions.servername = getDefaultAuthority(
+          const targetPath = getDefaultAuthority(
             parseUri(this.options['grpc.http_connect_target'] as string) ?? {
               path: 'localhost',
             }
           );
+          const hostPort = splitHostPort(targetPath);
+          connectionOptions.servername = (hostPort !== null) ? hostPort.host : targetPath;
```
murgatroid99, line 476: This can be simpler: `connectionOptions.servername = hostPort?.host ?? targetPath`. The same change can be made in the http_proxy changes.

wkchee: Thanks for review.

PR #1556, opened 2020-08-31
fix(grpc-js): Add support for impl type to server.addService

slavovojacek:
Adding support for `ImplementationType` to `server.addService(...)`.

It seems like the original (non-js) `grpc` type definitions support this - https://github.com/grpc/grpc-node/blob/grpc%401.24.x/packages/grpc-native-core/index.d.ts#L266.

Not sure if there was a reason not to do it in grpc-js?

Thanks!

### DIFF
--- a/packages/grpc-js/src/server.ts
+++ b/packages/grpc-js/src/server.ts
@@ -145,10 +145,9 @@ export class Server {
   }
 
   addService<
-    ImplementationType extends Record<
-      string,
-      UntypedHandleCall
-    > = UntypedServiceImplementation
+    ImplementationType extends {
+      readonly [index in keyof ImplementationType]: UntypedHandleCall;
+    }
   >(
     service: ServiceDefinition<ImplementationType>,
     implementation: ImplementationType
@@ -166,7 +165,7 @@ export class Server {
       throw new Error('addService() requires two objects as arguments');
     }
 
-    const serviceKeys = Object.keys(service);
+    const serviceKeys = Object.keys(service) as Array<keyof ImplementationType>;
 
     if (serviceKeys.length === 0) {
       throw new Error('Cannot add an empty service to a server');
@@ -194,13 +193,13 @@ export class Server {
       let impl;
 
       if (implFn === undefined && typeof attrs.originalName === 'string') {
-        implFn = implementation[attrs.originalName];
+        implFn = implementation[attrs.originalName as keyof ImplementationType];
       }
 
       if (implFn !== undefined) {
-        impl = implFn.bind(implementation);
+        impl = implFn.bind(implementation as any);
       } else {
-        impl = getDefaultHandler(methodType, name);
+        impl = getDefaultHandler(methodType, name as string);
       }
 
       const success = this.register(
--- a/packages/grpc-js/src/server.ts
+++ b/packages/grpc-js/src/server.ts
@@ -144,9 +144,14 @@ export class Server {
     throw new Error('Not implemented. Use addService() instead');
   }
 
-  addService(
-    service: ServiceDefinition,
-    implementation: UntypedServiceImplementation
+  addService<
+    ImplementationType extends Record<
+      string,
+      UntypedHandleCall
+    > = UntypedServiceImplementation
+  >(
+    service: ServiceDefinition<ImplementationType>,
+    implementation: ImplementationType
   ): void {
     if (this.started === true) {
       throw new Error("Can't add a service to a started server.");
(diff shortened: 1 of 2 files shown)


packages/grpc-js/src/server.ts
```diff
@@ -144,9 +144,14 @@ export class Server {
...
-  addService(
-    service: ServiceDefinition,
-    implementation: UntypedServiceImplementation
+  addService<
+    ImplementationType extends Record<
+      string,
+      UntypedHandleCall
+    > = UntypedServiceImplementation
+  >(
+    service: ServiceDefinition<ImplementationType>,
```
murgatroid99, line 153: I think this might need to be `Partial<ServiceDefinition<ImplementationType>>`. A service definition is allowed to be missing some methods.

slavovojacek: Don't think `Partial<T>` is needed here as `ServiceDefinition<T>` is defined in such a way that allows the service definition to be missing some methods?

```ts
export interface UntypedServiceImplementation {
  [name: string]: UntypedHandleCall;
}

export type ServiceDefinition<
  ImplementationType = UntypedServiceImplementation
> = {
  readonly [index in keyof ImplementationType]: MethodDefinition<any, any>;
};
```

But I may be wrong... 🤔

murgatroid99: Oh, right, the problem might actually be the other way around: if the implementation object has methods that aren't part of the service definition, those keys will be considered missing in the service definition.

murgatroid99: And I think the `Partial<T>` there solves that problem: the inferred `ServiceDefinition<ImplementationType>` type may have extra keys corresponding to implementation object methods that are not part of the service, and partial allows those to be omitted from the service definition object.

slavovojacek: Will action shortly – just bogged down with other work; sorry!

slavovojacek: @murgatroid99 – would `addService<ImplementationType extends UntypedServiceImplementation = UntypedServiceImplementation>(...)` work by any chance? Will also check locally how that behaves.

slavovojacek: So I looked into it, pushed new update – that's the best I can manage at this stage. Seems to be working.

Screenshot 2020-09-02 at 16 06 24

slavovojacek:
FYI – faced some issues related to https://github.com/microsoft/TypeScript/issues/15300 with the original implementation.

packages/grpc-js/src/server.ts
```diff
@@ -144,9 +144,13 @@ export class Server {
     throw new Error('Not implemented. Use addService() instead');
   }
 
-  addService(
-    service: ServiceDefinition,
-    implementation: UntypedServiceImplementation
+  addService<
+    ImplementationType extends {
+      readonly [index in keyof ImplementationType]: UntypedHandleCall;
```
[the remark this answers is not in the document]
slavovojacek: I changed this just now.

packages/grpc-js/src/server.ts
```diff
@@ -161,7 +165,7 @@ export class Server {
       throw new Error('addService() requires two objects as arguments');
     }
 
-    const serviceKeys = Object.keys(service);
+    const serviceKeys = Object.keys(service) as Array<keyof ImplementationType>;
```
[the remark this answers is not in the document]
slavovojacek: Reverted.

packages/grpc-js/src/server.ts
```diff
@@ -189,13 +193,13 @@ export class Server {
       let impl;
 
       if (implFn === undefined && typeof attrs.originalName === 'string') {
-        implFn = implementation[attrs.originalName];
+        implFn = implementation[attrs.originalName as keyof ImplementationType];
```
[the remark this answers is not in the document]
slavovojacek: Reverted.

packages/grpc-js/src/server.ts
```diff
@@ -189,13 +193,13 @@ export class Server {
...
       }
 
       if (implFn !== undefined) {
-        impl = implFn.bind(implementation);
+        impl = implFn.bind(implementation as any);
```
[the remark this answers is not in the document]
slavovojacek: Reverted.

murgatroid99 (approved):

murgatroid99:
I'm sorry, but I'm going to have to revert this. It broke my code in a different PR, and that probably means that it has the potential to break other people's code.

slavovojacek:
Hi, can you please let me know what broke exactly so I can try and fix?

It's a bit awkward having to write:

```ts
// @ts-ignore
server.addService(HealthService, healthImpl);
// @ts-ignore
server.addService(ContactsService, contactsImpl);
```

Would be good to work together on a solution that you think will work well here.

Issue #1558, opened 2020-09-01
[@grpc/grpc-js] Broken ClientOptions types


grissius:
### Problem description
[ClientOptions type](https://github.com/grpc/grpc-node/blob/be6bdb8c3db6f351493fad3f880d47011ab03daa/packages/grpc-js/src/client.ts#L52) is intersection of its definition and `Partial<ChannelOptions>`. `Partial<ChannelOptions>` defines constants and index signature `[key: string]: string | number | undefined;`, which contradicts with ClientOptions defintions: Since intersection is used, all string attributes of ClientOptions are required to be `string | number | undefined`, which contradicts for example ClientOptions's `interceptors`, which is array of functions.

This issue makes the new interceptor unusable with TS without force-casting.

### Reproduction steps
```typescript
import { Interceptor, Client, ChannelCredentials } from "@grpc/grpc-js";

const interceptor = (null as any) as Interceptor; // just type demonstration
new Client("", ChannelCredentials.createInsecure(), {
  interceptors: [interceptor], // triggers error, whole options object
});
```

```
Argument of type '{ interceptors: Interceptor[]; }' is not assignable to parameter of type 'ClientOptions'.
  Type '{ interceptors: Interceptor[]; }' is not assignable to type 'Partial<ChannelOptions>'.
    Property 'interceptors' is incompatible with index signature.
      Type 'Interceptor[]' is not assignable to type 'string | number'.
        Type 'Interceptor[]' is not assignable to type 'string'.
```

### Environment
- @grpc/js 1.1.5
- Tested on TS 4.0.2 and 3.9.7

### Additional context
We could solve this either by removing the index signature or removing it before the type intersection, which is a bit hard on eyes with generics. Will try to post a PR soon.

PR #1559, opened 2020-09-01
Fix ClientOptions types

grissius:
Remove index signature from ChannelOptions to fix intersection error
described in #1558 which causes issues on using ClientOptions direct
fields with TypeScript.

Removing of index signature required few minor changes:
 - adding few constant types that were used throughout the app
 - using `as const` assertion in xds-client
 - using not-so-great type cast in channelOptionsEqual

Alternative solution would be removing the index signature from
ChannelOptions explicitly in ClientOptions definition, which is not
trivial and probably calls for a generic type helper.

See: https://github.com/grpc/grpc-node/issues/1558
Fixes: 1558

### DIFF
--- a/packages/grpc-js/src/channel-options.ts
+++ b/packages/grpc-js/src/channel-options.ts
@@ -33,7 +33,9 @@ export interface ChannelOptions {
   'grpc.max_send_message_length'?: number;
   'grpc.max_receive_message_length'?: number;
   'grpc.enable_http_proxy'?: number;
-  [key: string]: string | number | undefined;
+  'grpc.http_connect_target'?: string;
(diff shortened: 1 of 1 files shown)


murgatroid99:
I think the correct solution here is actually to just make the index signature have the value type `any`. It's important to allow unspecified keys to break equality because that is how subchannel sharing is determined.

grissius:
@murgatroid99 Thank you for feedback, I am not too familiar with the codebase. I added the _index any_ signature and removed the type obscurities derived from the previous solution.

I justk kept the addition of previously added keys apart from that, which are used throughout the codebase.

murgatroid99 (approved):

PR #1561, opened 2020-09-03
fix(grpc-js): Add support for typed add service

slavovojacek:
Adds improved support for typed `addService`.

Previous attempt (https://github.com/grpc/grpc-node/pull/1556) broke due to https://source.cloud.google.com/results/invocations/a9b120c7-c487-453a-82c8-0d29b91b73ce/targets/grpc%2Fnode%2Fpull_request%2Flinux/log.

### DIFF
--- a/packages/grpc-js/src/server.ts
+++ b/packages/grpc-js/src/server.ts
@@ -144,9 +144,13 @@ export class Server {
     throw new Error('Not implemented. Use addService() instead');
   }
 
-  addService(
-    service: ServiceDefinition,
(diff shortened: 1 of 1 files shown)


murgatroid99:
The problem with the previous change was that it did not properly handle service definition objects that were typed as simply

```ts
interface ServiceDefinition {
  [key: string]: MethodDefinition
}
```

along with implementation objects with types that specified the individual member names. I'm not sure if this change addresses that.

PR #1568, opened 2020-09-10
grpc-js: xDS: handle insecure and google_default bootstrap creds

murgatroid99:
This is an implementation of grpc/proposal#204


stanley-cheung (approved):
LGTM

PR #1571, opened 2020-09-11
grpc-js: Allow clients and servers to send metadata of unlimited size

### LINKED ISSUE #1533: @grpc/grpc-js & grpc.max_metadata_size
In order to migrate from `grpc` to `grpc-js` as [suggested](https://github.com/grpc/grpc-node/issues/1531#issuecomment-671611729) I need to be able to specify [`gprc.max_metadata_size`](https://github.com/grpc/grpc/blob/master/include/grpc/impl/codegen/grpc_types.h#L286), which [isn't supported](https://github.com/grpc/grpc-node/blob/master/PACKAGE-COMPARISON.md) currently. 

If I understand correctly, the default of 64k [is implied by the core `http2` module](https://nodejs.org/api/http2.html#http2_settings_object) and supporting it seems to be [pretty easy](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/server.ts#L249-L254)

murgatroid99:
This fixes #1533.

[Documentation for this option](https://nodejs.org/api/http2.html#http2_http2_connect_authority_options_listener):

> Sets the maximum allowed size for a serialized, compressed block of headers. Attempts to send headers that exceed this limit will result in a `'frameError'` event being emitted and the stream being closed and destroyed.

### DIFF
--- a/packages/grpc-js/src/server.ts
+++ b/packages/grpc-js/src/server.ts
@@ -246,7 +246,9 @@ export class Server {
       throw new Error(`Could not get a default scheme for port "${port}"`);
     }
 
-    const serverOptions: http2.ServerOptions = {};
+    const serverOptions: http2.ServerOptions = {
+      maxSendHeaderBlockLength: Number.MAX_SAFE_INTEGER
+    };
     if ('grpc.max_concurrent_streams' in this.options) {
       serverOptions.settings = {
         maxConcurrentStreams: this.options['grpc.max_concurrent_streams'],
--- a/packages/grpc-js/src/subchannel.ts
+++ b/packages/grpc-js/src/subchannel.ts
@@ -291,6 +291,7 @@ export class Subchannel {
     );
     let connectionOptions: http2.SecureClientSessionOptions =
       this.credentials._getConnectionOptions() || {};
+    connectionOptions.maxSendHeaderBlockLength = Number.MAX_SAFE_INTEGER;
     let addressScheme = 'http://';
     if ('secureContext' in connectionOptions) {
       addressScheme = 'https://';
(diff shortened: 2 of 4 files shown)


nicolasnoble:
I'm a tiny bit conflicted about this one; is there any limit otherwise for metadata sizes on other languages? I understand the nature of the change and its reasons, but I'm pondering if "unlimited" is wise.

ejona86:
_Sending_ is unlimited. _Receiving_ is what we need to limit.

The user has created a Metadata object with all the contents; why are we scared about serializing it? From a memory perspective it doesn't seem an issue.

There are latency impacts of large headers, but receive-limits naturally keep header sizes sane and there are legitimate times users really do need to send large things, at least short-term. We've seen some large error messages be unable to be received and it was helpful to just increase the maximum allowed size on the client to see what error was occurring instead of needing to roll out a new server.

That setting is actually quite strange. I don't know how they've implemented it, as compressing the headers will change the HPACK table. Maybe if this occurs they clear the HPACK table.

nicolasnoble (approved):

murgatroid99:
The received metadata size enforcement is actually not implemented in grpc-js yet, but that's true independent of this change. I can add that in another change soon.

ejona86:
Seems like node-js's http2 [defaults to 64KB for maxHeaderListSize](https://nodejs.org/api/http2.html#http2_settings_object). I sort of assume that means it enforces it in addition to communicating it to the remote. (The HTTP/2 setting is simply advisory, so setting it doesn't guarantee anything by itself.)

murgatroid99:
I think the test in this PR demonstrates that that setting is not enforced automatically. The initial metadata sent by the client is 128KB and the server seems to accept it just fine.

murgatroid99:
My last comment was not correct. Something is definitely getting limited, because the server responds with RST_STREAM with code 9 (COMPRESSION_ERROR). The native server has its own issue: it seems to just hang after it receives the very large headers.

murgatroid99:
Update on this so far: I am having trouble testing this change because of nodejs/node#35218.

PR #1576, opened 2020-09-15
grpc-js: Double the xds test job timeout

murgatroid99:
The test takes more than an hour to finish. The equivalent test job in grpc/grpc also uses a 2 hour timeout: https://github.com/grpc/grpc/blob/master/tools/internal_ci/linux/grpc_xds.cfg.


nicolasnoble (approved):
A bit sad these take this long, but, okay.

murgatroid99:
On the plus side, these aren't actually PR tests. We just run them periodically, so in the long run it doesn't matter that much how long they take.

PR #1580, opened 2020-09-17
grpc-js: Add a timer that holds the event loop open while there are pending calls

### LINKED ISSUE #1579: On slow Internet connection, node process exists before response returned by gRPC
### Problem description

On a slow internet connection, I am finding that a script that performs a gRPC procedure call exits before receiving a response.

### Reproduction steps

The script I am running looks like this:

```js
  const {WorkflowsClient} = require('@google-cloud/workflows');
  const client = new WorkflowsClient();
  async function listWorkflows() {
    const [workflows] = await client.listWorkflows({
      parent: client.locationPath(projectId, location),
    });
    for (const workflow of workflows) {
      console.info(`name: ${workflow.name}`);
    }
  }
  listWorkflows();
```

On a slow connection, we sometimes exit before **console.info(`name: ${workflow.name}`);** is hit.

### Environment
 - OS name, version and architecture: Mac OS X 10.15.6
 - Node version: `v12.18.3`.
 - If applicable, compiler version: n/a
 - Package name and version: `@grpc/grpc-js@1.1.6`.

murgatroid99:
This should fix #1579.

We want to hold the event loop open as long as there is an active call. For most of the lifetime of most calls, this is accomplished by `ref`ing each subchannel as long as it is handling at least one active call. However, that does not account for calls that are never attached to subchannels. This closes that gap by creating an interval timer that lasts for the lifetime of the channel. We `ref` that timer as long as there is at least one pending call on the channel that has not been assigned to a subchannel.

### DIFF
--- a/packages/grpc-js/src/channel.ts
+++ b/packages/grpc-js/src/channel.ts
@@ -137,6 +137,14 @@ export class ChannelImplementation implements Channel {
   private defaultAuthority: string;
   private filterStackFactory: FilterStackFactory;
   private target: GrpcUri;
+  /**
+   * This timer does not do anything on its own. Its purpose is to hold the
+   * event loop open while there are any pending calls for the channel that
+   * have not yet been assigned to specific subchannels. In other words,
+   * the invariant is that callRefTimer is reffed if and only if pickQueue
+   * is non-empty.
+   */
+  private callRefTimer: NodeJS.Timer;
   constructor(
     target: string,
     private readonly credentials: ChannelCredentials,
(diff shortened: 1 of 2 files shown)


packages/grpc-js/src/channel.ts
```diff
@@ -230,6 +239,14 @@ export class ChannelImplementation implements Channel {
       new MaxMessageSizeFilterFactory(this.options),
       new CompressionFilterFactory(this),
     ]);
+
+    this.callRefTimer = setInterval(() => {}, 1 << 31 - 1);
```
nicolasnoble, line 243: Isn't there a maxint constant?

nicolasnoble, line 243: Also, https://www.npmjs.com/package/node-noop :-P

nicolasnoble (approved):

packages/grpc-js/src/channel.ts
murgatroid99: There is `Number.MAX_SAFE_INTEGER`, but it's larger than that, and `Number.MAX_VALUE`, which is even larger.

murgatroid99: Sadly, that's not even the only package that provides a noop function. We used to depend on one of them.

PR #1583, opened 2020-09-21
grpc-js: xDS script debugging

murgatroid99:
There is some problem with Node version mismatches.


murgatroid99:
The tests pass: https://source.cloud.google.com/results/invocations/4fc111bb-8be5-49cf-950f-d3ea1e3bf02e/targets

murgatroid99:
This fixes the xds script by updating the submodules to make sure that the envoy proto files are available, and by using the full path to the `node` executable in the client command passed to the `run_xds_tests.py` script so that it is not affected by changes to the `PATH`.

nicolasnoble (approved):

PR #1589, opened 2020-09-24
grpc-js: xDS: Do not remove watchers in onResourceDoesNotExist

murgatroid99:
The `onResourceDoesNotExist` event is triggered in the CDS LB policy when a CDS update does not have the requested cluster name, and in the EDS LB policy when an EDS update does not have the requested cluster/service name, and when a CDS update does not have any resources with the requested name.

In the first two cases we do not want to remove the watcher because we want to know if the requested resource exists again later. In the last case, we do want to remove the watcher, but that update should also result in the EDS policy either getting removed or getting a new configuration with a new cluster/service name, and either of those events will cause the watcher to be removed or replaced.


stanley-cheung (approved):
LGTM

Issue #1581, opened 2020-09-20
grpc-js & gpc-native-core: Types mismatch for ServerWritableStream


badsyntax:
### Problem description
I'm seeing a type difference in `grpc-js` and `gpc-native-core`.

https://github.com/grpc/grpc-node/blob/ea36f751a1874ed119556fb6782c10bbaafa64b8/packages/grpc-native-core/index.d.ts#L494

https://github.com/grpc/grpc-node/blob/5e0feb5d74d266d435939dd750a091b44c91ff1e/packages/grpc-js/src/server-call.ts#L90-L93

`ServerWritableStream` requires one type argument within `grpc-native-core` and 2 within `grpc-js`.

### Reproduction steps
I'm playing around with the typescript generator within https://github.com/grpc/grpc-node/pull/1474 which generates following service definition:

```ts
export interface ChatHandlers extends grpc.UntypedServiceImplementation {
  join(call: grpc.ServerWritableStream<_chat_package_ClientMessage__Output, _chat_package_ServerMessage>): void;
  send(call: grpc.ServerUnaryCall<_chat_package_ClientMessage__Output, _chat_package_ServerMessage>, callback: grpc.sendUnaryData<_chat_package_ServerMessage>): void;
}
```

<details><summary>Using this proto definition</summary>

```proto
syntax = "proto3";

package chat_package;

message ServerMessage {
  string user = 1;
  string text = 2;
}

message ClientMessage {
  string user = 1;
  string text = 2;
}

service Chat {
  rpc join(ClientMessage) returns (stream ServerMessage) {}
  rpc send(ClientMessage) returns (ServerMessage) {}
}
```

</details>

..which is incompatible with the types found in `grpc-native-core` (the `grpc` npm package)

badsyntax:
@murgatroid99 while https://github.com/grpc/grpc-node/pull/1587 is a step in the right direction, the types are still incompatible.

For example:

```ts
interface IChatServer extends grpc.UntypedServiceImplementation {
  join: grpc.handleServerStreamingCall<proto.ClientMessage, proto.ServerMessage>;
}

const chatServer: IChatServer = {
  // type error using `grpc`
  join(call: grpc.ServerWritableStream<proto.ClientMessage, proto.ServerMessage>): void {
    // implementation
  }
};
```

 I think it makes sense to align `grpc` with `@grpc/grpc-js`, so we can have typed server stream methods (`call.write(chunk: T)`). What do you think? I'll try send a PR today.

[...]

PR #1590, opened 2020-09-25
Add ResponseType to ServerWritableStream

badsyntax:
To provide better type compatibility with `@grpc/grpc-js`. 

See https://github.com/grpc/grpc-node/issues/1581

Here's the `@grpc/grpc-js` type definition for ServerWritableStream:

https://github.com/grpc/grpc-node/blob/5e0feb5d74d266d435939dd750a091b44c91ff1e/packages/grpc-js/src/server-call.ts#L90-L93


murgatroid99 (approved):

badsyntax:
@murgatroid99 any chance this fix can be released as a new patch version for grpc@1.24.x?

PR #1594, opened 2020-10-05
grpc-js: xds: Add more logging around adding and removing eds and cds watchers

murgatroid99:
This is an attempt to track down the cause of a flaky test in the xDS interop tests. The flake appears to be related to a watcher spuriously being removed, which results in the corresponding LB policy thinking that resource has been removed.


stanley-cheung (approved):
LGTM

PR #1596, opened 2020-10-05
tests: Update google-auth-library dependency from 0.9 to 6.1

murgatroid99:
Version 0.9.x of the google-auth-library package was so old that it was using an old metadata server version that is no longer supported. The modifications here are mostly inferred based on the documentation. Hopefully they don't break anything.


stanley-cheung (approved):
LGTM

PR #1607, opened 2020-10-21
grpc-js: Separate xds code into a separate plugin package

murgatroid99:
This PR does a few related things to grpc-js:

 - Replace the `LoadBalancingConfig` structure with one that uses classes, so that external packages can define new `LoadBalancingConfig` types.

 - Export an `experimental` namespace with APIs needed to implement the xDS functionality in a separate package. This namespace is explicitly unstable.

 - Move the xDS code, related submodules, generated code, and test code into a separate package tentatively named `@grpc/grpc-js-xds`


murgatroid99:
The Linux Tests for commit 0a4219e show that the xDS functionality still works in the new split out package.

stanley-cheung (approved):
Sorry only able to do one pass through this and didn't find anything obvious. LGTM.

PR #1610, opened 2020-10-27
Native: Copy slice data into Buffer when receiving binary metadata

### LINKED ISSUE #1407: Node 14 ArrayBuffer changes causes grpc to crash.

murgatroid99:
This should fix #1407, or at least the reproduction demonstrated in https://github.com/grpc/grpc-node/issues/1407#issuecomment-717170288. This change is needed because of change in Node 14 that prohibits multiple `Buffer` objects from pointing to the same underlying memory. That makes the previous zero-copy function error-prone in some situations.

### DIFF
--- a/packages/grpc-native-core/ext/call.cc
+++ b/packages/grpc-native-core/ext/call.cc
@@ -158,7 +158,7 @@ Local<Value> ParseMetadata(const grpc_metadata_array *metadata_array) {
       array = Local<Array>::Cast(maybe_array.ToLocalChecked());
     }
     if (grpc_is_binary_header(elem->key)) {
-      Nan::Set(array, array->Length(), CreateBufferFromSlice(elem->value));
+      Nan::Set(array, array->Length(), CopyBufferFromSlice(elem->value));
     } else {
       // TODO(murgatroid99): Use zero-copy string construction instead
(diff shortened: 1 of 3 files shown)


hugebdu:
@murgatroid99 thank you Michael
will you be able to release this fix?

hugebdu:
@murgatroid99 no way? :(

murgatroid99:
Sorry, I'm working on the release in #1615. There have been some build issues.

murgatroid99:
I have published version 1.24.4, which includes this change.

PR #1615, opened 2020-10-29
Native: add new Node and Electron versions, bump version

murgatroid99:
A couple of previous changes set the `package.json` file out of sync with the template and `build.yaml`, so I corrected that here.

Also add a note in the README specifying the latest version of Node and Electron that we support. We will only have to update that a few times before we stop updating the version support entirely.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,7 +13,7 @@ jobs:
     name: All Simple Native Builds
     strategy:
       matrix:
-        node: ['4', '5', '6', '7', '8', '9', '10', '11', '12', '13', '14', '15']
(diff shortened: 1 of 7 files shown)


nicolasnoble (approved):

kurayama:
@murgatroid99 since this PR increments the version to 1.24.4, will a tag be released soon in github/npm?

murgatroid99:
I am trying to get a build to succeed. When that happens, I will publish a new version.

jmeinke:
@murgatroid99 Could you maybe add electron 10.2 / 10.3 to the build targets for prebuilds? This would be awesome!

Issue #1542, opened 2020-08-17
@grpc/grpc-js & client call cancellation upon deadline


hugebdu:
### Problem description
Could that be grpc-js behaves differently from native grpc in terms of client side call cancellation upon deadline?

Looks like grpc-js's client does not cancel a call if the deadline has reached while the native client does.

See attached repo for reproduction.

Thanks

### Reproduction steps
https://github.com/hugebdu/grpc-js-client-cancellation

murgatroid99:
The issue there is the `parent: call` option, which does deadline and cancellation propagation. That feature is not implemented in `@grpc/grpc-js`, but the option is there for type compatibility between the libraries.

hugebdu:
@murgatroid99 thanks.
is it in your backlog or design decision is not to implement?

murgatroid99:
We prioritized features that we knew people were using, and until you filed this issue, I was not aware that anyone was using that feature. Now we can plan to implement it.

hugebdu:
IMO for SOA those propagations are more than nice to have :)

murgatroid99:
This is now implemented in version 1.2.0

hugebdu:
You're the best!

Thanks a lot

PR #1616, opened 2020-10-30
grpc-js: Implement deadline and cancellation propagation

murgatroid99:
This addresses #1542. The tests I added here are pretty closely adapted from the equivalent tests in the native library: https://github.com/grpc/grpc-node/blob/grpc%401.24.x/packages/grpc-native-core/test/surface_test.js#L1224-L1410.


stanley-cheung (approved):
LGTM

Issue #1535, opened 2020-08-12
@grpc/proto-loader & protobufjs JSON format


hugebdu:
### Is your feature request related to a problem? Please describe.
Any chance to export & make public the [`createPackageDefinition`](https://github.com/grpc/grpc-node/blob/master/packages/proto-loader/src/index.ts#L272)?
We have a use case where our build system crunches multiple protobuf APIs (as `*.proto` files) into a single JSON file using the [`protobufjs#toJSON`](https://github.com/protobufjs/protobuf.js/blob/master/index.d.ts#L195) and then at runtime we load it using the [`grpc#loadObject`](https://github.com/grpc/grpc-node/blob/grpc%401.25.x/packages/grpc-native-core/index.d.ts#L69).

murgatroid99:
You can probably already load those JSON files using `load` or `loadSync` in the `@grpc/proto-loader` library. The arguments to those functions are passed directly to Protobuf.js, so they can load whatever Protobuf.js can load.

We are not going to make `createPackageDefinition` public. One of the goals of the `@grpc/proto-loader` library is to avoid directly exposing Protobuf.js types and APIs.

hugebdu:
indeed, loading JSON instead of proto did work.

thanks!

Issue #807, opened 2019-03-29
Provide access to message objects loaded by protobuf.js from proto-loader load methods


jroper:
Currently, the proto-loader only allows you to access the grpc services defined in a protobuf file. If you want to access the message objects, for example, to be able to use oneof constants, you need to load the same protobuf file again, with the same options, using protobufjs, eg:

```javascript
let protoLoader = require("@grpc/proto-loader");
let protobufjs = require("protobufjs");
let packageDefinition = protoLoader.loadSync(
  "myfile.proto",
  {keepCase: true,
    longs: String,
    enums: String,
    defaults: true,
    oneofs: true
  });
let root = protobufjs.loadSync(
  "myfile.proto",
  {keepCase: true,
    longs: String,
    enums: String,
    defaults: true,
    oneofs: true
  });
```

This seems like a superfluous step, given that grpc proto-loader has already loaded it. Perhaps proto-loader should just modify the returned root namespace, replacing the grpc services with service definitions, and leaving the messages as is. Or, maybe proto-loader could accept an already loaded protobufjs root namespace (essentially, export the `createPackageDefinition` function).

nicolasnoble:
This would mean exposing the protobufjs API directly, which is the opposite of the purpose of this code, that tries to abstract and hide protobufjs to  protect the end user against API breaking changes coming from protobufjs.

murgatroid99:
First, the correspondence between the two functions isn't as close as you think. Most of the options to `protoLoader.loadSync` actually go to `Message#toObject`, not `protobufjs.loadSync`, and you don't mention `includeDirs`, but that is actually unique to `proto-loader`.

Second, the resolution here depends on how you want to use the message types:

 - If you want to construct `Protobuf.js` message objects, that is explicitly out of scope for this library. The library purposely accepts and outputs plain JS objects representing messages.

 - If you want to use information about the message type, the message and enum descriptor objects exported by `proto-loader` may be able to help with that.

novemberborn:
> Or, maybe proto-loader could accept an already loaded protobufjs root namespace (essentially, export the `createPackageDefinition` function).

This would be tremendously helpful. It would allow me to pre-process a root and rewrite certain types, dealing with `gogoproto` annotations.

I'm perfectly fine handling any `protobufjs` incompatibilities that may arise. Not that it's updated frequently anyway.

murgatroid99:
[...]

Issue #550, opened 2018-09-14
createPackageDefinition without doing file I/O?


dckc:
### Is your feature request related to a problem? Please describe.

I have a service `DeployService` that uses types such as `Par`. I'm using `Par.create()` and such in one part of my app where I'm not doing any I/O, and making calls to `DeployService` in other parts of the app.

I can make a `Protobuf.Root` from the .json version of my message types that I get from `pbjs`. This is fine for `Par.create()` and such. But I don't see how to get a `PackageDefinition` from this root. The only option I see is `loadPackageDefinition` that does all the work of pbjs again, only at runtime.

### Describe the solution you'd like

export `createPackageDefinition`.

Or at least make the `Protobuf.Root` available from the `PackageDefinition`.

### Describe alternatives you've considered

I considered forking this code and adding the export declaration.

But out of expediency: I'm loading the message types twice: once for `Par.create()` and such, and again for gRPC calls.

### Additional context

for reference, the project I'm working on is https://github.com/JoshOrndorff/RChain-API

cc @JoshOrndorff

murgatroid99:
One of the primary goals of the proto-loader library is to avoid tying the API too directly to Protobuf.js's API, because we've been burned by that before. We are looking into other solutions for exposing message reflection information in the proto-loader library output, but we never had the intention of allowing people to construct message objects using Protobuf.js APIs made available by the output of gRPC. So if you actually have to construct message objects, you should continue to load the message types separately as you are currently doing.

dckc:
OK, I'll try to forget I ever knew that Protobuf.js was in there.

nicolasnoble:
Also, there's no support for grpc using pbjs / pbts at the moment. See #528.

nicolasnoble:
Or in other words, if you want compile-time code generation of your proto definitions, please use protoc instead. See https://github.com/grpc/grpc/tree/master/examples/node/static_codegen

c0b:
from dcodeIO/protobuf.js#1117 also in need of the `createPackageDefinition` to be exported as well,

[...]

PR #1635, opened 2020-11-22
Add functions for loading and parsing binary-encoded or plain object file descriptor sets

### LINKED ISSUE #1627: Reopen: Support parsing of file descriptor sets into package definitions
Thank you for furthering the gRPC and protobuf communities by providing all these great open source packages!

I am resubmitting the feature request to export the `createPackageDefinition` function in the `@grpc/proto-loader` package.
This feature has been discussed many times across the following threads and beyond:

- https://github.com/grpc/grpc-node/issues/556
- https://github.com/protobufjs/protobuf.js/issues/1117

In fact, this issue is blocking cool features like https://github.com/Urigo/graphql-mesh/pull/1086 from using the canonical `@grpc/proto-loader` package.

I know this feature has already been rejected before, but I feel that it should not have been. If it does not pose a security risk, and the community would like this feature to enable it to build awesome gRPC tooling, then why not?

I've gone ahead and forked this repo and will be submitting a PR to close this issue. If rejected, I see no other way around the issue but to publish and maintain my new fork.

Export the `createPackageDefinition` function in the `@grpc/proto-loader` package.

If there is a better way to build a package definition from a file descriptor proto set, then I am open to closing this issue in favor of that, provided the grpc team can produce a definitive example.
Otherwise, publishing and maintaining a new fork is the best route.

tatemz:
Fixes #1627, #550, #807, and #1535


packages/proto-loader/src/index.ts
```diff
@@ -368,6 +379,38 @@ export function loadSync(
   return createPackageDefinition(root, options!);
 }
 
+export async function loadFileDescriptorSet(
+  descriptorSet: Protobuf.Message<descriptor.IFileDescriptorSet> &
```
murgatroid99, line 383: Instead of a `Message`, the input here should be a `Buffer` containing the binary-encoded message, which should then be decoded within the function. Alternatively, or in addition, the input could be a plain JavaScript object that could be passed to `Protobuf.Message<descriptor.IFileDescriptorSet>.fromObject()`.

As I mentioned in the issue, hiding the Protobuf.js API is a core design goal of this library. That means that the user should not need to have any interaction with the Protobuf.js library directly in order to use any of the APIs exported from this library.

packages/proto-loader/src/index.ts
```diff
@@ -368,6 +379,38 @@ export function loadSync(
   return createPackageDefinition(root, options!);
 }
 
+export async function loadFileDescriptorSet(
```
murgatroid99, line 382: None of the operations in this function are async, so the function itself does not need to be async. I would prefer to make it synchronous in that case for flexibility.

murgatroid99 (changes requested):

packages/proto-loader/src/index.ts
tatemz: That makes sense. Thanks for the advice.

packages/proto-loader/src/index.ts
```diff
@@ -368,6 +379,54 @@ export function loadSync(
...
+  }
+
+  const root = (Protobuf.Root as Protobuf.RootConstructor).fromDescriptor(
+    decodedDescriptorSet
+  );
+  root.resolveAll();
+  return createPackageDefinition(root, options);
+}
+
+export function loadFileDescriptorSetFile(
```
murgatroid99, line 411: I don't think this function is necessary. Its functionality is just "Read a file and pass the contents through a different exported function". That's not specific to this library, so I think we should just let the user determine where the data comes from.

That contrasts with the `load` and `loadSync` functions, in which dependency resolution across files is an essential part of their functionality.

murgatroid99 (changes requested):

murgatroid99 (approved):
Thank you for putting this together.

ardatan:
Do you have an estimated time to release this? Thanks!

murgatroid99:
I was taking time off last month, so I did not release anything during that time. I should be able to release this PR this week.

ardatan:
@murgatroid99 Any updates?

murgatroid99:
Sorry about that, this got lost in the shuffle with other things I was working on. I don't like doing Friday releases, so I'll get this out next week.

ardatan:
Thanks 🙏

ardatan:
Just reminder :) Sorry for bugging you

murgatroid99:
This is now out in version 0.5.6

ardatan:
Thank you 🙏

PR #1650, opened 2020-12-08
Remove `grpc` dependency from grpc-health-check

### LINKED ISSUE #926: request: remove grpc dependency from @grpc/health-check
Not necessarily a problem, but I've had a request to fork this package. I don't really want to. From what I can tell, `grpc.makeGenericClientConstructor()` and `grpc.status.NOT_FOUND` are the only production dependencies on the `grpc` module.

If `health.js` exported a function that accepted a `grpc`-like module, the module could use that as its gRPC dependency.

Forking - but that seems a bit extreme for such a small module.

STRML:
Requires requiring the module with `grpc` as an argument, which requires a major version bump.

Commits also rename the package to "@grpc/health-check" and increments the version to 2.0.0 due to the breaking change on import.

See the README.md changes for new usage.

Tests run on both native module and JS module for now.

Fixes #926


mad-it:
@murgatroid99 seeing as `grpc` has been deprecated, I think it would be nice to give this some love as it would allow the removal of the deprecated library.

murgatroid99:
I agree but I don't think this is the right way to go about it. I think the right way to handle cross-library compatibility is to export library-agnostic objects: the service definition and the service implementation. The client can be derived from the service definition in whichever implementation you're using.

Plus, this PR changes too many independent things at once: updating the version doesn't mean that we need to change the package name, and neither of those things requires the directory structure to change.

STRML:
I'm happy to revert the naming changes, but I thought them useful as this is a breaking change anyway and this package is clearly an outlier in how it is named.

As for your definition/implementation idea, perhaps that could work, but it certainly would require a lot more refactoring. As you can see once you remove whitespace (https://github.com/grpc/grpc-node/pull/1650/files?w=1), the code changes here are really quite minor, easy to review, and easy to test.

murgatroid99:
First, this specific change may be easy to review, but I don't like the API it creates. Passing another package in as an argument to this one feels overly heavyweight for the kind of functionality this package provides.

The health check package provides 4 connected things: the `Client` class, the `messages` namespace, the `service` object, and the `Implementation` class. The `Client` can be created by passing the `service` into `grpc.makeClientConstructor`, so it's essentially redundant. The `messages` exist entirely independent of `grpc`, and the same is true of the `service` object. The only thing that really depends on the gRPC implementation is the `Implementation`, and that's only because it has a single reference to `grpc.status.NOT_FOUND`, which is literally just an element of an enum. So, in short, the `grpc` package is only really needed to provide one element of one enum, which is defined independently in a [public spec](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md). We could just define it as a constant and the dependency on `grpc` can be removed entirely.

As for package naming, it's not really an outlier. It's more like 50/50 in terms of packages published from this repository. And I'm not even saying that changing the package name is wrong, just that it shouldn't be in this PR. Changing the name of the package because we're breaking the API anyway should be a separate decision from the one about how we break the API.

Similarly, moving stuff around in the directory structure is completely unnecessary and makes it more difficult to see changes. That could be done in a separate PR that only moves files, to make both changes easier to review.

PR #1658, opened 2021-01-05
grpc-js: Propagate internal stream errors from the http2 module

murgatroid99:
This should help us understand the causes of some of those errors.

The diff here is large because of an indentation change. The actual change to the `stream.on('close', ...)` handler moves the body into `process.nextTick` to delay processing until after the `error` event is handled, and adds a branch in the `INTERNAL` error case to handle error messages from the error event.


nicolasnoble:
Hint: when checking the "hide whitespace change" box in github, the diff is much more simple to look at :-)

image

nicolasnoble (approved):

PR #1668, opened 2021-01-22
Do not override existing useragent

pgherveou:
It looks like the gRPC client always override the useragent passed in the metdata headers.
Would you accept such a change, where we only set the useragent if there is not one that is passed through the headers already? Happy to add tests if that looks good to you

### DIFF
--- a/packages/grpc-js/src/subchannel.ts
+++ b/packages/grpc-js/src/subchannel.ts
@@ -687,7 +687,9 @@ export class Subchannel {
   ) {
     const headers = metadata.toHttp2Headers();
     headers[HTTP2_HEADER_AUTHORITY] = callStream.getHost();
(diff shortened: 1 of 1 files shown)


murgatroid99:
If you look at the code where `this.userAgent` is set, you can see that it uses channel arguments to determine that value. Our expectation is that any single client will have the same user agent for all requests it makes (or at least, all requests it makes to a single service), so you should set the value once for the whole client using those channel arguments, instead of setting it for every request.

pgherveou:
Thanks for the quick reply, We use the client to proxy upstream calls, so ideally we can just forward the ua, instead of overriding it

murgatroid99:
Our position is that even when used as a proxy, the library should send its own user agent because it is the actual source of the request. So, we are not going to add support for overriding the user agent in this way.

PR #1684, opened 2021-02-04
Create prebuilts for native electron 10 and 11

jmeinke:
Because of this issue, we can't rely on grpc-js: https://github.com/grpc/grpc-node/issues/1158#issuecomment-772490534
Therefore, it would be great if you could release a grpc native version with prebuilt binaries for Electron 10.3 or Electron v11.


.github/workflows/build.yml
```diff
@@ -46,7 +46,7 @@ jobs:
     name: All Electron Builds
     strategy:
       matrix:
-        electron: ['1.0', '1.1', '1.2', '1.3', '1.4', '1.5', '1.6', '1.7', '1.8', '2.0', '3.0', '3.1', '4.1', '4.2', '5.0', '6.0', '6.1', '7.0', '7.1', '7.2', '7.3', '8.0', '8.1', '8.2', '8.3', '8.4', '8.5', '9.0', '9.1', '9.2', '9.3', '10.0', '10.1']
+        electron: ['1.0', '1.1', '1.2', '1.3', '1.4', '1.5', '1.6', '1.7', '1.8', '2.0', '3.0', '3.1', '4.1', '4.2', '5.0', '6.0', '6.1', '7.0', '7.1', '7.2', '7.3', '8.0', '8.1', '8.2', '8.3', '8.4', '8.5', '9.0', '9.1', '9.2', '9.3', '10.0', '10.1', '10.2', '10.3', '11']
```
murgatroid99, line 49:
```suggestion
        electron: ['1.0', '1.1', '1.2', '1.3', '1.4', '1.5', '1.6', '1.7', '1.8', '2.0', '3.0', '3.1', '4.1', '4.2', '5.0', '6.0', '6.1', '7.0', '7.1', '7.2', '7.3', '8.0', '8.1', '8.2', '8.3', '8.4', '8.5', '9.0', '9.1', '9.2', '9.3', '10.0', '10.1', '10.2', '10.3', '11.0', '11.1', '11.2']

```

All of these versions need to have a major and minor version.

murgatroid99:
It looks like some build issues have appeared since we last used this build script. I think it will help if you change `setup-node` to version 2 in both of the places it appears in the workflow file.

.github/workflows/build.yml
jmeinke: Thanks for your suggestions!

murgatroid99:
It looks like it will also be necessary to upgrade `setup-msbuild` to version 1.0.2 in the two places it appears.

jmeinke:
@murgatroid99 The build was successful :tada: Thanks for your help on fixing the build issues - it would be great if you could release this soon-ish :smiley:

nicograef:
This PR would help us a lot!
Thanks for your work @jmeinke and @murgatroid99

jmeinke:
@murgatroid99 Could you please release the version to NPM. Thank you!

jmeinke:
Thanks!

PR #1693, opened 2021-02-17
Remove node requirements for grpc-js

approached:
I' am using the node version:
v16.0.0-head

If i tried to use grpc-js, that not possible with the new node version. I becomme follow error:
```
event - compiled successfully
error - Error: @grpc/grpc-js only works on Node ^8.13.0 || >=10.10.0
    at Object.<anonymous> (/Users/approach/workspace/TakeActiveWeb/node_modules/@grpc/grpc-js/build/src/index.js:49:11)
```

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -5,9 +5,6 @@
   "homepage": "https://grpc.io/",
   "repository": "https://github.com/grpc/grpc-node/tree/master/packages/grpc-js",
   "main": "build/src/index.js",
-  "engines": {
-    "node": "^8.13.0 || >=10.10.0"
-  },
   "keywords": [],
   "author": {
     "name": "Google Inc."


murgatroid99:
This is not the right thing to change. The version range in the package.json file is correct, and it is important to communicate that information to users. The error you mentioned is being thrown from [here](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/index.ts#L69-L72), where the version range is explicitly checked. That check is failing for you because the `-head` version modifier makes the version not match the specified range. You wouldn't get this error on version 16.0.0.

Removing that check would solve this problem. It was originally added because some version compatibility problems lead to more delayed and obscure errors, and this error makes them easier to address. But with Node 8 past EOL and Node 10 nearing the end of its maintenance window, I would be OK with removing that error to avoid problems like this.

murgatroid99:
Thank you for your contribution. I choose to accomplish this using #1793 instead, so this PR is no longer needed.

PR #1698, opened 2021-02-19
Fix bug where streaming errors put http2 session into broken state

### LINKED ISSUE #1490: Using client after dropped connection leads to freezing of the whole NodeJs Process
### Problem description
Please correct me if I'm wrong, this grpc client should automatically reconnect if the connection becomes invalid.
I see this behavior in most cases, however there is one error which makes the client unusable and the whole NodeJs app gets frozen. It' s "Connection dropped". Other errors such as "Channel has been shut down", "Received RST_STREAM with code 2" don't lead to such behavior. 

### Reproduction steps
I create a client at a start of my app, then I occasionally use this grpc client. It works for a few hours but then all of the sudden my app gets frozen. And the last log is always this:

`Connection dropped","metadata":{"internalRepr":{},"options":{}},"level":"warn","message":"Failed to execute grpc callConnection dropped","stack":"Error: Connection dropped\n    at Object.callErrorFromStatus (/usr/src/app/node_modules/@grpc/grpc-js/src/call.ts:81:24)\n    at Object.onReceiveStatus (/usr/src/app/node_modules/@grpc/grpc-js/src/client.ts:334:36)\n    at Object.onReceiveStatus (/usr/src/app/node_modules/@grpc/grpc-js/src/client-interceptors.ts:434:34)\n    at Object.onReceiveStatus (/usr/src/app/node_modules/@grpc/grpc-js/src/client-interceptors.ts:397:48)\n    at Http2CallStream.outputStatus (/usr/src/app/node_modules/@grpc/grpc-js/src/call-stream.ts:230:22)\n    at Http2CallStream.maybeOutputStatus (/usr/src/app/node_modules/@grpc/grpc-js/src/call-stream.ts:280:14)\n    at Http2CallStream.endCall (/usr/src/app/node_modules/@grpc/grpc-js/src/call-stream.ts:263:12)\n    at Http2CallStream.disconnectListener (/usr/src/app/node_modules/@grpc/grpc-js/src/call-stream.ts:215:12)\n    at Socket.session.socket.once (/usr/src/app/node_modules/@grpc/grpc-js/src/subchannel.ts:518:13)\n    at Object.onceWrapper (events.js:286:20)\n    at Socket.emit (events.js:203:15)\n    at Socket.EventEmitter.emit (domain.js:448:20)\n    at TCP._handle.close (net.js:607:12)`

### Temporarily fix
Now if I get an exception during executing a grpc call I just close the current client and create a new one. But I wonder if it's the right approach. 

### Environment
 - Debian GNU/Linux 9 (stretch)
 - Node 10
 - Node installation method: apt
 - "@grpc/grpc-js": "^1.1.1",

### Additional context

ganemone:
Currently there is an issue where errors streaming requests can put the http2 session into a broken state, causing all future requests to timeout. fixes #1490

### DIFF
--- a/packages/grpc-js/src/call-stream.ts
+++ b/packages/grpc-js/src/call-stream.ts
@@ -638,12 +638,12 @@ export class Http2CallStream implements Call {
        * RST_STREAM as a result after we have the status */
       let code: number;
       if (this.finalStatus?.code === Status.OK) {
-        code = http2.constants.NGHTTP2_NO_ERROR;
+        this.http2Stream.close(http2.constants.NGHTTP2_NO_ERROR);
+        this.trace('close http2 stream with code ' + http2.constants.NGHTTP2_NO_ERROR);
       } else {
-        code = http2.constants.NGHTTP2_CANCEL;
+        this.http2Stream.destroy(new Error('Stream error'));
+        this.trace('destroy http2 stream after error ');
       }
-      this.trace('close http2 stream with code ' + code);
-      this.http2Stream.close(code);
     }
   }
 
--- a/packages/grpc-js/src/subchannel.ts
+++ b/packages/grpc-js/src/subchannel.ts
@@ -716,6 +716,9 @@ export class Subchannel {
     }
     logging.trace(LogVerbosity.DEBUG, 'call_stream', 'Starting stream on subchannel ' + this.subchannelAddressString + ' with headers\n' + headersString);
     callStream.attachHttp2Stream(http2Stream, this, extraFilterFactory);
+    http2Stream.on('error', () => {
+      this.session?.close();
+    });
   }
 
   /**


ganemone:
It appears this only fixes the issue for node versions < 14.... With node 14 it seems to get caught in an infinite loop somewhere

pedramphp:
Hi Team,

To give more context to the problem, the issue occurs when a client-streaming connection is destroyed by the gRPC server over a proxy. 

Once http2 session is destroyed by the server, the session in the client transitions to a broken and unusable state. Therefore, all consecutive requests (Unary or Stream) to the server fails, as a result the Node.js app becomes unusable and we would need to restart the Node.js server to get the app functioning normally again.

The expected behavior is that the gRPC client, detects http2 connection failure, logs it and creates a new http2 session for upcoming requests.

I would strongly suggest to prioritize the issue, and work on a fix immediately.

Optionally, we can try to develop a sample demo page, so it can help to reproduce the issue, but looking at the code and the above PR, you can obviously detect the issues.

murgatroid99:
We can't accept this PR. I don't think any of the changes here are correct in terms of gRPC protocol behavior, and I don't understand how they would solve any problems.

Please provide the requested information and anything else that could help debug this problem on the linked issue.

ganemone:
Yea after seeing this doesn't fix the issue for node14 I understand not accepting it. Hopefully it can still be useful for future debugging as we try to find the root cause. Could be useful for others as a short term fix applied as a patch

PR #1654, opened 2020-12-16
Security Fix for Prototype Pollution - huntr.dev

huntr-helper:
https://huntr.dev/users/d3v53c has fixed the Prototype Pollution vulnerability 🔨. Think you could fix a vulnerability like this?

Get involved at https://huntr.dev/

Q | A
Version Affected | ALL
Bug Fix | YES
Original Pull Request | https://github.com/418sec/grpc-node/pull/3
Vulnerability README | https://github.com/418sec/huntr/blob/master/bounties/npm/grpc/1/README.md

### 📊 Metadata *

grpc is vulnerable to Prototype Pollution. This package allowing for modification of prototype behavior, which may result in Information Disclosure/DoS/RCE.

### ⚙️ Description *

Prototype Pollution refers to the ability to inject properties into existing JavaScript language construct prototypes, such as objects.
JavaScript allows all Object attributes to be altered, including their magical attributes such as proto, constructor and prototype.
An attacker manipulates these attributes to overwrite, or pollute, a JavaScript application object prototype of the base object by injecting other values.
Properties on the Object.prototype are then inherited by all the JavaScript objects through the prototype chain.

### 💻 Technical Description *

Fixed by avoiding setting magical attributes. The bug is fixed by validating the input strArray to check for prototypes. It is implemented by a simple validation to check for prototype keywords (proto, constructor and prototype), where if it exists, the function returns the object without modifying it, thus fixing the Prototype Pollution Vulnerability.

### 🐛 Proof of Concept (PoC) *

Create the following PoC file:

`// poc.js`
`var grpc =require('grpc')`
`grpc.loadPackageDefinition({'constructor.prototype.polluted': { service: "Yes! Its Polluted" }});`
`console.log({}.polluted)`

Execute the following commands in another terminal:

`npm i grpc # Install affected module`
`node poc.js #  Run the PoC`

Check the Output:

[Function: ServiceClient] { service: 'Yes! Its Polluted' }`

### 🔥 Proof of Fix (PoF) *

Before:
image

After:
image

### 👍 User Acceptance Testing (UAT)

After fix functionality is unaffected.

grpc-proto-test-1

grpc-proto-test-2
grpc-proto-test-3
grpc-proto-test-4
grpc-proto-test-5
grpc-proto-test-6
grpc-proto-test-7

PR #1701, opened 2021-02-24
grpc: Fix prototype pollution possibility in loadPackageDefinition

murgatroid99:
This is a port of #1654 to the old grpc library.

Credit to @d3v53c for the original fix.

### DIFF
--- a/packages/grpc-native-core/index.js
+++ b/packages/grpc-native-core/index.js
@@ -161,7 +161,7 @@ exports.loadPackageDefinition = function loadPackageDefintion(packageDef) {
   for (const serviceFqn in packageDef) {
     const service = packageDef[serviceFqn];
     const nameComponents = serviceFqn.split('.');
-    if (nameComponents.some(comp => comp === '__proto__')) {
+    if (nameComponents.some(comp => common.isPrototypePolluted(comp))) {
       continue;
     }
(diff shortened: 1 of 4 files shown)


d3v53c:
Hi @murgatroid99 , @nicolasnoble ,

I've tried to improve the fix a little more by using `indexOf` instead of `includes`. Please have a look at the implementations.

1. [PR-1](https://github.com/418sec/grpc-node/pull/7/files)
2. [PR-2](https://github.com/418sec/grpc-node/pull/6)

Cheers!

murgatroid99:
Are the changes in this PR incorrect in some way? Or, in other words, why specifically do you consider the PRs you linked improvements over this one?

d3v53c:
> 
> 
> Are the changes in this PR incorrect in some way? Or, in other words, why specifically do you consider the PRs you linked improvements over this one?

Sorry for the late reply.

The changes are correct when considering the newer versions of JavaScript. The `Array.prototype.includes` is only introduced in ECMAScript 2016 version.
Since, we're looking at an old library and for the sake of compatibility on older versions, it'd probably be better to fallback to `Array.prototype.indexOf` implementations. Please feel free to correct me if I'm wrong.

Cheers!!

nicolasnoble (approved):

PR #1704, opened 2021-03-01
grpc-js-xds: Do traffic splitting and routing in the resolver

murgatroid99:
There are a few changes included in this PR:

 - Refactor some helper classes out of `xds-client.ts` into the directory `xds-stream-state`
 - Modify the `XdsClient` class to be a singleton shared by all channels. In particular, it no longer takes the resolver's `ServiceConfig` watcher in the constructor. Instead, the resolver now watches for LDS and RDS updates the same way the EDS and CDS load balancers watch for their corresponding updates.
 - Add traffic splitting and routing logic to the xDS resolver class, enabled by the environment variable `GRPC_XDS_EXPERIMENTAL_ROUTING`. It creates a `ConfigSelector` as implemented in #1681 and an `XdsClusterManager` as implemented in #1687.

### DIFF
--- a/packages/grpc-js-xds/package.json
+++ b/packages/grpc-js-xds/package.json
@@ -42,7 +42,8 @@
     "yargs": "^15.4.1"
   },
   "dependencies": {
-    "@grpc/proto-loader": "^0.6.0-pre14"
+    "@grpc/proto-loader": "^0.6.0-pre14",
+    "re2-wasm": "^1.0.1"
   },
   "peerDependencies": {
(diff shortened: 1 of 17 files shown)


murgatroid99:
The xDS interop tests, including the newly enabled ones, passed at https://source.cloud.google.com/results/invocations/afea6480-f5fc-40bc-a228-2f5203b8fb1e/targets.

This PR includes the fixes from #1707 to get LRS working so that the load_report_based_failover test would pass.

nicolasnoble (approved):

PR #1705, opened 2021-03-02
grpc-js: Speculative fix for ECONNRESET errors

murgatroid99:
A couple of people in googleapis/nodejs-firestore#1023 have reported calls ending with INTERNAL errors triggered by an underlying error that says "read ECONNRESET". Those should be reported as UNAVAILABLE. My guess is that Node is reporting these errors in the form of a [`SystemError`](https://nodejs.org/api/errors.html#errors_class_systemerror) with the code `ECONNRESET`. If that is the case, we can distinguish those errors and change the code. I also added some logging for the errors from Node to hopefully get a better idea of the error structure if this doesn't work.

If that information isn't actually available in a structured way in the error, we might have to fall back to parsing the error message but I really hope we can avoid that because those messages are supposed to be human-readable, not machine-parseable.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/grpc-js",
-  "version": "1.2.8",
+  "version": "1.2.9",
   "description": "gRPC Library for Node - pure JS implementation",
   "homepage": "https://grpc.io/",
   "repository": "https://github.com/grpc/grpc-node/tree/master/packages/grpc-js",
--- a/packages/grpc-js/src/call-stream.ts
+++ b/packages/grpc-js/src/call-stream.ts
@@ -16,6 +16,7 @@
  */
 
 import * as http2 from 'http2';
+import * as os from 'os';
 
 import { CallCredentials } from './call-credentials';
 import { Propagate, Status } from './constants';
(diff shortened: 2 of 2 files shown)


nicolasnoble (approved):

forconz:
> grpc-web-simple-example@0.1.0 demo       
> .\node_modules\.bin\ts-node .\src\main.ts

D:\Projects\Test\helloworld\node_modules\@grpc\grpc-js\src\call.ts:81
  return Object.assign(new Error(message), status);
                       ^
Error: 14 UNAVAILABLE: read ECONNRESET
    at Object.callErrorFromStatus (D:\Projects\Test\helloworld\node_modules\@grpc\grpc-js\src\call.ts:81:24)
    at Object.onReceiveStatus (D:\Projects\Test\helloworld\node_modules\@grpc\grpc-js\src\client.ts:338:36)
    at Object.onReceiveStatus (D:\Projects\Test\helloworld\node_modules\@grpc\grpc-js\src\client-interceptors.ts:426:34)
    at Object.onReceiveStatus (D:\Projects\Test\helloworld\node_modules\@grpc\grpc-js\src\client-interceptors.ts:389:48)
    at D:\Projects\Test\helloworld\node_modules\@grpc\grpc-js\src\call-stream.ts:276:24
    at processTicksAndRejections (node:internal/process/task_queues:78:11) {
  code: 14,
  details: 'read ECONNRESET',
  metadata: Metadata { internalRepr: Map(0) {}, options: {} }
}

the error code is 14
grpc-js: How to resolve ECONNRESET errors?

The Grpc service work well , I test it by BloomRPC


grpc-client.ts
import * as grpc from "@grpc/grpc-js";
import {PadBotGrpcServiceClient} from './grpc/wechat_grpc_pb';
export default new PadBotGrpcServiceClient('localhost:44392', grpc.credentials.createInsecure());

getTestKeyword.ts
import { TestKeywordRequest, TestKeywordResponse } from './grpc/wechat_pb';
import grpcClient from './grpc-client';

export default async () => {
  return new Promise<TestKeywordResponse>((resolve, reject) => {
    const request = new TestKeywordRequest();
    var metadata = {};

    grpcClient.getTestKeyword(request, metadata, (err, response) => {
      if (err){
        // debug(`[getTestKeyword] err:\nerr.message: ${err.message}\nerr.stack:\n${err.stack}`);
        reject(err);
        return;
      } 
      // log(`[getTestKeyword] TestKeywordResponse: ${JSON.stringify(response.toObject())}`);
      if(response)
      {
        const keyword = response.getKeyword();
        console.log(`keyword = ${keyword}`);
        resolve(response);
      } 
    });
  });
}

main.ts
import getTestKeyword from './getTestKeyword';
const main = async () => {
  await getTestKeyword();
}
  
main();

The code above throw ECONNRESET errors.

PR #1707, opened 2021-03-04
grpc-js-xds: Fix handling of empty LRS server names

murgatroid99:
The LRS load reporting server name is usually the empty string to indicate that the primary xDS server name should be used, but the empty string is falsey so this was taking the wrong branch and LRS was getting disabled when it should have been used.


murgatroid99:
xDS tests passed, including the newly enabled test, at https://source.cloud.google.com/results/invocations/f88530c2-fd18-42f8-94fa-4cfcd8cf4a20/targets

nicolasnoble (approved):

Issue #1077, opened 2019-10-18
Uncaught TypeError: setInterval(...).unref is not a function (when called from electron)


mrfelton:
### Problem description
I'm getting this error when trying to use `@grpc/grpc-js@0.6.9` from within a web worker running under Electron.

> Uncaught TypeError: setInterval(...).unref is not a function

The same code works ok in grpc-js 5.x version.

### Reproduction steps

Try to use latest release in a node enabled web worker within electron.

### Environment
 - OS name, version and architecture: Mac
 - Node version 12.x
 - Node installation method nvm
 - Package name and version @grpc/grpc-js@0.6.9

### Additional context

```
Uncaught TypeError: setInterval(...).unref is not a function
    at new SubchannelPool (subchannel-pool.js?f8bb:55)
    at eval (subchannel-pool.js?f8bb:99)
    at Object../node_modules/@grpc/grpc-js/build/src/subchannel-pool.js (grpc.worker.js:504)
    at __webpack_require__ (grpc.worker.js:20)
    at eval (channel.js?e974:21)
    at Object../node_modules/@grpc/grpc-js/build/src/channel.js (grpc.worker.js:192)
    at __webpack_require__ (grpc.worker.js:20)
    at eval (index.js?0156:22)
    at Object../node_modules/@grpc/grpc-js/build/src/index.js (grpc.worker.js:276)
    at __webpack_require__ (grpc.worker.js:20)
```

murgatroid99:
I am very hesitant to start supporting environments that do not properly implement the Node API. I suggest first filing an issue with Electron asking them to fix how their Node-enabled web workers implement the [timers API](https://nodejs.org/api/timers.html).

mrfelton:
It's a fair comment @murgatroid99 . Unfortunately though I have not had much luck in trying to get to the bottom of the root issue and patching lower down the stack. So until I can do that, I feel that this simple fix would at least enable people to use this in a webpack/electron environment again as they could previously.

murgatroid99:
My assumption is that the root cause of the issue is that the timers API you have access to in that context is not the Node timers API but the web timers API, which doesn't have the reference handling APIs. I would like to at least see an issue on Electron to see what action they will take, if any, to correct the problem.

mrfelton:
https://github.com/electron/electron/issues/21162#issuecomment-554792447

Thoughts?

murgatroid99:
Thank you. I see that they are unwilling to address the issue themselves.

PR #1709, opened 2021-03-05
grpc-js: Timer ref and unref might not exist

### LINKED ISSUE #1708: Timer issue when updating grpc-js from 1.2.6 to 1.2.7
### Problem description
When updating from grpc-js 1.2.6 to 1.27 (and greater - I tried 1.2.8, .9, and .10 as well), there is an issue that arises at least when being used in electronjs, I think it has to do with setTimeout/setInterval using the NodeJS version or browser version, some kind of incompatibility there.

It causes electron's preload.js to fail to load due to the following error with BackoffTimerout.unref:
```
TypeError: this.timerId.unref is not a function
    at BackoffTimeout.unref (VM167 C:\Users\keith…ckoff-timeout.js:99)
    at new ResolvingLoadBalancer (VM162 C:\Users\keith…oad-balancer.js:159)
    at new ChannelImplementation (VM131 C:\Users\keith…\src\channel.js:123)
    at new Client (VM188 C:\Users\keith…ld\src\client.js:58)
    at new ServiceClientImpl (VM191 C:\Users\keith…c\make-client.js:58)
    at new CoreServices (VM123 C:\Users\keith…CoreServices.js:109)
    at Object.<anonymous> (VM87 C:\Users\keithl…ctron\preload.js:62)
    at Object.<anonymous> (VM87 C:\Users\keithl…tron\preload.js:105)
    at Module._compile (VM42 loader.js:1152)
    at Object.Module._extensions..js (VM42 loader.js:1173)
```

### Reproduction steps
I am building this off of secure-electron-template with the addition of gRPC with the following dependencies:
```
  "devDependencies": {
    "@grpc/proto-loader": "^0.5.6",
    "@types/google-protobuf": "^3.7.4",
    "@types/node": "^14.14.31",
    "cross-env": "^7.0.3",
    "grpc-tools": "1.10.0",
    "grpc_tools_node_protoc_ts": "5.1.3",
    "tslint": "^6.1.3",
    "typescript": "^4.2.2"
  },
  "dependencies": {
    "@grpc/grpc-js": "1.2.7",
    "google-protobuf": "3.15.5"
  }

```
See: https://github.com/reZach/secure-electron-template

### Environment
 - OS name, version and architecture: [e.g. Linux Ubuntu 18.04 amd64]
Windows 10 x64
 - Node version [e.g. 8.10.0]
v14.16.0
6.14.11

murgatroid99:
This fixes #1708. We actually ran into the same issue before in #1077 but I forgot when I implemented #1688.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/grpc-js",
-  "version": "1.2.10",
+  "version": "1.2.11",
   "description": "gRPC Library for Node - pure JS implementation",
   "homepage": "https://grpc.io/",
   "repository": "https://github.com/grpc/grpc-node/tree/master/packages/grpc-js",
--- a/packages/grpc-js/src/backoff-timeout.ts
+++ b/packages/grpc-js/src/backoff-timeout.ts
@@ -76,7 +76,7 @@ export class BackoffTimeout {
       this.running = false;
     }, this.nextDelay);
     if (!this.hasRef) {
-      this.timerId.unref();
+      this.timerId.unref?.();
     }
     const nextBackoff = Math.min(
       this.nextDelay * this.multiplier,
@@ -109,11 +109,11 @@ export class BackoffTimeout {
 
   ref() {
     this.hasRef = true;
-    this.timerId.ref();
+    this.timerId.ref?.();
   }
 
   unref() {
     this.hasRef = false;
-    this.timerId.unref();
+    this.timerId.unref?.();
   }
 }


nicograef:
Hey @murgatroid99 , I ran into this issue today and could not find any solution to the issue (besides now finding this PR).
I checked the respective file `grpc-js/src/backoff-timeout.ts` and just removed all the ref/unref the logic. Without it it worked perfectly fine.

As I understand it, this code is made for running in a Node Environment and not inside a Browser. Am I correct?

So the first problem is the wrong type `NodeJS.Timer`

```typescript
// node_modules/@grpc/grpc-js/src/backoff-timeout.ts : 45

private timerId: NodeJS.Timer;
```

Which is wrong for the Browser Environment - it's `number` there.
Hence, the code fails in the Browser (if you don't remove the `.ref()` `.unref()` calls).

Anyways, cool that you fixed this already. For when is the release planned?

Besides, how does grpc-js generally handle the two environments?

Thanks!

murgatroid99:
This library is made to work on Node, not browsers. The reason this timers thing is an issue is Electron: Electron fully supports the Node APIs, *except* that it uses browser timers instead of web timers. So we support Electron, but we have to make changes like this specifically for Electron.

This should release soon, once it gets a code review.

nicolasnoble (approved):

PR #1713, opened 2021-03-12
protobuf.js JSON descriptor

SOVLOOKUP:
**new feature:**
make `@grpc/proto-loader` can parse JSON descriptors in [protobuf.js](https://github.com/protobufjs/protobuf.js#using-json-descriptors)

**code change:**
encapsulated `fromJSON` function
and its test

### DIFF
--- a/packages/proto-loader/src/index.ts
+++ b/packages/proto-loader/src/index.ts
@@ -396,10 +396,9 @@ export function loadSync(
 }
 
 export function fromJSON(
-  json: Protobuf.INamespace,
-  root?: Protobuf.Root
+  json: Protobuf.INamespace
 ): PackageDefinition {
-  const newRoot: Protobuf.Root = root || new Protobuf.Root();
+  const newRoot: Protobuf.Root = new Protobuf.Root();
   const loadedRoot = Protobuf.Root.fromJSON(json, newRoot);
   loadedRoot.resolveAll();
   return createPackageDefinition(newRoot, {});
--- a/packages/proto-loader/src/index.ts
+++ b/packages/proto-loader/src/index.ts
@@ -400,9 +400,6 @@ export function fromJSON(
   options?: Options
 ): PackageDefinition {
   options = options || {};
-  if (!!options.includeDirs) {
-    throw new Error('The fromJSON does not need to load any files, checkout your options');
-  }
   const loadedRoot = Protobuf.Root.fromJSON(json);
   loadedRoot.resolveAll();
   return createPackageDefinition(loadedRoot, options!);
--- a/packages/proto-loader/src/index.ts
+++ b/packages/proto-loader/src/index.ts
@@ -399,17 +399,10 @@ export function fromJSON(
   json: Protobuf.INamespace,
   root?: Protobuf.Root
 ): PackageDefinition {
-  const options: Options = json.options || {};
   const newRoot: Protobuf.Root = root || new Protobuf.Root();
-  if (!!options.includeDirs) {
-    if (!Array.isArray(options.includeDirs)) {
-      throw new Error('The includeDirs option must be an array');
-    }
-    addIncludePathResolver(newRoot, options.includeDirs as string[]);
-  }
   const loadedRoot = Protobuf.Root.fromJSON(json, newRoot);
   loadedRoot.resolveAll();
-  return createPackageDefinition(newRoot, options!);
+  return createPackageDefinition(newRoot, {});
 }
 
 export function loadFileDescriptorSetFromBuffer(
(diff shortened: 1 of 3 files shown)


merlinnot:
Amazing, I was looking for it just last Friday!

murgatroid99:
What is the use case for this?

SOVLOOKUP:
@murgatroid99 I am building a plugin-system based on grpc. I need a protobuf definition with both readability and transmission convenience. I think the [protobufjs official json definition](https://github.com/protobufjs/protobuf.js#using-json-descriptors) is very suitable , but grpc-proto-loader does not support parsing this format. So I implemented it.

merlinnot:
> What is the use case for this?

Currently proto-loader can only load files from disk (it has to read them), which also makes it harder to use in environments for which bundling is preferred (e.g. using Rollup or Webpack) or which disallow disk access completely. It is the most convenient to just import a JSON file (`import * as proto from './my/path/to/proto.json'`) and then call `fromJSON(proto)`.

An example of an environment where bundling to a single file is necessary: Cloudflare Workers.

merlinnot:
Also, using it with ES modules is slightly inconvenient:

```ts
const packageDefinition = loadSync(
  new URL(join(dirname(import.meta.url), '../lib/bundle.json')).pathname,
);
```

as there is no `__dirname`.

packages/proto-loader/src/index.ts
```diff
@@ -395,6 +395,23 @@ export function loadSync(
   return createPackageDefinition(root, options!);
 }
 
+export function fromJSON(
+  json: Protobuf.INamespace,
+  root?: Protobuf.Root
```
murgatroid99, line 400: Protobuf.js objects do not belong in this API. The `root` argument should be removed.

packages/proto-loader/src/index.ts
```diff
@@ -395,6 +395,23 @@ export function loadSync(
...
+): PackageDefinition {
+  const options: Options = json.options || {};
```
murgatroid99, line 402: Does `json.options` actually contain the same options that are used in `createPackageDefinition`?

packages/proto-loader/src/index.ts
```diff
@@ -395,6 +395,23 @@ export function loadSync(
...
+  const newRoot: Protobuf.Root = root || new Protobuf.Root();
+  if (!!options.includeDirs) {
```
murgatroid99, line 404: My understanding is that we are not loading any files, so `includeDirs` is not relevant.

murgatroid99 (changes requested):

SOVLOOKUP:
I remove `json.options` because protobufjs has processed it

https://github.com/protobufjs/protobuf.js/blob/7b1efdb8b6f8a677b8770296f750c47763e33aaf/src/root.js#L46

packages/proto-loader/src/index.ts
```diff
@@ -395,6 +395,15 @@ export function loadSync(
   return createPackageDefinition(root, options!);
 }
 
+export function fromJSON(
+  json: Protobuf.INamespace
+): PackageDefinition {
+  const newRoot: Protobuf.Root = new Protobuf.Root();
```
murgatroid99, line 401: At this point, I don't see any reason to create a new `Root` here. `fromJSON` creates one by default anyway.

packages/proto-loader/src/index.ts
```diff
@@ -395,6 +395,15 @@ export function loadSync(
...
+  const loadedRoot = Protobuf.Root.fromJSON(json, newRoot);
+  loadedRoot.resolveAll();
+  return createPackageDefinition(newRoot, {});
```
murgatroid99, line 404: The options should be an argument, instead of forcing the default values.

murgatroid99 (changes requested):

packages/proto-loader/src/index.ts
SOVLOOKUP: Thank you for your reply. This `{}` indeed a bit ugly, but `options` arg is already included in the definition of json description, I think it maybe unnecessary to add additional argument.

packages/proto-loader/src/index.ts
```diff
@@ -395,6 +395,19 @@ export function loadSync(
   return createPackageDefinition(root, options!);
 }
 
+export function fromJSON(
+  json: Protobuf.INamespace,
+  options?: Options
+): PackageDefinition {
+  options = options || {};
+  if (!!options.includeDirs) {
```
murgatroid99, line 403: Just ignore this option. It doesn't matter.

murgatroid99 (approved):

murgatroid99:

PR #1720, opened 2021-03-19
Update underlying protobuf library to force lower snake case when gen…

taozytroy:
…erate field name


taozytroy:
This is by mistake, should happen in a forked repo

PR #1723, opened 2021-03-22
grpc-js: don't send accept-encoding: gzip

### LINKED ISSUE #1719: @grpc/grpc-js cannot get receivedMessage when gzip is enabled in nginx
### Problem description
When nginx enables gzip, grpc response receivedMessage is always null
When nginx closes gzip, grpc response receivedMessage is normal 

### Reproduction steps
nginx.conf:
...
    gzip on;
    gzip_vary on;
    gzip_types *;
    gzip_proxied  expired no-cache no-store private no_last_modified no_etag auth;
...

src/client-interceptors.ts
... line 423
  if (!receivedMessage) {
      listener?.onReceiveMessage?.(null);
  }
...

### Environment
 - windows 10
 - Node version v14.x
 - @grpc/grpc-js@1.2.11

murgatroid99:
It turns out that if the server sends gzip-encoded content, you have to actually do gzip decompression. grpc-js does not currently have that functionality, so putting `gzip` in the `accept-encoding` header is inaccurate. It can always be added again later.

This should fix #1719.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/grpc-js",
-  "version": "1.2.11",
+  "version": "1.2.12",
   "description": "gRPC Library for Node - pure JS implementation",
   "homepage": "https://grpc.io/",
   "repository": "https://github.com/grpc/grpc-node/tree/master/packages/grpc-js",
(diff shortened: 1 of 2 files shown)


nicolasnoble (approved):

JunweiSUN:
Hi, does this mean that grpc-js 's server side will not support compression? Cause I found that the `grpc-accept-encoding` of `grpc-js` is `identity` while `grpc` is `identity, deflate, gzip`.

murgatroid99:
We needed to make this change because gRPC is implemented using the HTTP/2 protocol, and even though gRPC does not explicitly use the `accept-encoding` header, it still has meaning in HTTP/2 itself, so we need to send the correct value of the header that corresponds to the library's behavior. In this case, since we have gRPC-level compression, we do not also want to be handling HTTP/2-level compression.

Regarding the server, it is true that the grpc-js server currently does not support compression, but that is unrelated to why we had to make this change. This change is in the client code.

And if you read the actual code change in this PR, you can see that it is also sending the `grpc-accept-encoding` header. It sends both headers, because they have different meanings.

PR #1724, opened 2021-03-22
Enable path_matching and header_matching xDS interop tests

murgatroid99:
These are the additional interop tests enabled by #1715.

This PR includes a major refactor of the matching and route action logic in the xDS resolver class, moving them into separate files and multiple classes. The main purpose was to improve the ability to log them for debugging, but that should also improve readability. The refactor also fixed a bug with how the safe regex header match configuration was handled.


murgatroid99:
The xDS interop tests, including the newly enabled test, passed in this run: https://source.cloud.google.com/results/invocations/4b800484-229b-483c-b40e-f570e031aafd/targets

nicolasnoble (approved):

PR #1728, opened 2021-03-23
grpc-js: Add random subset selection to round robin LB policy

murgatroid99:
This adds support for an environment variable `GRPC_EXPERIMENTAL_ROUNDROBIN_RANDOM_MAX_CONNECTIONS` that generally limits the number of connections a round robin load balancing policy will proactively initiate. Those connections are chosen randomly from all of the available addresses. If that environment variable is unset this change should not cause any observable behavior change.

I also removed the `subchannelStateCounts` field because it is redundant with information provided directly by the subchannels, which is now also being used in more places.


murgatroid99:
It looks like we currently do not expect to need this. I will leave the branch around so that this can be reopened if that changes.

murgatroid99:
And now we need it again.

murgatroid99:
And now we don't need it, again

PR #1733, opened 2021-04-01
Update Google Credentials interface to the current version from google-auth-library-nodejs

dragon3:
The current interface `GoogleOAuth2Client` with `getRequestMetadata` function is very old, the function is already removed.

Now `getRequestHeaders` is the function to be used for CallCredentials.

https://github.com/googleapis/google-auth-library-nodejs/blob/6d2fe39d96e5ca607eb4b195aa77936a808c8454/src/auth/authclient.ts#L60

It's better to update the interface so that we can use it in TypeScript 😄

### DIFF
--- a/packages/grpc-native-core/index.d.ts
+++ b/packages/grpc-native-core/index.d.ts
@@ -991,7 +991,7 @@ declare module "grpc" {
      * @param googleCredential The Google credential object to use
      * @return The resulting credentials object
      */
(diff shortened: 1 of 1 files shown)


murgatroid99:
OK, first, the 1.25.x branch is not in use. The 1.24.x branch is the active one. And second, the `createFromGoogleCredential` actually accepts both kinds of objects, so that's what the interface should reflect.

PR #1752, opened 2021-04-16
grpc-js: Add support for ipv4 and ipv6 schemes

murgatroid99:
This adds support for the `ipv4` and `ipv6` schemes as described in the [naming doc](https://github.com/grpc/grpc/blob/master/doc/naming.md). These schemes support multiple addresses in a single target string, and uniquely, multiple different ports. This makes it much easier to meaningfully test the round_robin load balancing policy, and in particular #1728.

### DIFF
--- a/packages/grpc-js/src/resolver-ip.ts
+++ b/packages/grpc-js/src/resolver-ip.ts
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2021 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


packages/grpc-js/src/resolver-ip.ts
```diff
@@ -0,0 +1,107 @@
...
+import { GrpcUri, splitHostPort, uriToString } from "./uri-parser";
+import * as logging from './logging';
+
+const TRACER_NAME = 'ip_resolver';
+
+function trace(text: string): void {
+  logging.trace(LogVerbosity.DEBUG, TRACER_NAME, text);
+}
+
+const IPV4_SCHEME = 'ipv4';
```
nicolasnoble, line 33: Sorts of feels weird that these are private strings here, but I guess there might not be any needs to expose a list of all the known schemes somewhere.

nicolasnoble (approved):

packages/grpc-js/src/resolver-ip.ts
murgatroid99: If they were going to be exposed in any way, it would probably be through the `registerResolver` function, so the string constants in this file would still be private.

PR #1761, opened 2021-04-21
grpc-js: stricter function check than instanceof

zereraz:
`instanceof Function` does not work in vm context, related to https://github.com/grpc/grpc-node/pull/204 https://github.com/grpc/grpc-node/pull/795 and https://github.com/grpc/grpc-node/pull/1759

### DIFF
--- a/packages/grpc-js/src/client.ts
+++ b/packages/grpc-js/src/client.ts
@@ -55,7 +55,7 @@ const INTERCEPTOR_SYMBOL = Symbol();
 const INTERCEPTOR_PROVIDER_SYMBOL = Symbol();
 const CALL_INVOCATION_TRANSFORMER_SYMBOL = Symbol();
 
-function isFunction<ResponseType>(arg: Metadata | CallOptions | UnaryCallback<ResponseType>): boolean {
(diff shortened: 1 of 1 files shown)


packages/grpc-js/src/client.ts
```diff
@@ -198,9 +198,9 @@ export class Client {
     options: CallOptions;
     callback: UnaryCallback<ResponseType>;
   } {
-    if (arg1 instanceof Function) {
+    if (Object.prototype.toString.call(arg1) === '[object Function]') {
```
murgatroid99, line 201: Can you extract this condition into an `isFunction` check?

zereraz: certainly, in this file itself, outside the class?

murgatroid99:
It looks like this doesn't compile. The test log is here: https://source.cloud.google.com/results/invocations/f8f4ecea-d9b2-4d70-8893-604b5a7f30b0/targets/grpc%2Fnode%2Fpull_request%2Flinux/log. Can you fix that?

murgatroid99 (approved):

PR #1780, opened 2021-05-06
grpc-js: Look for ECONNRESET errors by checking the error code

murgatroid99:
This is a different approach to solving the problem I tried to fix in #1705.


nicolasnoble (approved):
The associated log is surprising. This seems brittle. Should we check both `errno` and `code` with a logical or?

PR #1783, opened 2021-05-07
Add CI tests for linux aarch64 (and run them under an emulator).

jtattermusch:
Run unit tests under an aarch64 docker container and use qemu to emulate. Similar approach is already being used for several protobuf languages and for grpc-java and grpc-go.

### DIFF
--- a/test/kokoro_linux_aarch64.sh
+++ b/test/kokoro_linux_aarch64.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+# Copyright 2021 The gRPC Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


jtattermusch:
This is now ready to review. The tests are known to be flaky (see example test results), but it's easier to handle that flakiness once we have a continuously running job on kokoro. So'd prefer to merge this, create a corresponding kokoro job on master branch and then look into how the tests can be deflaked (or whether there'a a aarch64 specific bug).

test/kokoro_linux_aarch64.sh
```diff
@@ -0,0 +1,42 @@
...
+if [[ -t 0 ]]; then
+  DOCKER_TTY_ARGS="-it"
+else
+  # The input device on kokoro is not a TTY, so -it does not work.
+  DOCKER_TTY_ARGS=
+fi
+
+# the test command to run under an emulated aarch64 docker container.
+# we only run tests for a single version of node, since tests under an emulator are significantly slower.
+TEST_NODE_COMMAND="node_versions='8' ./run-tests.sh"
```
murgatroid99, line 34: Node 8 is past EOL. We still support it, but if we can just choose one version to test on, this one shouldn't be the priority. I think it would be best to future proof it and use the latest version, `16`.

PR #1765, opened 2021-04-27
grpc-js-xds: Add xDS v3 API support

murgatroid99:
This PR implements [this proposal](https://github.com/grpc/proposal/blob/master/A30-xds-v3.md#details-of-the-v2-to-v3-transition). It looks enormous but that's mostly the generated code changes in the first commit, which probably don't need to be reviewed.

Issue #1769, opened 2021-04-30
Error: 13 INTERNAL: Received RST_STREAM with code 2 triggered by internal client error: read ECONNRESET


gcxfd:
i reuse a connection for all request , sometime seems connection will disconnect when idle for a long times (client error: read ECONNRESET)

how to set grpc-js for auto reconnection ?

```
Error: 13 INTERNAL: Received RST_STREAM with code 2 triggered by internal client error: read ECONNRESET
    at Object.callErrorFromStatus (/api/node_modules/@grpc/grpc-js/build/src/call.js:31:26)
    at Object.onReceiveStatus (/api/node_modules/@grpc/grpc-js/build/src/client.js:176:52)
    at Object.onReceiveStatus (/api/node_modules/@grpc/grpc-js/build/src/client-interceptors.js:336:141)
    at Object.onReceiveStatus (/api/node_modules/@grpc/grpc-js/build/src/client-interceptors.js:299:181)
    at /api/node_modules/@grpc/grpc-js/build/src/call-stream.js:145:78
    at processTicksAndRejections (node:internal/process/task_queues:76:11)
 {
  code: 13,
  details: 'Received RST_STREAM with code 2 triggered by internal client error: read ECONNRESET',
  metadata: Metadata { internalRepr: Map(0) {}, options: {} }
}
```

murgatroid99:
What version of the library are you using?

murgatroid99:
Would you mind running your client with the environment variables `GRPC_TRACE=call_stream` and `GRPC_VERBOSITY=DEBUG`, and sharing the output from around when you get that error?

gcxfd:
```
→ POST /xlsx
2021-05-08T07:45:59.154Z | call_stream | [6] Sending metadata
2021-05-08T07:45:59.155Z | call_stream | [6] write() called with message of length 29703
2021-05-08T07:45:59.155Z | call_stream | [6] end() called
2021-05-08T07:45:59.155Z | call_stream | Starting stream on subchannel 172.25.187.79:5301 with headers
                grpc-accept-encoding: identity,deflate,gzip
                accept-encoding: identity
                :authority: 172.25.187.79:5301
                user-agent: grpc-node-js/1.3.0
                content-type: application/grpc
                :method: POST
                :path: /office.Office/diff
                te: trailers

2021-05-08T07:45:59.155Z | call_stream | [6] attachHttp2Stream from subchannel 172.25.187.79:5301
[...]

PR #1785, opened 2021-05-12
grpc-js: Apply timeouts from service configs

murgatroid99:
This is the grpc-js side of the timeout part of [proposal A31](https://github.com/grpc/proposal/blob/master/A31-xds-timeout-support-and-config-selector.md): the mechanism for applying a timeout from a service config to each individual request. The xDS side uses xDS v3-specific fields, so I will do that after #1765 is merged.

I changed the method config definition to more closely match [the spec](https://github.com/grpc/grpc/blob/master/doc/service_config.md). The validation code still accepts the same inputs it accepted before, so that shouldn't break anything.

This puts the deadline filter in a weird state. It already had some implicit coupling with the deadline exposed by the call stream class, but now that deadline can change and the filter needs to take action when that happens. I don't really know what to do about that, so I'm deferring that to another time.

### DIFF
--- a/packages/grpc-js/src/call-stream.ts
+++ b/packages/grpc-js/src/call-stream.ts
@@ -70,6 +70,17 @@ function getSystemErrorName(errno: number): string {
 
 export type Deadline = Date | number;
 
+function getMinDeadline(deadlineList: Deadline[]): Deadline {
+  let minValue: number = Infinity;
+  for (const deadline of deadlineList) {
+    const deadlineMsecs = deadline instanceof Date ? deadline.getTime() : deadline;
+    if (deadlineMsecs < minValue) {
+      minValue = deadlineMsecs;
+    }
+  }
+  return minValue;
+}
+
 export interface CallStreamOptions {
   deadline: Deadline;
   flags: number;
@@ -234,6 +245,8 @@ export class Http2CallStream implements Call {
   private listener: InterceptingListener | null = null;
 
   private internalError: SystemError | null = null;
+
+  private configDeadline: Deadline = Infinity;
 
   constructor(
     private readonly methodName: string,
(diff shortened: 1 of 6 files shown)


nicolasnoble (approved):

max-mathieu:
Hi 👋 

Is this going to be added to grpc-js 1.3.x ? I see it got merged in May but isn't in 1.3.7 (latest release). If not, what's the ETA for 1.4.x?
We are trying to add client timeouts across our app (we are facing the exact same issue as https://github.com/grpc/grpc-node/issues/1769#issuecomment-880486957 ) and it be cleaner/safer/more scalable to not pass a `deadline` option with every call, and from the docs, I thought I could use `grpc.service_config` for that

In the meantime, would you consider showing some warnings when validating `service_config` that some of the options passed are not supported by the library? Seems like retries are not supported (which is documented in https://github.com/grpc/grpc-node/blob/master/PACKAGE-COMPARISON.md), but deadlines are listed as supported there... Just thinking that I may not be the only one that will spend hours trying to understand why my service config is not taken into account when the main branch here seems to support it :)

Thanks!

murgatroid99:
This is part of 1.4. We currently don't intend to backport it to 1.3. 1.4 will be released as soon as possible, whenever that ends up being.

The focus of the current service config validation code is validating it according to the spec, so that we don't incorrectly reject service configs retrieved over the network. But I can look into adding some logs there.

Regarding the package comparison doc, "deadlines" there refers to providing deadlines when making requests, not to configuring timeouts in service configs.

You may be right that other users have faced similar problems, but nobody has ever filed an issue or otherwise contacted us about that, so I wouldn't know about it. In fact, the goal of this PR wasn't even to support this use case, it was to support using timeouts with service configs retrieved from an xDS server.

murgatroid99:
Some comments from my team regarding the usage you described:

> The locally provided service config API is really primarily intended to be used in cases where (a) you want to set a default service config for the channel to be used when the resolver cannot find a service config or does not support service configs and (b) use in testing.
>
> There's no reason it won't work in the case you describe, although that's not necessarily the intended use of the API.

> if you rely on this and then a remote service config comes along, then you either:
>1. ignore it, because you disabled the remote config locally
>or
> 2. use its values, which might be missing a deadline on some RPC you were relying upon

> "safer" is to ensure your RPCs always have explicit deadlines determined by the caller.  Add an interceptor and panic if one isn't set if you're worried you will forget to set one.

PR #1789, opened 2021-05-17
grpc-js-xds: Propagate timeouts from xDS responses to method config

murgatroid99:
And add the corresponding interop test. This is the xDS side of [A31](https://github.com/grpc/proposal/blob/master/A31-xds-timeout-support-and-config-selector.md#supported-fields), following up from #1785.


murgatroid99:
Passing test job, including the timeout test: https://source.cloud.google.com/results/invocations/2a12a798-10eb-4257-9cef-c8f96cf4f307/targets

nicolasnoble (approved):

PR #1792, opened 2021-05-19
Don't clean after testing on each Node version

murgatroid99:
Without the native library in this branch, this step may not be needed at all.

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -71,7 +71,6 @@ do
   node -e 'process.exit(process.version.startsWith("v'$version'") ? 0 : -1)'
 
   # Install dependencies and link packages together.
-  ./node_modules/.bin/gulp cleanAll
   ./node_modules/.bin/gulp setup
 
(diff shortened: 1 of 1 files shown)


murgatroid99:
Passing test job, including the timeout test: https://source.cloud.google.com/results/invocations/2a12a798-10eb-4257-9cef-c8f96cf4f307/targets

jtattermusch (approved):
I tried with an adhoc run, and the test build under an emulator is now passing.
There are still a few broken/flaky tests (that will need to be looked at in the future),
but this PR definitely improves the situation a lot.

The adhoc test result:
https://source.cloud.google.com/results/invocations/3f4b535d-aee0-4b93-8513-0576a1732640/targets/github%2Fgrpc-node%2Freports%2Fnode12/tests;passed=true

Thanks for the help, this would have been difficult for me to figure out on my own.
LGTM.

PR #1811, opened 2021-06-08
grpc-js: add log level

gt299792:
before: All log use console.error
after: According to severity, use console.debug, console.info, and console.error.

### DIFF
--- a/packages/grpc-js/src/logging.ts
+++ b/packages/grpc-js/src/logging.ts
@@ -51,7 +51,17 @@ export const setLoggerVerbosity = (verbosity: LogVerbosity): void => {
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const log = (severity: LogVerbosity, ...args: any[]): void => {
   if (severity >= _logVerbosity && typeof _logger.error === 'function') {
-    _logger.error(...args);
+    switch (severity) {
+      case LogVerbosity.DEBUG:
+        _logger.debug(...args);
+        break;
+      case LogVerbosity.INFO:
+        _logger.info(...args);
+        break;
+      case LogVerbosity.ERROR:
+        _logger.error(...args);
+        break;
+    }
   }
 };
 


murgatroid99:
Why?

gt299792:
> Why?

I try to use `setLogger` to catch grpc logger, but all level error use `.error`, I have no idea about which error is error, info or debug

murgatroid99:
I see your point. Unfortunately, this would change a useful aspect of the current behavior, which is that it sends all of the logs to `stderr`, so that they can be redirected consistently.

gt299792:
Maybe write the severity into the second parameter of [this function](https://github.com/grpc/grpc-node/blob/8b34b7a091ef1a8a0d0ff77a936928ebb8c1402c/packages/grpc-js/src/logging.ts#L78) is helpful to distinguish 

https://github.com/grpc/grpc-node/blob/8b34b7a091ef1a8a0d0ff77a936928ebb8c1402c/packages/grpc-js/src/logging.ts#L78

murgatroid99:
I think some of the other libraries log a letter for each severity level as part of the log line. We could get that letter in `log` and then pass it as a new first argument to `_logger.error`.

gt299792:
Should I close this pull request and create a issue?

murgatroid99:
That would probably be a good idea.

franck-romano:
Hey there, 

Thanks for the hard work on this project !

I'm curious, @murgatroid99 why did the solution proposed by @gt299792 has not been accepted ?
We are facing the same issue in our project, and I don't understand why logging using `error`  is the default option here.

Let's say we want to use a logger like [pino](https://github.com/pinojs/pino) with `grpcjs`, if we want to see something (at least in the console), we have to:

1. Set `GRPC_VERBOSITY=DEBUG`, otherwise nothing is logged (because every trace/log is sent to `log()` using `DEBUG` as severity)
2. Use the custom logger

Everything is now logged as error (even successful calls) which make logging a bit unusable in production environment (how to distinguish between errors logs and info logs ?)

Would it be possible to:

- remove the severity check [here](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/logging.ts#L57) since it forces us to use DEBUG mode in our app 
- use the right method to log based on the severity defined by `GRPC_VERBOSITY` ?


Let me know your thoughts, I would be happy to help !

murgatroid99:
I have put together an alternative solution in #1851. The main difference is that it supports sending logs to different functions by severity without modifying the default behavior of sending all of the logs to `stderr`.

I do not see a reason to change the default verbosity. gRPC is a library, so it should not log verbosely by default. Changing the verbosity is as simple as changing a single environment variable (`GRPC_VERBOSITY`) or making a single function call (`grpc.setLogVerbosity()`). I don't think either of those is sufficiently onerous to justify a change to the default behavior. I would be willing to change the severity of individual log lines on a case-by-case basis.

franck-romano:
Nice 👍 thanks for the reactivity !

franck-romano:
Hey, @murgatroid99 do you know when we could expect this change to be released ?

murgatroid99:
The features we are planning to include in the next release should be finished with #1870. So, hopefully pretty soon.

franck-romano:
Nice ! Thank you ☺️

PR #1819, opened 2021-06-18
grpc-js: fix unit test and add semver  types

RenShine:
I want to patch some function,But the unit test is bad, So first mention a pr to fix with the problem of unit testing

The content of this PR is the result of running `npm run fix` and `npm i @types/semver`

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -56,7 +56,8 @@
     "posttest": "npm run check"
   },
   "dependencies": {
-    "@types/node": ">=12.12.47"
+    "@types/node": ">=12.12.47",
+    "@types/semver": "^7.3.6"
   },
   "files": [
     "src/**/*.ts",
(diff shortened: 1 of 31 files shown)


RenShine:
> What problem is this solving?

Unit tests will report many errors, because unit tests include eslint, you can run npm run test if this pr is not included, many errors will be reported

murgatroid99:
OK, I understand fixing the formatting, but why install `@types/semver`? The package doesn't even have a dependency on `semver` anymore.

RenShine:
> OK, I understand fixing the formatting, but why install `@types/semver`? The package doesn't even have a dependency on `semver` anymore.

Used in gulpfile.ts, I moved it to devDependencies

murgatroid99:
OK, I can accept that. Please update from master and sign the CLA, then I can merge this.

murgatroid99:
Thank you for your contribution, but this change has been superseded by #1837.

The issue with the `@types/semver` dependency can be addressed by running `npm install` in the root of the repository.

Issue #1821, opened 2021-06-21
Avoid circular dependencies in @grpc/grpc-js


merlinnot:
### Is your feature request related to a problem? Please describe.

While Node.js supports circular dependencies, it comes with some drawbacks and in my experiences it is best to avoid them. As an example, tree shaking with Rollup becomes less effective.

### Describe the solution you'd like

Could we refactor the code in such a way, that there are no import cycles?

### Describe alternatives you've considered

-

### Additional context

-

murgatroid99:
Are you having issues with circular dependencies in the TypeScript code, or in the compiled JavaScript code? The latter is a subset of the former, and looks easier to fix.

merlinnot:
In my case it's JavaScript. Long-term we could look into TypeScript too, I don't know if that's still the case, but TypeScript compiler used to be sluggish when a project had circular dependencies. I'm not sure if that would matter only for contributors to the library or for TypeScript projects having it as a dependency as well.

Anyway, we can take it one step at a time :slightly_smiling_face:

tom-andersen:
Bump. see: https://github.com/firebase/firebase-js-sdk/issues/8231

murgatroid99:
@tom-andersen In that issue, it looks like you are trying to load the library in the browser. That's not going to work anyway. The Node functionality that grpc-js depends on does not exist in the browser.

tom-andersen:
@murgatroid99 You are correct! There appears to be an additional problem of bundling node dependencies on a web build.

PR #1829, opened 2021-06-24
grpc-js: Refactor code to eliminate runtime dependency cycles

murgatroid99:
The main changes here are moving the `ConnectivityState` enum and `SubchannelAddress` type (plus related definitions) into separate files so that other files can use them without importing `channel.ts` and `subchannel.ts` respectively. The load balancer registry code also makes the default configurable instead of hardcoding it as pick_first, to avoid a dependency cycle between that code and the pick_first code.

This only addresses dependency cycles in the compiled code, because that's what I had a tool to detect. Plus, dependency cycles that include type-only imports shouldn't really be a problem, as long as the types themselves are not cyclically defined.

This addresses #1821.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -29,6 +29,7 @@
     "gulp": "^4.0.2",
     "gulp-mocha": "^6.0.0",
     "lodash": "^4.17.4",
+    "madge": "^5.0.1",
     "mocha-jenkins-reporter": "^0.4.1",
     "ncp": "^2.0.0",
     "pify": "^4.0.1",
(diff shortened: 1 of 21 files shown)


merlinnot:
Thank you for such a quick resolution, helps a lot!

nicolasnoble (changes requested):
There's quite a few places where you've used double quotes for strings instead of simple quotes, (during imports that is), but aside from this, lgtm.

murgatroid99:
Those double quotes were definitely added by VSCode's automatic refactoring functionality (which is otherwise very convenient).

murgatroid99:
I think that the simplest way to handle that will be to just merge this as-is and then do a full reformatting as a separate PR.

nicolasnoble (approved):

PR #1832, opened 2021-06-28
Prevent early return in proto-loader containsDefinition

### LINKED ISSUE #1830: Missing type imports in generated TypeScript declarations
The latest release of @grpc/proto-loader, v0.6.3, generates the following code:

```ts
import type * as grpc from '@grpc/grpc-js';

import type { GreeterClient as _ridedott_run_v1alpha1_GreeterClient, GreeterDefinition as _ridedott_run_v1alpha1_GreeterDefinition } from './ridedott/run/v1alpha1/Greeter';

type SubtypeConstructor<Constructor extends new (...args: any) => any, Subtype> = {
  new(...args: ConstructorParameters<Constructor>): Subtype;
};

export interface ProtoGrpcType {
  ridedott: {
    run: {
      v1alpha1: {
        /**
         * The greeting service definition.
         */
        Greeter: SubtypeConstructor<typeof grpc.Client, _ridedott_run_v1alpha1_GreeterClient> & { service: _ridedott_run_v1alpha1_GreeterDefinition }
        SayHelloRequest: MessageTypeDefinition
        SayHelloResponse: MessageTypeDefinition
      }
    }
  }
}
```

for this proto file:

```protobuf
syntax = "proto3";

package ridedott.run.v1alpha1;

// The request message containing the user's name.
message SayHelloRequest {
  // User's name.
  string name = 1;
}

// The response message containing the greetings.
message SayHelloResponse {
  // Greetings.
  string message = 1;
}

// The greeting service definition.
service Greeter {
  // Sends a greeting.
  rpc SayHello(SayHelloRequest) returns (SayHelloResponse) {}
}
```

`MessageTypeDefinition` is not defined, causing compilation errors.

mtlewis:
Closes #1830.

f289c343b3393aad73795c0657415f4160bcdcb5 introduced a bug - the recursive for-loop descended into the first element's nested array and returned that value without iterating over the other members of the array. This means that the code would only work correctly when the protofile contained a definition whose name was alphabetically before the first  first amongst its siblings.

This commit fixes the issue by only returning early when `containsDefinition` returns true.

This change results in no effect on the output of `yarn generate-golden` - it seems that the echo.proto file used to generate the test output has enough cases where the target type appears before a namespace base sibling that the bug is not exposed.

### DIFF
--- a/packages/proto-loader/bin/proto-loader-gen-types.ts
+++ b/packages/proto-loader/bin/proto-loader-gen-types.ts
@@ -583,8 +583,8 @@ function containsDefinition(definitionType: typeof Protobuf.Type | typeof Protob
   for (const nested of namespace.nestedArray.sort(compareName)) {
     if (nested instanceof definitionType) {
       return true;
-    } else if (isNamespaceBase(nested) && !(nested instanceof Protobuf.Type) && !(nested instanceof Protobuf.Enum)) {
-      return containsDefinition(definitionType, nested);
+    } else if (isNamespaceBase(nested) && !(nested instanceof Protobuf.Type) && !(nested instanceof Protobuf.Enum) && containsDefinition(definitionType, nested)) {
+      return true;
     }
   }
 


murgatroid99:
I would prefer changing line 589 to

```ts
if (containsDefinition(definitionType, nested)) {
  return true;
}
```

and leaving the rest unmodified. As long as all early returns are explicitly `return true`, they can't break anything, and I think the code is clearer that way.

mtlewis:
Thanks @murgatroid99 - I've merged the two if-statements, let me know what you think.

murgatroid99 (approved):
I guess that's fine. Thanks for the quick fix.

merlinnot:
Thanks for a quick fix!

PR #1840, opened 2021-07-01
Increase xDS job timeouts to match Java and Go

ericgribkoff:
With the timeout test, these are already average something like 2:40 so this goes ahead and gives the jobs a (hopefully) excessive time limit.

We are seeing lots of failures due to hitting the 120 minute timeout set on the v1.3.x branch, so this should be backported to v1.3.x and v1.2.x as well.

cc @murgatroid99 

b/192479958


murgatroid99 (approved):

murgatroid99:
Backport in #1843.

Issue #1850, opened 2021-07-15
@grpc/grpc-js throws TypeError "creds._getSettings is not a function".


archana-maharjan:
### Problem description
I'm a newbie to gRPC, and I've been trying to add TLS using the certificates generated by following openssl scripts, however, I'm getting following error, I wonder if I'm missing any config or some parameters for localhost, please let me know what am I doing wrong!
         
- Step 1. Create certificates.conf then add following contents by replacing Country, State and Organization Name

           [req]
            default_bits = 4096
            prompt = no
            default_md = sha256
            req_extensions = req_ext
            distinguished_name = dn
            [dn]
            C = Country Name
            ST = State Name
            O = Organization Name.
            CN = localhost
            [req_ext]
            subjectAltName = @alt_names
            [alt_names]
            DNS.1 = localhost
            IP.1 = ::1
            IP.2 = 127.0.0.1
        
-  Step 2. Create Root signing Key

            openssl genrsa -out ca.key 4096
        
-  Step 3. Generate self-signed Root certificate by replacing Country, State and Organization Names.

            openssl req -new -x509 -key ca.key -sha256 -subj "/C=Country Name/ST=State Name/O=Organization Name" -days 365 -out ca.cert
        
- Step 4. Create a Key certificate for the Server

            openssl genrsa -out service.key 4096
        
- Step 5. Create a signing CSR, note create conf file with

            openssl req -new -key service.key -out service.csr -config certificates.conf
        
- Step 6. Generate a certificate for the Server

            openssl x509 -req -in service.csr -CA ca.cert -CAkey ca.key -CAcreateserial -out service.pem -days 365 -sha256 -extfile certificates.conf -extensions req_ext
       
### Reproduction steps
[grpc-mtls](https://github.com/archana-maharjan/grpc-mtls) has replicable issue. Please add "certificates" folder in "grpc-mtls" root folder and add your valid certificates into it or generate one using above scripts.

### Environment
 - macOs Big Sur, 11.4, i7
 - 14.16.1
 - macOs Installer(.pkg)
 - yarn 1.22.10
 - @grpc/grpc-js 1.3.4, @grpc/proto-loader 0.6.4

### Additional context
```
(node:93745) UnhandledPromiseRejectionWarning: TypeError: creds._getSettings is not a function
    at setupServer (/Users/archanamaharjan/Desktop/dev/practice/grpc-mtls/node_modules/@grpc/grpc-js/src/server.ts:277:17)
[...]

PR #1852, opened 2021-07-15
grpc-js: Tighten server.bindAsync creds typecheck

murgatroid99:
The goal is to give more immediate feedback for errors like the one reported in #1850. Currently if the user passes an object of the wrong type, the error when it is used occurs asynchronously, which loses relevant context and results in a less-helpful error message.

### DIFF
--- a/packages/grpc-js/test/test-server.ts
+++ b/packages/grpc-js/test/test-server.ts
@@ -119,6 +119,10 @@ describe('Server', () => {
       }, /creds must be a ServerCredentials object/);
 
       assert.throws(() => {
+        server.bindAsync('localhost:0', grpc.credentials.createInsecure() as any, noop);
+      }, /creds must be a ServerCredentials object/);
+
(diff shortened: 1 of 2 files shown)


packages/grpc-js/test/test-server.ts
```diff
@@ -116,7 +116,7 @@ describe('Server', () => {
 
       assert.throws(() => {
         server.bindAsync('localhost:0', null as any, noop);
-      }, /creds must be an object/);
+      }, /creds must be a ServerCredentials object/);
```
nicolasnoble, line 119: Maybe we should have another test for an object of the wrong type being passed?

nicolasnoble (approved):

packages/grpc-js/test/test-server.ts
murgatroid99: I added another test that matches the problem the person who filed the linked issue had.

PR #1855, opened 2021-07-16
grpc-js: support grpc+json content type

RenShine:
Golang has a library called jsonpb, which can support grpc requests using json format data. The example library is as follows: 
[https://github.com/johanbrandhorst/grpc-json-example](https://github.com/johanbrandhorst/grpc-json-example)

In order to support back-end development using this library, we need to support the grpc+json content type when we initiate a grpc request

At the same time, if I merge this pr, it can also support other forms of custom codec.

For more information about custom codec, please refer to: [https://pkg.go.dev/google.golang.org/grpc/encoding#Codec](https://pkg.go.dev/google.golang.org/grpc/encoding#Codec)

For more information about jsonpb, please refer to: [https://pkg.go.dev/github.com/golang/protobuf/jsonpb](https://pkg.go.dev/github.com/golang/protobuf/jsonpb)

### DIFF
--- a/packages/grpc-js/src/channel-options.ts
+++ b/packages/grpc-js/src/channel-options.ts
@@ -37,6 +37,7 @@ export interface ChannelOptions {
   'grpc.http_connect_target'?: string;
   'grpc.http_connect_creds'?: string;
   'grpc-node.max_session_memory'?: number;
+  'grpc.http_content_type'?: string;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   [key: string]: any;
 }
--- a/packages/grpc-js/src/channel.ts
+++ b/packages/grpc-js/src/channel.ts
@@ -402,7 +402,8 @@ export class ChannelImplementation implements Channel {
                     pickResult.subchannel!.startCallStream(
                       finalMetadata,
                       callStream,
-                      pickResult.extraFilterFactories
+                      pickResult.extraFilterFactories,
+                      this.options
                     );
                     /* If we reach this point, the call stream has started
                      * successfully */
--- a/packages/grpc-js/src/subchannel.ts
+++ b/packages/grpc-js/src/subchannel.ts
@@ -655,14 +655,15 @@ export class Subchannel {
   startCallStream(
     metadata: Metadata,
     callStream: Http2CallStream,
-    extraFilterFactories: FilterFactory<Filter>[]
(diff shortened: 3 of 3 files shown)


RenShine:
Usage example :

`const creator = makeGenericClientConstructor(testService,'example.UserService')
const client = new creator('127.0.0.1:31702', credentials.createInsecure(),{
  'grpc.http_content_type': 'application/grpc+json'
});`

murgatroid99:
I don't think we need this.

Only the Go library does this. The behavior of the Node library is modeled more after the libraries in other languages, particularly the C core library that the original `grpc` package is built on, and those don't have any functionality like this. The fact is that most of the other libraries also don't check that value or do anything with it. Generally, the message serialization and deserialization functions are associated with each method, and those can be anything, including JSON serialization and deserialization.

RenShine:
> I don't think we need this.
> 
> Only the Go library does this. The behavior of the Node library is modeled more after the libraries in other languages, particularly the C core library that the original `grpc` package is built on, and those don't have any functionality like this. The fact is that most of the other libraries also don't check that value or do anything with it. Generally, the message serialization and deserialization functions are associated with each method, and those can be anything, including JSON serialization and deserialization.


It means that if we support grpc+json format, nodejs and golang communication will no longer need to load protobuf files

RenShine:
If the communication between nodejs and golang requires protobuf files, then in actual use, it is necessary to frequently update and synchronize the protobuf repositories of both parties in actual use. This is a problem we encountered in our use. This pr is also to solve this problem.

RenShine:
You can like this

```
const testService = {
  // Sends a greeting
  GetUserInfo: {
    path: 'example.UserService',
    requestStream: false,
    responseStream: false,
    requestSerialize:  (arg: any) => {
      return Buffer.from(JSON.stringify(arg))
    },
    responseSerialize: () => { return new Buffer("test")},
    requestDeserialize: () => {},
    responseDeserialize:   (res: any) => {
      return JSON.parse(res.toString())
    },
  },
};

const creator = makeGenericClientConstructor(testService,'example.UserService')
const client = new creator('127.0.0.1:31702', credentials.createInsecure(),{
  'grpc.http_content_type': 'application/grpc+json'
});
client.GetUserInfo({"user_id":2},(err: any,res: any) => {
  console.log("err:",err,"res:",res)
});
````

PR #1866, opened 2021-07-30
Change certificate in callback

askidmobile:
On communicate module we need to verifity certificate on server.
For verifity we use fingerprint, subjectaltname.
But this grpc version callback only raw data. 

For this way we use nodejs doc:
https://nodejs.org/api/tls.html#tls_certificate_object

I'm change file for coorect. Please apply pull request, so we can use them through npm. Thnk's.

### DIFF
--- a/packages/grpc-js/src/channel-credentials.ts
+++ b/packages/grpc-js/src/channel-credentials.ts
@@ -28,16 +28,6 @@ function verifyIsBufferOrNull(obj: any, friendlyName: string): void {
 }
 
 /**
- * A certificate as received by the checkServerIdentity callback.
- */
-export interface Certificate {
-  /**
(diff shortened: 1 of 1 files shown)


1gkx (approved):

murgatroid99:
Sorry for the delayed response. I think we can accept this, but you can simplify it further. With your change, now the `checkServerIdentity` function defined there is just passing its arguments through unchanged. So, we can just use the `verifyOptions.checkServerIdentity` function directly. In other word, you can replace line 191 with

```ts
this.connectionOptions = {
  secureContext,
  checkServerIdentity: verifyOptions?.checkServerIdentity
};
```

And remove the entire `if` block after that.

murgatroid99:
This was superseded by #1968.

PR #1869, opened 2021-08-02
add xds support reading bootstrap config directly from env var (#1868)

huan:
See: #1868 

Link to: https://github.com/grpc/grpc-java/pull/7826

### DIFF
--- a/packages/grpc-js-xds/src/xds-bootstrap.ts
+++ b/packages/grpc-js-xds/src/xds-bootstrap.ts
@@ -244,34 +244,66 @@ export async function loadBootstrapInfo(): Promise<BootstrapInfo> {
   if (loadedBootstrapInfo !== null) {
     return loadedBootstrapInfo;
   }
+
+  /**
+   * If GRPC_XDS_BOOTSTRAP exists
+   *  then use its value as the name of the bootstrap file.
+   *
+   * If the file is missing or the contents of the file are malformed,
+   *  return an error.
+   */
   const bootstrapPath = process.env.GRPC_XDS_BOOTSTRAP;
-  if (bootstrapPath === undefined) {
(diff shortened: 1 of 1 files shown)


packages/grpc-js-xds/src/xds-bootstrap.ts
```diff
@@ -244,34 +244,66 @@ export async function loadBootstrapInfo(): Promise<BootstrapInfo> {
...
+   */
+  const bootstrapConfig = process.env.GRPC_XDS_BOOTSTRAP_CONFIG;
+  if (bootstrapConfig) {
+    try {
+      const parsedConfig = JSON.parse(bootstrapConfig);
+      const loadedBootstrapInfoValue = validateBootstrapFile(parsedConfig);
+      loadedBootstrapInfo = Promise.resolve(loadedBootstrapInfoValue);
+    } catch (e) {
+      throw new Error(
+        `Failed to parse xDS bootstrap config from process.env.GRPC_XDS_BOOTSTRAP_CONFIG with error ${e.message}`
```
murgatroid99, line 297:
```suggestion
        `Failed to parse xDS bootstrap config from environment variable GRPC_XDS_BOOTSTRAP_CONFIG with error ${e.message}`
```

From the user's point of view, they didn't set `process.env.whatever`, they set an environment variable, so that's what the error message should refer to.

murgatroid99 (approved):
Other than a change to the wording of the new error, this looks good. Thank you for adding this.

packages/grpc-js-xds/src/xds-bootstrap.ts
huan: Yes, you are correct, I have followed your suggestion and fixed the message.

PR #1874, opened 2021-08-06
fix(docs): s/supporst/supports

jason-yolabs:


murgatroid99 (approved):
Thanks for the fix. Unfortunately, even with a trivial change like this, I can't accept it unless you sign the CLA.

Neilblaze:
@jason-yolabs This issue has been pending for more than 28 days, so if you don't mind, I'm taking this up!

jason-yolabs:
@Neilblaze go for it :)

PR #1870, opened 2021-08-04
grpc-js-xds: Add fault injection HTTP filter

murgatroid99:
This is an implementation of [Proposal A33: Fault Injection](https://github.com/grpc/proposal/blob/master/A33-Fault-Injection.md), using the HTTP filter API added in #1853.

PR #1879, opened 2021-08-10
grpc-js-xds: Port fixes from fault injection PR

murgatroid99:
This is a port of fixes to existing files from #1870.


murgatroid99:
The main problem here is that the HTTP filter implementation is functionally incomplete without fault injection, because TD always sends non-optional filter configs, and the environment variable protection doesn't affect these tests because the feature is automatically enabled [here](https://github.com/grpc/grpc/blob/master/tools/run_tests/run_xds_tests.py#L3225). Fixing the HTTP filter implementation won't change that.

PR #1882, opened 2021-08-17
grpc-js-xds: Distinguish v2 and v3 when handling messages

murgatroid99:
The main purpose of this change is to implement the following line of [proposal A39](https://github.com/grpc/proposal/blob/master/A39-xds-http-filters.md):

> gRPC will support HTTP filters in xDS v3 only. Filters will continue to be ignored when speaking xDS v2.


murgatroid99:
The xDS test results are [here](https://source.cloud.google.com/results/invocations/ee72d3da-35d4-4a1f-b0b4-20530fbb64e5/targets;collapsed=). The metadata_filter test is failing because of a different bug, but the others passed.

nicolasnoble (approved):

PR #1896, opened 2021-09-07
grpc-js: fixed miss content type

RenShine:
https://chromium.googlesource.com/external/github.com/grpc/grpc/+/HEAD/doc/PROTOCOL-HTTP2.md

In this document, it is pointed out `Content-Type → “content-type” “application/grpc” [(“+proto” / “+json” / {custom})]`

So I think the content type is not always `application/grpc`, the server is entirely possible to have its own implementation

As a library that may be used for grpc client, we should support passing content type from outside

And ChannelOptions may also have other functions, such as keepalive_timeout_ms, max_receive_message_length

example code: `const client = new creator('127.0.0.1:31702', credentials.createInsecure(),{ 'grpc.http_content_type': 'application/grpc+json' });`

### DIFF
--- a/packages/grpc-js/src/channel-options.ts
+++ b/packages/grpc-js/src/channel-options.ts
@@ -37,6 +37,7 @@ export interface ChannelOptions {
   'grpc.http_connect_target'?: string;
   'grpc.http_connect_creds'?: string;
   'grpc-node.max_session_memory'?: number;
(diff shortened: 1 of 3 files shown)


murgatroid99:
Please stop. This change was already rejected in #1855. Sending the same change in a new PR won't change that.

PR #1900, opened 2021-09-13
adds sometimes required Host header to proxy connection

bbesset:
my corporate proxy requires the Host header to be defined to the target host, orelse it returns a 403 error


murgatroid99:
For future reference for anyone else looking at this PR, the [HTTP CONNECT spec](https://httpwg.org/specs/rfc7231.html#CONNECT) says to include the Host header with this value.

murgatroid99:
Thank you for the contribution. I can't merge it until you resolve the CLA issue.

bbesset:
thank you

it took me a while, but the CLA issue is resolved now.

murgatroid99 (approved):

PR #1908, opened 2021-09-20
grpc-js-xds: Log loaded bootstrap info in xDS client (1.4.x)

murgatroid99:
We need this to correlate client logs with TD logs. (PR for 1.4.x)

### DIFF
--- a/packages/grpc-js-xds/src/xds-client.ts
+++ b/packages/grpc-js-xds/src/xds-client.ts
@@ -371,6 +371,13 @@ export class XdsClient {
           ...nodeV3,
           client_features: ['envoy.lrs.supports_send_all_clusters'],
         };
+        if (this.apiVersion === XdsApiVersion.V2) {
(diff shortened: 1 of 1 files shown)


packages/grpc-js-xds/src/xds-client.ts
```diff
@@ -340,6 +340,7 @@ export class XdsClient {
         if (this.hasShutdown) {
           return;
         }
+        trace('Loaded bootstrap info: ' + JSON.stringify(bootstrapInfo, undefined, 2));
```
lidizheng, line 343: Is it possible to print the final bootstrap with the build_version, user_agent_name, server_features fields?

murgatroid99: In this code that's not really treated as the "final bootstrap", but I can log the final `Node` messages that will be sent in the ADS and LRS requests.

murgatroid99: I added logging for those too.

lidizheng (approved):
Thanks for adding this log.

PR #1915, opened 2021-09-27
Add channelz support and the admin interface

murgatroid99:
Add channelz functionality to grpc-js (see [gRFC A14](https://github.com/grpc/proposal/blob/master/A14-channelz.md)), along with the admin interface (see [gRFC A38](https://github.com/grpc/proposal/blob/master/A38-admin-interface-api.md)).

There are three main parts of this change:

 - The channelz bookkeeping logic, which is spread among the first part of `channelz.ts`, `channel.ts`, `subchannel.ts`, and `server.ts`, with small parts in `call-stream.ts` and `server-call.ts` to count messages.
 - The channelz service handler code, which all lives in the second part of `channelz.ts`.
 - The admin interface, which is simply a registry of service definitions and corresponding service handlers to be added to a server as a group using the `addAdminServicesToServer` function. The registration function takes getters instead of plain objects to support retrieving those things lazily, which allows us to load the channelz service lazily so that we do not need to load `protobufjs` in applications that do not use channelz.

This change adds the following public APIs:

 - `getChannelzServiceDefinition(): ServiceDefinition`
 - `getChannelzHandlers(): ServiceHandler`
 - `addAdminServicesToServer(server: Server): void`
 - `Channel#getChannelzRef(): ChannelRef`
 - `Server#getChannelzRef(): ServerRef`

The intended use is for a user to do one of the following to serve the channelz service:

```ts
const server = new grpc.Server();
server.addService(grpc.getChannelzServiceDefinition(), grpc.getChannelzHandlers());
```

OR

```ts
const server = new grpc.Server();
grpc.addAdminServicesToServer(server);
```

In addition, I modified the channel, subchannel, and server trace logs to include the channelz ID to make them easier to cross-reference.

The new `channelz_manual_test.js` file just starts a server and a client that continuously makes requests to that server, and starts another server serving the channelz service, for a separate channelz client to examine.

### DIFF
--- a/packages/grpc-js/proto/channelz.proto
+++ b/packages/grpc-js/proto/channelz.proto
@@ -1,0 +1,564 @@
+// Copyright 2018 The gRPC Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This file defines an interface for exporting monitoring information
+// out of gRPC servers.  See the full design at
+// https://github.com/grpc/proposal/blob/master/A14-channelz.md
+//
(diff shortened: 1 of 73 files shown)


packages/grpc-js/proto/channelz.proto
```diff
@@ -0,0 +1,564 @@
+// Copyright 2018 The gRPC Authors
```
nicolasnoble, line 1: It'd still be great at some point if we had a common repository for those...

nicolasnoble (approved):

Issue #1823, opened 2021-06-21
Please add support for server compression


jlecomte:
### Is your feature request related to a problem? Please describe.
Not really, other than the fact that I cannot migrate my server from `grpc` to `@grpc/grpc-js` until `@grpc/grpc-js` supports compression. Our clients (millions of them...) send messages compressed (gzip) in a bi-directional stream.

### Describe the solution you'd like
Something along the lines of what is implemented in https://www.npmjs.com/package/grpc-server-js (I just perused through that code and found out that they already support compression)

### Describe alternatives you've considered
Keep using the `grpc` package for now, even though it explicitly says that it has been deprecated. Compression is a major feature IMO, so maybe the deprecation was a little hasty?

murgatroid99:
We have prioritized feature work based on what features we were aware of our users using. The grpc-js project has existed for 3 years and as far as I know this is the first time someone has mentioned server compression, which is why we have not implemented it. The only part of compression that is fully implemented is client-side parsing of compressed messages, because we knew of users that needed that.

It looks like the server does have a bug where it assumes that all messages it receives are uncompressed, without checking. Technically, the server is not required to return UNIMPLEMENTED as soon as it gets a `grpc-compression` header other than `identity`. That client can subsequently send every message uncompressed, and if it does, the server can accept them.

jlecomte:
Yup, understood on the prioritization. I only recently started looking at the `grpc-js` project. I wish I had done so sooner to let you know of my need. Thanks for the help!

merlinnot:
If I only knew user feedback is needed, I'd open an issue a long time ago 🙂 Me and other devs at my company are waiting for it ever since we switched from `grpc`. We just just accepted lack of it for now.

We weren't putting pressure on it because we've seen continuous progress on other features that were useful to us, thank you for that. And I still think there might be more important ones, such as addressing the performance differences (raised in other issues), so I get your point about priorities - I'm dropping this message here just to let you know it would be useful for more people.

murgatroid99:
I'm sorry for not communicating that more effectively. I mentioned it in the [1.0 announcement blog post](https://grpc.io/blog/grpc-js-1.0/), but I don't know how many people saw that. Other than that, I expected that people would say something if a feature they needed was missing.

Lutzifer:
We also have a use case for this feature in our projects.

murgatroid99:
[...]

PR #1952, opened 2021-10-25
grpc-js: Allow per-channel request compression from the client and decompression from the server

b0b3rt:
This pull requests implements `deflate` and `gzip` compression on a per-channel basis (when instantiating a new client) and the corresponding decompression on the server side.

For the client, this is done through the addition of the `grpc.default_compression_algorithm` option to `ChannelOptions`, which is translated to the appropriate `grpc-encoding` header.

The server reads the `grpc-encoding` header and attempts the corresponding decompression, and fails with the status `INTERNAL` if the decompression fails, per the [spec](https://github.com/grpc/grpc/blob/master/doc/compression.md#test-cases).

Support for this functionality has been requested in this [issue](https://github.com/grpc/grpc-node/issues/1823).

I also recently found myself in a position where it'd be nice to have, so decided to give the whole open source thing a go :)

Other notes:

- changes to package.json: when I first pulled down the repo and started running the setup scripts, they failed because a recent version of `@types/node` [changed the type](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/55965) of `Socket.localAddress` from `string` to `string?`, causing compilation failures.  If this change is actually a fix that reflects a pre-existing truth that `localAddress` could be missing/undefined at runtime, that could indicate that there's a fix required in `Server.getChannelzSessionInfoGetter` and `Subchannel.getChannelzSocketInfo`, but that'd be better addressed by a separate PR.  The build similarly complained about not having `@types/semver` installed.
- newly generated TestService files - these are used as fixtures for the tests I added to cover the new functionality, similar to how the files generated for `channelz.proto` are used in `test-channels.ts`.
- usage of message flags - I'm not _totally_ clear on the original motivation for using flags passed down through the message context to decide whether compression occurs (in `compression-filter.ts`).  As far as I could determine, there isn't any way for a user to pass down flags when sending a message, so trying to use them to enable/disable compression would require strictly more work than just checking the ChannelOptions.  Happy to modify this if I'm missing something.


packages/grpc-js/src/compression-filter.ts
```diff
@@ -167,10 +183,44 @@ function getCompressionHandler(compressionName: string): CompressionHandler {
 export class CompressionFilter extends BaseFilter implements Filter {
   private sendCompression: CompressionHandler = new IdentityHandler();
   private receiveCompression: CompressionHandler = new IdentityHandler();
+  private defaultCompressionAlgorithm: CompressionAlgorithm | undefined;
```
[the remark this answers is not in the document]
b0b3rt: Makes sense.

packages/grpc-js/src/compression-filter.ts
```diff
@@ -167,10 +183,44 @@ function getCompressionHandler(compressionName: string): CompressionHandler {
...
+      }
+    }
+  }
+
   async sendMetadata(metadata: Promise<Metadata>): Promise<Metadata> {
     const headers: Metadata = await metadata;
     headers.set('grpc-accept-encoding', 'identity,deflate,gzip');
     headers.set('accept-encoding', 'identity');
+
+    if (this.defaultCompressionAlgorithm && ['deflate', 'gzip'].includes(this.defaultCompressionAlgorithm)) {
```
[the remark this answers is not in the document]
b0b3rt: Yeah, good call, there's no reason to have this check anymore.  I think explicitly excluding `identity` does make sense, if only to save on the bandwidth.

packages/grpc-js/src/server-call.ts
```diff
@@ -136,12 +137,13 @@ export class ServerReadableStreamImpl<RequestType, ResponseType>
   constructor(
     private call: Http2ServerCallStream<RequestType, ResponseType>,
     public metadata: Metadata,
-    public deserialize: Deserialize<RequestType>
+    public deserialize: Deserialize<RequestType>,
+    encoding?: string
```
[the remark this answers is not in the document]
b0b3rt: Yep, went ahead and made that consistent everywhere.

packages/grpc-js/src/server-call.ts
```diff
@@ -439,6 +442,47 @@ export class Http2ServerCallStream<
...
+              resolve();
+            } else {
+              const joined = Buffer.concat([message.slice(0, 5), output]);
+              resolve(joined);
+            }
+          });
+        });
+      }
+  
+      default:
```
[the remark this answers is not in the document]
b0b3rt: 👍

packages/grpc-js/src/compression-filter.ts
```diff
@@ -194,7 +257,7 @@ export class CompressionFilter extends BaseFilter implements Filter {
     const resolvedMessage: WriteObject = await message;
     const compress =
```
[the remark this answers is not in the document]
b0b3rt: Yeah, it's a bit too much for a ternary expression.  Split it out as suggested.

packages/grpc-js/src/server-call.ts
```diff
@@ -439,6 +442,56 @@ export class Http2ServerCallStream<
...
+      }
+
+      case 'identity': {
+        return Promise.resolve(message);
+      }
+  
+      default: {
+        this.sendError({
+          code: Status.UNIMPLEMENTED,
+          details: `Received "grpc-encoding" header "${encoding}", which is not supported`,
```
murgatroid99, line 488:
```suggestion
          details: `Received message compressed with unsupported encoding "${encoding}"`,
```

The server sends this error only when it receives a message that is actually compressed with an unsupported encoding, not just when it receives the header.

packages/grpc-js/src/compression-algorithms.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+export const CompressionAlgorithms = {
```
murgatroid99, line 18: What is the benefit of using `const` here instead of `enum`? I think an enum would be better for helping the user populate the channel option, because they could use `CompressionAlgorithm.gzip`, for example.

packages/grpc-js/src/server-call.ts
b0b3rt: True!

b0b3rt: True!

packages/grpc-js/src/compression-algorithms.ts
b0b3rt: Generally I disprefer enums in Typescript because they're one of the few language features which aren't runtime-only, and they interact somewhat poorly with the rest of the type system, since they're somewhat poorly supported and the language designers seem to pushing people in the direction of [using object literals (with `as const`) instead](https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums).

Those probably aren't huge concerns in this case; I don't see a way that making it an enum would frustrate a typical user of the library.  For user convenience and consistency with the way other such options are handled, I'll change it to an enum.

packages/grpc-js/src/compression-filter.ts
```diff
@@ -183,6 +232,19 @@ export class CompressionFilter extends BaseFilter implements Filter {
...
+
+    /* Check to see if the compression we're using to send messages is supported by the server
+     * If not, reset the sendCompression filter and have it use the default IdentityHandler */
+    const serverSupportedEncodingsHeader = metadata.get('grpc-accept-encoding')[0] as string | undefined;
+    if (serverSupportedEncodingsHeader) {
+      this.sharedFilterConfig.serverSupportedEncodingHeader = serverSupportedEncodingsHeader;
+      const serverSupportedEncodings = serverSupportedEncodingsHeader.split(',');
+
+      if ((this.sendCompression instanceof DeflateHandler && !serverSupportedEncodings.includes('deflate'))
+          || (this.sendCompression instanceof GzipHandler && !serverSupportedEncodings.includes('gzip'))) {
```
murgatroid99, line 244: I think this whole `if` statement can be reduced to `if (!serverSupportedEncodings.includes(this.currentCompressionAlgorithm))`, and it should also set `this.currentCompressionAlgorithm`.

packages/grpc-js/test/test-server.ts
```diff
@@ -604,3 +621,259 @@ describe('Generic client and server', () => {
...
+        ServerCredentials.createInsecure(),
+        (err, port) => {
+          assert.ifError(err);
+          server.start();
+          assignedPort = port;
+          client = new testServiceGrpcObject.TestService(
+            `localhost:${assignedPort}`,
+            grpc.credentials.createInsecure(),
+            {
+              'grpc.default_compression_algorithm': 1
```
murgatroid99, line 686: The test code should use the `CompressionAlgorithms` enum when setting this option, both for clarity and to demonstrate that the enum works as intended.

packages/grpc-js/src/compression-filter.ts
b0b3rt: Yep, simplified (and added).

packages/grpc-js/test/test-server.ts
b0b3rt: Good catch, thanks.

b0b3rt:
Hm, looks like the tests are [failing](https://source.cloud.google.com/results/invocations/f747ecd3-03bb-4f88-9647-42cbd9a1064c/targets/grpc%2Fnode%2Fpull_request%2Flinux/log):
```
Error: ENOENT: no such file or directory, open 'test/fixtures/test_service.proto'
...
    at Object.loadSync (/tmpfs/src/github/grpc-node/packages/grpc-js/node_modules/@grpc/proto-loader/build/src/index.js:194:31)
    at Object.<anonymous> (/tmpfs/src/github/grpc-node/packages/grpc-js/test/test-server.ts:38:44)
```

which is
`const loadedTestServiceProto = protoLoader.loadSync('test/fixtures/test_service.proto', {`

Since this passes locally but not in the test runner, does the test runner have something other than `grpc-node/packages/grpc-js` as the root dir when running the test?

murgatroid99:
It looks like the problem is that you didn't update the `includeDirs` list when you moved the proto file. The path that you load needs to be an absolute path or relative to one in that list. The fixtures directory gets copied over to the build directory, so you can just load `${__dirname}/fixtures/test_service.proto` without any include path.

b0b3rt:
Got it, changed the path accordingly.

packages/grpc-js/src/server-call.ts
```diff
@@ -439,6 +442,56 @@ export class Http2ServerCallStream<
...
+  
+      default: {
+        this.sendError({
+          code: Status.UNIMPLEMENTED,
+          details: `Received message compressed with unsupported encoding "${encoding}"`,
+        });
+        return Promise.resolve();
+      }
+    }
+  }
```
murgatroid99, line 493: I didn't notice this before, but this function has to actually remove the 5-byte chunk at the beginning of the message. That chunk contains the compressed flag and the length of the encoded message, which are both invalid and useless after the message has been decompressed. Then `deserializeMessage` also has to be modified to not slice off the first 5 bytes.

b0b3rt: @murgatroid99 It does remove the 5-byte chunk before decompressing it; do you mean to avoid re-attaching it before returning it?
```
              const joined = Buffer.concat([message.slice(0, 5), output]);
              resolve(joined);
```

Which would obviate the need to remove it again later:
```
  deserializeMessage(bytes: Buffer) {
    // TODO(cjihrig): Call compression aware deserializeMessage().
    const receivedMessage = bytes.slice(5);

    return this.handler.deserialize(receivedMessage);
  }
```

If that's what you meant, I'll make that change.

On a separate note, I'm not able to reproduce the test failures in `Linux Tests`.  When I run the tests locally (on Ubuntu 20.04.1 via WSL2), the following two tests fail:
```
  1) Name Resolver
       DNS Names
         Should resolve localhost properly:
     Error: Timeout of 4000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.

  2) Name Resolver
       DNS Names
         Should default to port 443:
     Error: Timeout of 4000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

However, those tests were failing from the moment I pulled down the repo; I assumed it was a quirk of the local environment (or some minor dependency mismatch somewhere).

murgatroid99: Yes, I meant to remove the 5 byte header from the output of that function, and the other code you referenced that removes it again.

Don't worry about those test failures. I see the same failures on WSL; it seems to be a quirk of how Windows or WSL resolves `localhost`.

b0b3rt: Sounds good, pushed a commit that simplifies the removal of the "prefix".

murgatroid99 (approved):
Thank you for your contribution, and for working with me to improve this PR.

desy83:
@murgatroid99 is there any approximate release date for the compression support? Thanks

murgatroid99:
I will probably be able to publish this in early January after the holidays.

murgatroid99:
This has been published in version 1.5.0.

PR #1959, opened 2021-11-03
grpc-js: Fix trailers-only response handling

murgatroid99:
We received a user report that when receiving a server stream with no messages from a Java server in particular, users would not get an `end` event from the call. Java servers send trailers-only responses, so this seems like a bug in the trailers-only response handling. The only thing I can think of causing this is the `ClientHttp2Stream` not emitting an `end` event when receiving a trailers-only response, causing the `readsClosed` flag to never be set and the `Http2CallStream` to never output a status, because of [this check](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/call-stream.ts#L343-L351). So, this change sets that flag and adds logging to the `ClientHttp2Stream` `end` event to verify what is happening.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/grpc-js",
-  "version": "1.4.2",
+  "version": "1.4.3",
   "description": "gRPC Library for Node - pure JS implementation",
   "homepage": "https://grpc.io/",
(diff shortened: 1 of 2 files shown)


nicolasnoble (approved):

murgatroid99:
I did not actually understand the user report in question correctly. That was not the problem, and this is not the solution.

PR #1981, opened 2021-12-06
Make changes to make all of the test jobs green

murgatroid99:
This skips three groups of tests:

 1. All tests against the native library on Node versions 14 and greater. The purpose of those tests is to test that grpc-js is interoperable and behaviorally consistent with the native library, not to test the function of the native library itself. That fails if the native library itself is not behaving correctly, so it's best to skip the tests where that is the case. We are still testing that in Node versions 10 and 12.
 2. The `NoCompress` flag test. We currently use the `encoding` argument of the `write` method to pass write flags in. This is not supported functionality, but it has worked for a long time. However, Node 16 apparently added validation for that argument, and any flags we pass in are not considered valid. So, that now does not work at all, and as a result we cannot pass in write flags at all. We will need to find a new way to do that, but for now we are skipping the test of the way it would work currently.
 3. The DNS resolver tests with the target names `loopback4.unittest.grpc.io` and `loopback46.unittest.grpc.io`. In the first case, since there is no IPv6 address, the Mac DNS resolver returns an IPv4-embedded IPv6 address, which is out of the scope of the test to test for. In the second case, that address should resolve to `127.0.0.1` and `::1`, and it works correctly on Linux. However, on Windows we only get the IPv4 address and on Mac we only get the IPv6 address, so there is nothing we can consistently test for on all platforms.

That revealed an actual bug: with the introduction of compressed message handling on the server, the message handling code now performs an asynchronous operation, which can result in reordering of messages, or messages with the end of stream event. The second commit corrects that.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -44,7 +44,7 @@
   ],
   "scripts": {
     "build": "npm run compile",
-    "clean": "node -e 'require(\"rimraf\")(\"./build\", () => {})'",
+    "clean": "rimraf ./build",
     "compile": "tsc -p .",
     "format": "clang-format -i -style=\"{Language: JavaScript, BasedOnStyle: Google, ColumnLimit: 80}\" src/*.ts test/*.ts",
     "lint": "npm run check",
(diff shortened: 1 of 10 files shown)


murgatroid99:
The remaining Mac failure is bizarre. The failing test is to have the DNS resolver resolve the target name `loopback4.unittest.grpc.io`. On Linux and Windows, it correctly returns the IP address `127.0.0.1`, but on Mac, it returns `64:ff9b::7f00:1`, which appears nowhere in the DNS record.

nicolasnoble (approved):

PR #1987, opened 2021-12-10
grpc-js: Send backoffOptions to BackoffTimeout

cosmin19:
Send the BackoffOptions to the BackoffTimeout just as in **subchannel.ts, line 221**. Without this code, **grpc.initial_reconnect_backoff_ms** and **grpc.max_reconnect_backoff_ms** are not used at all in the gRPC client and the retry will always use the default values (1000 and 120000)

### DIFF
--- a/packages/grpc-js/src/resolving-load-balancer.ts
+++ b/packages/grpc-js/src/resolving-load-balancer.ts
@@ -26,7 +26,7 @@ import { ConnectivityState } from './connectivity-state';
 import { ConfigSelector, createResolver, Resolver } from './resolver';
 import { ServiceError } from './call';
 import { Picker, UnavailablePicker, QueuePicker } from './picker';
-import { BackoffTimeout } from './backoff-timeout';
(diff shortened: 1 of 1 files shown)


murgatroid99:
Thank you for the PR. This looks like a good change. I am not allowed to merge it until you sign the CLA.

cosmin19:
I have signed the CLA but the PR still appears as unauthorized. What should I do next?

murgatroid99:
It looks like you are a different GitHub user than the one that created the commit. The CLA bot comment says that the commit is not authorized, which means that the CLA was not signed for that user. You can fix this by changing the commit to have your email as the committer, or by signing the CLA for that other GitHub user.

cloverheap:
Fixed

murgatroid99 (approved):
Thank you for your contribution. This will be published in an upcoming patch release soon.

murgatroid99:
I have published this in version 1.4.4

PR #1993, opened 2021-12-20
grpc-js: Set provided class options to generated class

DavyJohnes:
Currently neither `serviceName` nor `classOptions` are not used in `makeClientConstructor` at all. In some cases it is very handy to access at least `serviceName` in generated class to identify which service exactly generated class belongs to.

### DIFF
--- a/packages/grpc-js/src/make-client.ts
+++ b/packages/grpc-js/src/make-client.ts
@@ -91,6 +91,7 @@ export interface ServiceClientConstructor {
     options?: Partial<ChannelOptions>
   ): ServiceClient;
   service: ServiceDefinition;
+  serviceName: string;
 }
 
 /**
@@ -127,6 +128,7 @@ export function makeClientConstructor(
 
   class ServiceClientImpl extends Client implements ServiceClient {
     static service: ServiceDefinition;
+    static serviceName: string;
     [methodName: string]: Function;
   }
 
@@ -171,6 +173,7 @@ export function makeClientConstructor(
   });
 
   ServiceClientImpl.service = methods;
+  ServiceClientImpl.serviceName = serviceName;
 
   return ServiceClientImpl;
 }


DavyJohnes:
HI @murgatroid99,
Any chances this will be reviewed and merged soon? :)

murgatroid99:
I am on vacation. I will review this when I am back after the new year.

murgatroid99:
I don't think this is a good idea. The `options` argument originally existed on the `grpc` package to provide a single deprecated option that we did not carry over when creating this library. I don't think it should be repurposed like this.

However, I would be OK with attaching just the service name directly to the class. But, you should also look at the existing call to `makeClientConstructor` in `loadPackageDefinition` and determine whether the `serviceName` it is passing there is what you actually want to have access to. The service name was actually not used even in the original implementation of this function, so we can change what value we pass in without changing any semantics.

DavyJohnes:
> I don't think this is a good idea. The `options` argument originally existed on the `grpc` package to provide a single deprecated option that we did not carry over when creating this library. I don't think it should be repurposed like this.
> 
> However, I would be OK with attaching just the service name directly to the class. But, you should also look at the existing call to `makeClientConstructor` in `loadPackageDefinition` and determine whether the `serviceName` it is passing there is what you actually want to have access to. The service name was actually not used even in the original implementation of this function, so we can change what value we pass in without changing any semantics.

Ok, I've dropped changes regarding `options`. Currently I only set static `serviceName` options on a result class.
I've checked `loadPackageDefinition`, it uses key of `PackageDefinition` item (slightly transformed). I guess this is expected behaviour.

murgatroid99 (approved):

DavyJohnes:
@murgatroid99 are there any plans when this will be published to npm?

murgatroid99:
This missed the 1.5.x release which went out last week, so I was planning to release it in 1.6.x, which will probably be a little while.

DavyJohnes:
@murgatroid99 sorry for disturbing, is there any ETA of 1.6.x will be released? This blocks usage of our library in some cases.

murgatroid99:
I don't know exactly. Probably some time in February at the earliest.

DavyJohnes:
@murgatroid99 hi! Are there any updates about 1.6.X release?

DavyJohnes:
😢

murgatroid99:
I'm sorry about the delayed response, and the delayed release. We have published version 1.6.0.

PR #1996, opened 2021-12-27
Document grpc-js supported channel args in readme

### LINKED ISSUE #1982: grpc.lb_policy_name option no longer works in grpc-js
### Problem description
Setting the client load balancer to `round_robin` via the `'grpc.lb_policy_name'` option works in `grpc`, but seems to have no effect in `@grpc/grpc-js`.

However, setting the client load balancer via the `'grpc.service_config'` option works in both libraries, so this is not a blocker. Depending on the intention for `'grpc.lb_policy_name'`, maybe this is a bug or maybe this only needs to be documented as a breaking change to watch out for when switching over to grpc-js.

### Reproduction steps
```node
// const grpc = require('grpc'); // works (uses round_robin)
const grpc = require('@grpc/grpc-js'); // broken (uses pick_first)

// <get stub constructor via protoLoader.loadSync and grpc.loadPackageDefinition>

const options = {
    'grpc.ssl_target_name_override': ...,
    'grpc.lb_policy_name': 'round_robin', // <--- has no effect in grpc-js
    // 'grpc.service_config': 
    //     JSON.stringify({loadBalancingConfig: [{round_robin: {}}]}),  // <--- but this still works
};

const client = new MyService(
    'myservice.netflix.com:8980',
    grpc.credentials.createSsl(...),
    options
);

// <make a call with the client>
```

Setting env vars `GRPC_VERBOSITY=debug` and `GRPC_TRACE=round_robin,pick_first`, I can see that when running under `grpc` the round_robin LB is successfully activated, but under `@grpc/grpc-js` it still uses pick_first.

### Environment
 - OS name, version and architecture: Linux and macOS 
 - Node version: 12.22
 - Node installation method: NA
 - Package name and version: `grpc@v1.24.11` and `@grpc/grpc-js@v1.4.4`

josephharrington:
This copies the list of supported channel arguments from PACKAGE_COMPARISON.md into the readme, and also adds a link to the package comparison doc.

resolves #1982, resolves #1983


murgatroid99:
I am OK with this overall plan but I don't like how this list of supported channel options overwhelms the rest of the "Migrating from `grpc`" list. I would prefer to have a separate "Supported channel options" heading with this list and a brief explanation. Then the bullet point in the "Migrating from `grpc`" can have a brief explanation of what the user should do to migrate from `grpc` when they are using an unsupported channel option, including a reference to that other heading.

In order to avoid duplicating information that needs to be updated, I would prefer to also remove this information from the package comparison doc and replace it with a link to this README, saying something either about channel options in particular or API differences in general.

josephharrington:
@murgatroid99 Good points! Thanks for the feedback.

> a brief explanation of what the user should do to migrate from `grpc` when they are using an unsupported channel option

For this, I think the best recommendation would just be to remove usage of any unsupported channel options. Can you think of anything else they should do?

murgatroid99:
I don't think that really addresses the main issue. Removing unsupported options is not necessary because grpc-js just ignores unrecognized options, and if the option ever is supported in the future, it will have the same semantics. The more important thing is for users to be aware of options that are not supported, and adjust their own code to compensate. In your own example in #1982, your problem was not solved by removing the unsupported option, but by finding and using the alternative option that provided the desired functionality.

So, maybe it could say something like this:

> If you are using any channel options supported in `grpc` but not supported in `@grpc/grpc-js`, you may need to adjust your code to handle the different behavior. Refer to...

josephharrington:
@murgatroid99 Yep, agreed! I updated the change with the wording you suggested.

murgatroid99 (approved):

PR #2018, opened 2022-01-11
Add envoy/extensions files

sevaorlov:
Extension files are also referenced in the grpc-js-xds code https://github.com/grpc/grpc-node/blob/master/packages/grpc-js-xds/src/resources.ts#L68, so they need to be included when grpc-js-xds is installed as a dependency.


murgatroid99 (approved):
Thank you for fixing this.

murgatroid99:
This has been published in version 1.5.1

PR #2022, opened 2022-01-13
grpc-js: Don't use http_proxy for uds connections

raksooo:
The environment variable `http_proxy` will be used for UDS connections which prevents the connection from working. Since UDS runs on the local machine there's no use for a sending it through a proxy. This pull request adds a check which disables the http proxy if the target is a Unix Domain Socket.


murgatroid99 (approved):
This change looks good. I just can't merge it until you sign the CLA.

raksooo:
@murgatroid99 Great! It's signed now :+1:

murgatroid99:
This has been published in version 1.5.1.

Issue #2023, opened 2022-01-13
DNS storm with round robin load balancing in grpc-js


iscopetta:
### Problem description
We are having some problems with the load balancing in `grpc-js`.

We are seeing uneven distribution of calls to our service pods, which ends up sometimes overloading some of them, while keeping others at very low load.

We think this might be because of the default load balancing strategy of "pick first", so we tried enabling round robin but this caused a bunch of issues

- The distribution, albeit more consistent, was still not uniform
- The pods where the client was running started using 3 times the CPU and created a flurry of requests to our DNS

Any ideas how we could address this uneven distribution issue, and what could be wrong with load balancing? 

### Reproduction steps
Our (singleton) clients get instantiated with the DNS address of the service.
The DNS returns the IP of all the available pods for the given service.
We enable round robin load balancing by providing this configuration to the client: 
```
'grpc.service_config': JSON.stringify({ loadBalancingConfig: [{ round_robin: {} }], })
```
There was no other change to the clients besides the lb config.

### Environment
 - OS name, version and architecture: Debian GNU/Linux 10 (buster) x86
 - Node version: 14.17.6
 - Node installation method: yarn
 - Package name and version: 1.4.5

### Additional context
When we tried to deploy the mentioned config change this the behavior we saw:   
(the baseline is for ~100 pods, while the spike is for just 4 canary pods where a single client configuration was changed)
CPU:
<img width="1113" alt="CPU spike" src="https://user-images.githubusercontent.com/74217591/149392446-58ad29f8-d02b-4fa7-8b51-eadd6d918256.png">

DNS requests:  
<img width="1474" alt="Screen Shot 2022-01-13 at 11 01 04 AM" src="https://user-images.githubusercontent.com/74217591/149392700-5e928650-1822-4d46-83ed-05dbd892562d.png">

murgatroid99:
I think I see what is happening here: the clients are failing to connect to some of the addresses returned by the DNS. Those connection failures trigger DNS re-resolution attempts, which do not back off in this situation. The lack of a backoff here is a bug that I will fix. The connection failures would also explain the uneven request distribution.

You can get logs with more information about what is happening here by setting the environment variables `GRPC_TRACE=channel,round_robin,subchannel,dns_resolver` and `GRPC_VERBOSITY=DEBUG`.

iscopetta:
Will try to gather some extra logging like you suggested and see if we can add some information here, thank you!

What I am wondering is why would we have connection failures in the first place?
[...]

PR #2024, opened 2022-01-14
grpc-js: Add backoff to DNS resolution attempts

murgatroid99:
This change is to address the increased DNS resolution traffic when using round robin reported in #2023. With this change, any single channel will space out DNS requests based on the configured connection backoff settings.

### DIFF
--- a/packages/grpc-js/src/resolver-dns.ts
+++ b/packages/grpc-js/src/resolver-dns.ts
@@ -32,6 +32,7 @@ import { SubchannelAddress, TcpSubchannelAddress } from './subchannel-address';
 import { GrpcUri, uriToString, splitHostPort } from './uri-parser';
 import { isIPv6, isIPv4 } from 'net';
 import { ChannelOptions } from './channel-options';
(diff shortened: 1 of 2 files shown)


packages/grpc-js/src/resolver-dns.ts
```diff
@@ -119,6 +122,18 @@ class DnsResolver implements Resolver {
       details: `Name resolution failed for target ${uriToString(this.target)}`,
       metadata: new Metadata(),
     };
+    
+    const backoffOptions: BackoffOptions = {
+      initialDelay: channelOptions['grpc.initial_reconnect_backoff_ms'],
```
stanley-cheung, line 127: what if the key doesn't exist? is it oK?

stanley-cheung (approved):

packages/grpc-js/src/resolver-dns.ts
murgatroid99: Then the value will be `undefined`, so the `BackoffTimeout` will fall back to the default value, which is 1000 ms.

PR #2039, opened 2022-02-08
Add subchannel_settings tracer

nimf:
The subchannel_settings tracer adds HTTP/2 settings frame information to debug logs.

HTTP/2 settings frame contains important information like max_concurrent_streams and initial_window_size useful for debugging concurrency, latency, and throughput issues.

If it's ok we can also add that to the `subchannel` tracer and not introduce `subchannel_settings` at all.

### DIFF
--- a/packages/grpc-js/src/logging.ts
+++ b/packages/grpc-js/src/logging.ts
@@ -108,10 +108,12 @@ export function trace(
   tracer: string,
   text: string
 ): void {
-  if (
-    !disabledTracers.has(tracer) &&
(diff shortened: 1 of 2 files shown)


murgatroid99:
This should be fine in the main `subchannel` tracer, because settings frames should be infrequent. The refcount tracer is separate because the call refcount changes twice with every request, and that can drown out the rest of the subchannel trace logs in a high volume client.

nimf:
@murgatroid99 I put it in the `subchannel` tracer. This PR is ready to be reviewed.

PR #2058, opened 2022-03-07
grpc-js: Add outlier detection LB policy

murgatroid99:
This implements the non-xDS part of [gRFC A50](https://github.com/grpc/proposal/blob/master/A50-xds-outlier-detection.md). I also needed to add the subchannel interface and wrapper that that spec refers to. Other languages already had that.

The most important new code is in `load-balancer-outlier-detection.ts`.

Currently the only test is a simple regression test verifying that the client functions normally when the `outlier_detection` LB policy is configured. The test file has the machinery for other tests, but I don't know how to implement those yet. As mentioned in the spec, the ability to use the `outlier_detection` LB policy is disabled by default, and enabled only if the corresponding environment variable is set. That layer of protection will be removed when there is enough testing, which will include xDS tests.

### DIFF
--- a/packages/grpc-js/src/duration.ts
+++ b/packages/grpc-js/src/duration.ts
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2022 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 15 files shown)


packages/grpc-js/src/duration.ts
```diff
@@ -0,0 +1,36 @@
...
+    seconds: (millis / 1000) | 0,
+    nanos: (millis % 1000) * 1_000_000 | 0
+  };
+}
+
+export function durationToMs(duration: Duration): number {
+  return (duration.seconds * 1000 + duration.nanos / 1_000_000) | 0;
+}
+
+export function isDuration(value: any): value is Duration {
```
nicolasnoble, line 34: Maybe `value: object` here?

nicolasnoble:
I think this looks good to me, I think there's maybe a few locations where we could've used `object` instead of `any` for validation purposes? I've marked one, but there's a few others throughout the code. Dunno :-)

nicolasnoble (approved):

PR #2071, opened 2022-03-25
Expose MetadataOptions interface in grpc-js.

kskalski:
Since Metadata is exposed and its constructor takes MetadataOptions, those should also be exposed.


murgatroid99:
You will also need to update the `export` on line 122.

murgatroid99 (approved):

Issue #2072, opened 2022-03-26
Unhandled ERR_STREAM_WRITE_AFTER_END due possible no 'error' handler on ClientWritableStreamImpl


happy15:
### Problem description
Unhandled ERR_STREAM_WRITE_AFTER_END due possible no 'error' handler on ClientWritableStreamImpl.

### Reproduction steps
It happened on production and several node instances crash due this problem:
```
events.js:288
    throw er; // Unhandled 'error' event
    ^

Error [ERR_STREAM_WRITE_AFTER_END]: write after end
    at writeAfterEnd (_stream_writable.js:264:14)
    at ClientHttp2Stream.Writable.write (_stream_writable.js:313:5)
    at Http2CallStream.writeMessageToStream (/home/appops/approot/server/node_modules/@grpc/grpc-js/build/src/call-stream.js:332:26)
    at /home/appops/approot/server/node_modules/@grpc/grpc-js/build/src/call-stream.js:665:26
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
Emitted 'error' event on ClientWritableStreamImpl instance at:
    at errorOrDestroy (internal/streams/destroy.js:108:12)
    at onwriteError (_stream_writable.js:462:5)
    at onwrite (_stream_writable.js:483:5)
    at processTicksAndRejections (internal/process/task_queues.js:83:21) {
  code: 'ERR_STREAM_WRITE_AFTER_END'
}
```
Unfortunately It's unknown how to reproduce it. By checking grpc-js source code it seems there is no 'error' handler on ClientWritableStreamImpl.

### Environment
 - OS name, version and architecture: [e.g. Linux Ubuntu 18.04 amd64]
 - OS
Distributor ID: Debian
Description:    Debian GNU/Linux 9.1 (stretch)
Release:        9.1
x86_64
 - Node version v12.16.1
 - @grpc/grpc-js v1.5.8

### Additional context
There is a warning on event loop block before the crash:
```
(node:33) Warning: Event loop delay exceeded 1 hour: 6154818098838 nanoseconds
```

murgatroid99:
First, `ClientWritableStreamImpl` is the class of an object that is returned to you when you call a client streaming method. You should register an `error` handler on it so that you can handle whatever errors that may occur.

Second, are you sure that you are not calling `write` after calling `end` in your own interactions with that object? That is the general cause of that error.

[...]

PR #2073, opened 2022-03-28
grpc-js: Avoid surfacing errors without gRPC error codes

murgatroid99:
#2072 involves an error that has come up a couple of times before: `ERR_STREAM_WRITE_AFTER_END`, which is output by Node's Writable stream class when `write` is called after `end`. This has come up previously and my approach to addressing it has been focused on adding ordering restrictions to avoid triggering the error in the first place. This change instead insures that if the error does occur, we output it through the standard error output path with a gRPC error code. I choose the default `UNAVAILABLE` to match the synchronous error handling below at what is now line 866, but this particular error is more accurately `INTERNAL` because it is a violation of an API usage sequencing constraint.

### DIFF
--- a/packages/grpc-js/src/call-stream.ts
+++ b/packages/grpc-js/src/call-stream.ts
@@ -839,7 +839,16 @@ export class Http2CallStream implements Call {
       message,
       flags: context.flags,
     };
-    const cb: WriteCallback = context.callback ?? (() => {});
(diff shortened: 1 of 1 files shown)


nicolasnoble (approved):

nicolasnoble:
I'm not totally convinced this is the right play here, but I'm open to see how it turns out.

PR #2076, opened 2022-04-01
grpc-js: Add support for `grpc.dns_min_time_between_resolutions_ms` channel arg

murgatroid99:
This adds a delay after a successful DNS request before another DNS request can be started, configurable with the `grpc.dns_min_time_between_resolutions_ms` channel arg, defaulting to 30 seconds. This all matches the current behavior of the C core.

### DIFF
--- a/packages/grpc-js/src/resolver-dns.ts
+++ b/packages/grpc-js/src/resolver-dns.ts
@@ -45,6 +45,8 @@ function trace(text: string): void {
  */
 const DEFAULT_PORT = 443;
 
+const DEFAULT_MIN_TIME_BETWEEN_RESOLUTIONS_MS = 30_000;
+
 const resolveTxtPromise = util.promisify(dns.resolveTxt);
 const dnsLookupPromise = util.promisify(dns.lookup);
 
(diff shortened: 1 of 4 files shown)


packages/grpc-js/src/resolver-dns.ts
```diff
@@ -88,6 +96,8 @@ class DnsResolver implements Resolver {
   private defaultResolutionError: StatusObject;
   private backoff: BackoffTimeout;
   private continueResolving = false;
+  private nextResolutionTimer: NodeJS.Timer;
+  private isNextResolutionTimerRunning = false;
```
nicolasnoble, line 100: This almost feels like I'd want a special timer wrapper that'd add this property. Almost.

packages/grpc-js/src/resolver-dns.ts
```diff
@@ -301,9 +335,9 @@ class DnsResolver implements Resolver {
   }
 
   destroy() {
-    /* Do nothing. There is not a practical way to cancel in-flight DNS
```
nicolasnoble, line 304: This comment is interesting because in the meantime, I learned that there is, in fact, a way to cancel in-flight DNS requests in libuv. Funny they didn't necessarily expose this to the upper JavaScript?

http://docs.libuv.org/en/v1.x/request.html

`uv_cancel(uv_req*);`

nicolasnoble (approved):

packages/grpc-js/src/resolver-dns.ts
murgatroid99: If they wanted to do that they'd have to return some kind of handle object from `dns.lookup` that would exist solely for the unusual edge case that you want to cancel a DNS request.

packages/grpc-js/src/resolver-dns.ts
murgatroid99: `BackoffTimeout` actually does this, but we're not exponentially backing off here so it felt wrong to use it.

PR #2077, opened 2022-04-01
grpc-js-xds: Don't stop backoff timers for ADS streams

murgatroid99:
This means that the client will always back off from starting a new ADS stream by the base backoff time.

### DIFF
--- a/packages/grpc-js-xds/package.json
+++ b/packages/grpc-js-xds/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/grpc-js-xds",
-  "version": "1.6.0",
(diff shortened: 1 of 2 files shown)


murgatroid99:
This needs some more work, to change the behavior of `reset` so that it resets the currently running timer.

nicolasnoble (approved):

murgatroid99:
The most relevant change to backoff-timeout.ts is the code added to `reset`. As part of that, I factored out the method `runTimer` from `runOnce`. The rest is just added documentation.

nicolasnoble (approved):

PR #2087, opened 2022-04-08
grpc-js: Don't clear ping timeout when still connected

murgatroid99:
By default, we don't send pings when there are no active streams because some servers don't like that. But that doesn't mean that existing pending pings should be invalidated when the last stream ends. This change fixes that. I changed this line instead of changing `stopKeepalivePings` because there are two other call sites of `stopKeepalivePings` that should have the other behavior.

### DIFF
--- a/packages/grpc-js/src/subchannel.ts
+++ b/packages/grpc-js/src/subchannel.ts
@@ -384,6 +384,11 @@ export class Subchannel {
      * sending pings should also involve some network activity. */
   }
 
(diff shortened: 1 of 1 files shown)


nicolasnoble:
Might be worth adding a comment around `stopKeepalivePings` explaining the distinction.

nicolasnoble (approved):

nicolasnoble (approved):

PR #2096, opened 2022-04-15
Fix ts errors with typescript 4.6.3 #2107

jimmypresto:
This fixes couple typescript files failing to transpile with typescript 4.6.x. Typescript 4.6 enforces object's default type to be 'unknown', instead of 'any' and no implicit the argument value for the resolve call.

The transpilation errors under typescript 4.6:
a. Object is of type 'unknown'
b. Argument of type 'unknown' is not assignable to parameter of type XYZ
c. Expected 1 arguments, but got 0. Did you forget to include 'void' in your type argument to 'Promise'?

To repro:
1. cd packages/grpc-js
2. npm install
3. npm install typescript@4.6.3
4. npm run build

### DIFF
--- a/packages/grpc-js/src/call-credentials-filter.ts
+++ b/packages/grpc-js/src/call-credentials-filter.ts
@@ -54,7 +54,7 @@ export class CallCredentialsFilter extends BaseFilter implements Filter {
     const resultMetadata = await metadata;
     try {
       resultMetadata.merge(await credsMetadata);
-    } catch (error) {
+    } catch (error: any) {
       this.stream.cancelWithStatus(
         Status.UNAUTHENTICATED,
         `Failed to retrieve auth metadata with error: ${error.message}`
--- a/packages/grpc-js/src/call-credentials.ts
+++ b/packages/grpc-js/src/call-credentials.ts
@@ -115,7 +115,7 @@ export abstract class CallCredentials {
                 reject(err);
                 return;
               }
-              resolve(headers);
+              resolve(headers || {});
             }
           );
         });
--- a/packages/grpc-js/src/call-stream.ts
+++ b/packages/grpc-js/src/call-stream.ts
@@ -553,7 +553,7 @@ export class Http2CallStream implements Call {
           let metadata: Metadata;
           try {
             metadata = Metadata.fromHttp2Headers(headers);
-          } catch (error) {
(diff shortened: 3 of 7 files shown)


murgatroid99:
We don't currently use that version of the compiler, so these issues do not impact us. Why do you want to make this change?

jimmypresto:
I can live with other typescript 4.x versions you prefer. Any plans to move up the 4.x? I think it has been released more than a year.

murgatroid99:
Is there some specific reason that you want this package to switch to using TypeScript 4.x?

jimmypresto:
Apps with ts 4.x is unable to consume current grpc-js release build with ts 3.x implementation. Transpilation fails.

murgatroid99:
If you are having a problem with using grpc-js, please file an issue with the details of that problem.

jimmypresto:
The pr description has transpilation errors, repro steps and ts changes that builds. 

Sure. I can create a ticket.

murgatroid99:
I understand that this package does not build with a different compiler version. What I am asking for is an actual use case that is actually impacted by this incompatibility. If you `npm install` the library, can you use it?

jimmypresto:
It is when npm install a grpc-js build in any apps using ts 4.x, the `npm run build` would fail. 

As to the reasons of why I use TS 4.x:
1. TS 4.x language features. https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html
2. Better toolchain, perf and features.

murgatroid99:
I cannot reproduce this failure you are describing. Please provide a specific sequence of actions to reproduce it.

jimmypresto:
Can you do a "npm install" before the typescript 4.x install command? I have revised the steps as follows.

1. cd packages/grpc-js
2. npm install
3. npm install typescript@4.6.3
4. npm run build

murgatroid99:
You said

> when npm install a grpc-js build in any apps using ts 4.x, the `npm run build` would fail.

Cloning this repository and changing the compiler version is not the normal procedure to use this package in an app. Did you experience a problem with the `@grpc/grpc-js` package when installing it from the npm repository?

NoxSensei:
What about skipping publish of source files?
Won't it be enough to just publish `.d.ts` and `.js` files and just ignore `.ts` files?

PR #2110, opened 2022-05-02
proto-loader: Update to Long 5.x

### LINKED ISSUE #2009: [Bug]: 'Long' can only be imported by using a default import.
node_modules/@grpc/proto-loader/build/src/index.d.ts(22,10): error TS2595: 'Long' can only be imported by using a default import.

murgatroid99:
This fixes #2009, as far as I can tell. The `long` package exports the type using `export = Long;` which it says is "compatible with `import Long from "long"`", but as far as I can tell that is only true if we set the TypeScript option `"esModuleInterop": true`. Unfortunately, if we were to set that, it appears to generate code that transitively requires all of our users to also set that option to compile any code that depends on `@grpc/proto-loader`. The alternative is to use the [`import =` syntax](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require), which seems to work no matter how that option is set.


nicolasnoble (approved):

alexander-fenster:
Note: https://github.com/googleapis/gax-nodejs/issues/1240, https://github.com/googleapis/gax-nodejs/issues/1241. I will pin `@grpc/proto-loader` to `0.6.9` for now.

PR #2117, opened 2022-05-13
grpc-js-xds: Refactor xDS stream state and add resource timer

murgatroid99:
The main goal of this change is to add the resource timer, which is described in [this section of the xDS docs](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#knowing-when-a-requested-resource-does-not-exist). Essentially, the protocol does not consistently tell us when a resource does not exist, so we assume that the server will respond to any request within 15 seconds, and if the client does not get the requested resource in that time, it assumes that it does not exist. Since the timer is watching server response latency, we only run it while we have an active ADS stream. The specific algorithm is as follows:

 - When a watcher subscribes to a new resource name, if the ADS stream is alive, start the timer for that resource name.
 - When the ADS stream dies, cancel all timers.
 - When a new ADS stream starts, start the timer for all resource names for which we do not have a cached response.
 - When we receive a resource, cancel its timer.
 - When the timer fires, report that the resource does not exist.

Some significant changes were needed to the watchers structure to implement this in a reasonable way, and the simplest way to do that was to refactor most of the logic into the new `BaseXdsStreamState` abstract class. This refactor ended up slightly modifying the nonexistent resource reporting for LDS and CDS, to account for whether we already have a cached response. Otherwise I think it should all be functionally the same.

I also added a change to clear the nonce when the ADS stream ends, because apparently we're supposed to do that too. It's not really related to this change but I was touching that part of the code anyway and it's a quick change.

### DIFF
--- a/packages/grpc-js-xds/src/xds-client.ts
+++ b/packages/grpc-js-xds/src/xds-client.ts
@@ -309,7 +309,7 @@ export class XdsClient {
     const edsState = new EdsState(() => {
       this.updateNames('eds');
     });
-    const cdsState = new CdsState(edsState, () => {
+    const cdsState = new CdsState(() => {
       this.updateNames('cds');
     });
     const rdsState = new RdsState(() => {
@@ -630,6 +630,7 @@ export class XdsClient {
           this.updateNames(service);
         }
       }
+      this.reportAdsStreamStarted();
     }
   }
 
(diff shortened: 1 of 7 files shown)


murgatroid99:
The tests failed (https://source.cloud.google.com/results/invocations/08de3287-754d-4983-8a6d-d23120284303/targets), but I think it's actually a configuration issue unrelated to this change. The test logs show that the driver never successfully confirms that the servers start, and the client logs appear to show the correct general behavior, in terms of correctly processing the xDS responses.

I think it would be best to just submit and let it run with the regular config.

nicolasnoble (approved):

PR #2123, opened 2022-05-19
bugfix: handle disconnectListeners in reverse to allow listener removal in loop

bartslinger:
I had a situation with 1 subchannel for multiple streams. If a keepalive timeout occured, it would not trigger a 'Connection dropped' for each stream.

I found a bug in the code that handles the disconnect listeners. The listener function itself can [remove itself from the array](https://github.com/bartslinger/grpc-node/blob/0a0e13eedecedb31e91ad2d250c5fca6f31e46b6/packages/grpc-js/src/call-stream.ts#L334). The removal [uses splice](https://github.com/bartslinger/grpc-node/blob/0a0e13eedecedb31e91ad2d250c5fca6f31e46b6/packages/grpc-js/src/subchannel.ts#L981). Thereby, the next item shifts one place and will not be called from the loop. This is a common problem and is usually handled by looping in reverse.

### DIFF
--- a/packages/grpc-js/src/subchannel.ts
+++ b/packages/grpc-js/src/subchannel.ts
@@ -646,7 +646,8 @@ export class Subchannel {
     this.transitionToState(
       [ConnectivityState.READY],
       ConnectivityState.TRANSIENT_FAILURE);
-    for (const listener of this.disconnectListeners) {
+    for (let i = this.disconnectListeners.length - 1; i >= 0; i--) {
+      const listener = this.disconnectListeners[i];
       listener();
     }
   }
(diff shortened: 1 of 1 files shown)


murgatroid99:
I would prefer to avoid index-based iteration where possible, because it is more error-prone than a simple `for..of` loop. One change I have considered before that should also address this problem is representing `disconnectListeners` as a `Set` instead of an array, because the information I can find indicates that `Set` handles deletion during iteration safely.

If you would prefer not to make a larger change like that, another option would be to iterate over a copy of the array.

bartslinger:
Using a Set is a good suggestion. It is still a small change.

murgatroid99 (approved):

PR #2125, opened 2022-05-23
grpc-js-xds: Add support for k8s interop test framework

murgatroid99:
The test script is copied with slight modifications from [the Python test script](https://github.com/grpc/grpc/blob/master/tools/internal_ci/linux/grpc_xds_k8s_lb_python.sh). The docker image is large (a bit over 1GB), but when I tried to make it smaller with a build step like in [the Go Dockerfile](https://github.com/grpc/grpc-go/blob/master/interop/xds/client/Dockerfile), it caused the tests to fail and I couldn't tell why.


test/kokoro/xds_k8s_lb.cfg
```diff
@@ -0,0 +1,26 @@
...
+# Location of the continuous shell script in repository.
+build_file: "grpc-node/packages/grpc-js-xds/scripts/xds_k8s_lb.sh"
+timeout_mins: 180
+action {
+  define_artifacts {
+    regex: "artifacts/**/*sponge_log.xml"
+    regex: "artifacts/**/*sponge_log.log"
+    strip_prefix: "artifacts"
+  }
+}
```
sergiitk, line 26: Nit: missing new line at the EOF.

packages/grpc-js-xds/scripts/xds_k8s_lb.sh
```diff
@@ -0,0 +1,166 @@
...
+  for test in "${test_suites[@]}"; do
+    run_test $test || (( failed_tests++ ))
+  done
+  echo "Failed test suites: ${failed_tests}"
+  if (( failed_tests > 0 )); then
+    exit 1
+  fi
+}
+
+main "$@"
```
sergiitk, line 166: Nit: missing new line at the EOF.

packages/grpc-js-xds/scripts/xds_k8s_lb.sh
```diff
@@ -0,0 +1,166 @@
...
+main() {
+  local script_dir
+  script_dir="$(dirname "$0")"
+
+  # Source the test driver from the master branch.
+  echo "Sourcing test driver install script from: ${TEST_DRIVER_INSTALL_SCRIPT_URL}"
+  source /dev/stdin <<< "$(curl -s "${TEST_DRIVER_INSTALL_SCRIPT_URL}")"
+
+  activate_gke_cluster GKE_CLUSTER_PSM_SECURITY
+  activate_secondary_gke_cluster GKE_CLUSTER_PSM_SECURITY
```
sergiitk, line 143: I think this is in error in the python file, and this should be
```suggestion
  activate_gke_cluster GKE_CLUSTER_PSM_LB
  activate_secondary_gke_cluster GKE_CLUSTER_PSM_LB
```

Similar to cpp: https://github.com/grpc/grpc/blob/master/tools/internal_ci/linux/grpc_xds_k8s_lb.sh

packages/grpc-js-xds/interop/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+# Make a grpc-node directory and copy the repo into it.
+WORKDIR /node/src/grpc-node
+COPY . .
+
+WORKDIR /node/src/grpc-node/packages/grpc-js
+RUN npm install
+WORKDIR /node/src/grpc-node/packages/grpc-js-xds
+RUN npm install
+
+ENTRYPOINT [ "node", "/node/src/grpc-node/packages/grpc-js-xds/build/interop/xds-interop-client" ]
```
sergiitk, line 30: I think the issue was with using `WORKDIR` multiple times. WORKDIR sets the path relative to which the rest of commands are executed. Consider this:

```suggestion
RUN cd packages/grpc-js && npm install
RUN cd packages/grpc-js-xds && npm install

ENTRYPOINT [ "node", "packages/grpc-js-xds/build/interop/xds-interop-client" ]
```

sergiitk (commented):
The image size over 1 GB is pretty surprising. I'd recommend investing a bit more time into reducing the size. I'll give it a try and report back.

sergiitk:
@murgatroid99 Take a look at this article on building node images: https://docs.docker.com/language/nodejs/build-images/

AFAIK `RUN npm install --production` and https://docs.docker.com/language/nodejs/build-images/#create-a-dockerignore-file should reduce the size at least partially.

murgatroid99:
I did manage to decrease the Docker image size with a build step. I tried again and it worked, and I'm not sure what I did differently, but I'll take it.

packages/grpc-js-xds/interop/Dockerfile
murgatroid99: I am pretty sure that this doesn't matter now that my entrypoint is an absolute path.

packages/grpc-js-xds/interop/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+FROM node:16-alpine as build
+
+# Make a grpc-node directory and copy the repo into it.
+WORKDIR /node/src/grpc-node
+COPY . .
+
+WORKDIR /node/src/grpc-node/packages/grpc-js
+RUN npm install
+WORKDIR /node/src/grpc-node/packages/grpc-js-xds
+RUN npm install
```
sergiitk, line 28: I still recommend against using `WORKDIR` for just just changing a folder. Note that it's also creating a folder if it doesn't exist.

```suggestion
RUN cd packages/grpc-js && npm install
RUN cd packages/grpc-js-xds && npm install
```

sergiitk (approved):
LGTM. Thanks for the improvement.

packages/grpc-js-xds/interop/Dockerfile
sergiitk: https://docs.docker.com/engine/reference/builder/#workdir

murgatroid99:
For some reason, that last commit breaks the tests, so I am undoing it.

murgatroid99:
Passing test run against this branch: https://source.cloud.google.com/results/invocations/b8fc66d3-1931-46b6-aec4-40d6f72f93b0/targets

nicolasnoble (approved):

PR #2127, opened 2022-05-24
Fallback for building on m1

fundthmcalculus:


murgatroid99:
This won't work. This package doesn't have a `binding.gyp` file, and the installed package does not include the files necessary to build it. `node-pre-gyp` is only used to fetch the platform-appropriate binary.

PR #2132, opened 2022-06-01
Expose path and callEnd event in ServerSurfaceCall

ajmath:
I'm working on some server interceptors that are injected as part of our generation process.  These interceptors need to know the final gRPC status and the path (really the service and function) that the request was sent to.  In the current model, both of these bits of information are hidden within the internal call implementation. This PR exposes both of these so that they can be accessed without crossing the internal implementation boundary.


ajmath:
@murgatroid99 any thoughts on this?

murgatroid99 (approved):

PR #2138, opened 2022-06-15
xDS k8s interop script: publish version-tagged docker images

murgatroid99:
These images are used in the staging tests, which don't build docker images themselves. The code for this is copied with slight modification from https://github.com/grpc/grpc/blob/1be6e2c9ebfb4a26bebe6b3f3c45cffc70e71b68/tools/internal_ci/linux/psm-security.sh#L46-L51

### DIFF
--- a/packages/grpc-js-xds/scripts/xds_k8s_lb.sh
+++ b/packages/grpc-js-xds/scripts/xds_k8s_lb.sh
@@ -45,9 +45,8 @@ build_test_app_docker_images() {
     -t "${CLIENT_IMAGE_NAME}:${GIT_COMMIT}" \
     .
 
-  if [[ -n $KOKORO_JOB_NAME ]]; then
-    branch_name=$(echo "$KOKORO_JOB_NAME" | sed -E 's|^grpc/node/([^/]+)/.*|\1|')
-    tag_and_push_docker_image "${CLIENT_IMAGE_NAME}" "${GIT_COMMIT}" "${branch_name}"
+  if is_version_branch "${TESTING_VERSION}"; then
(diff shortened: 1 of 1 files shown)


packages/grpc-js-xds/scripts/xds_k8s_lb.sh
```diff
@@ -45,6 +45,11 @@ build_test_app_docker_images() {
     -t "${CLIENT_IMAGE_NAME}:${GIT_COMMIT}" \
     .
 
+  if [[ -n $KOKORO_JOB_NAME ]]; then
+    branch_name=$(echo "$KOKORO_JOB_NAME" | sed -E 's|^grpc/node/([^/]+)/.*|\1|')
+    tag_and_push_docker_image "${CLIENT_IMAGE_NAME}" "${GIT_COMMIT}" "${branch_name}"
+  fi
```
sergiitk, line 51: New better way to do it:

```suggestion
  if is_version_branch "${TESTING_VERSION}"; then
    tag_and_push_docker_image "${CLIENT_IMAGE_NAME}" "${GIT_COMMIT}" "${TESTING_VERSION}"
  fi
```

sergiitk: (introduced in https://github.com/grpc/grpc/pull/30027)

murgatroid99: I made a single change to update both parts of the code that would use `TESTING_VERSION`.

sergiitk (approved):

PR #2141, opened 2022-06-23
Add xDS k8s url_map test script

murgatroid99:
The script is a copy of https://github.com/grpc/grpc/blob/master/tools/internal_ci/linux/grpc_xds_url_map_python.sh, with the variables changed.


packages/grpc-js-xds/scripts/xds_k8s_url_map.sh
```diff
@@ -0,0 +1,152 @@
+#!/usr/bin/env bash
+# Copyright 2021 gRPC authors.
```
sergiitk, line 2:
```suggestion
# Copyright 2022 gRPC authors.
```

nit

sergiitk (approved):

PR #2143, opened 2022-06-24
xDS k8s tests: fix a line in the url_map test script

murgatroid99:
The line copied from the Python script was hardcoded and not handled by a variable, so I did not change it on the first pass.

### DIFF
--- a/packages/grpc-js-xds/scripts/xds_k8s_url_map.sh
+++ b/packages/grpc-js-xds/scripts/xds_k8s_url_map.sh
@@ -40,7 +40,7 @@ build_test_app_docker_images() {
 
   pushd "${SRC_DIR}"
   docker build \
-    -f src/python/grpcio_tests/tests_py3_only/interop/Dockerfile.client \
+    -f ${BUILD_APP_PATH} \
     -t "${CLIENT_IMAGE_NAME}:${GIT_COMMIT}" \
     .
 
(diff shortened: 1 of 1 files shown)


packages/grpc-js-xds/scripts/xds_k8s_url_map.sh
```diff
@@ -40,7 +40,7 @@ build_test_app_docker_images() {
 
   pushd "${SRC_DIR}"
   docker build \
-    -f src/python/grpcio_tests/tests_py3_only/interop/Dockerfile.client \
+    -f ${BUILD_APP_PATH} \
```
sergiitk, line 43: [ShellCheck: SC2086 – Double quote to prevent globbing and word splitting.](https://www.shellcheck.net/wiki/SC2086)

```suggestion
    -f "${BUILD_APP_PATH}" \
```

sergiitk (changes requested):

sergiitk (approved):

sergiitk (approved):
nit: Just noticed that `BUILD_APP_PATH` points to the docker file itself. It'll make more sense to either set it to `packages/grpc-js-xds/interop`, and then provide the argument as `-f "${BUILD_APP_PATH}/Dockerfile"`, rename the variable to something like `CLIENT_DOCKERFILE_PATH`. Or maybe just remove the variable and inline the value. This applies to the other buildscript too (xds_k8s_lb.sh).

But this is optional, and can be done in another PR. This is good to go.

PR #2144, opened 2022-06-24
xDS k8s tests: fix docker auth error in the build scripts

sergiitk:
Docker auth needed to be done before pushing thats to the GCR.
Also add missing versioned image tagging to xds_k8s_url_map.sh.


murgatroid99 (approved):
The missing push of the versioned docker image was in the python url_map script I copied. But I don't object to changing it here.

PR #2145, opened 2022-06-28
Remove `grpc` dependency

mic-max:
Hard code the `NOT_FOUND` status code from the `grpc` library (deprecated).

The enum `grpc.status` has not been changed in 5 years from viewing the git blame and IMO likely won't. Further, when people download this package from NPM they could see a warning in the console saying that a package has been deprecated since `grpc-health-check` relies on `grpc`.

---

Actually, can't fully remove the `grpc` dependency from the `package.json` since the https://github.com/grpc/grpc-node/blob/master/packages/grpc-health-check/v1/health_grpc_pb.js uses it for the `makeGenericClientConstructor` function - which doesn't look like it would be a simple copy to this repository if that would even be desirable.

### DIFF
--- a/packages/grpc-health-check/health.js
+++ b/packages/grpc-health-check/health.js
@@ -18,13 +18,14 @@
 
 'use strict';
 
-var grpc = require('grpc');
-
 var _get = require('lodash.get');
 var _clone = require('lodash.clone')
 
 var health_messages = require('./v1/health_pb');
 var health_service = require('./v1/health_grpc_pb');
+
+// https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/constants.ts
+const NOT_FOUND_STATUS_CODE = 5;
 
 function HealthImplementation(statusMap) {
   this.statusMap = _clone(statusMap);
@@ -38,8 +39,7 @@ HealthImplementation.prototype.check = function(call, callback){
   var service = call.request.getService();
   var status = _get(this.statusMap, service, null);
   if (status === null) {
-    // TODO(murgatroid99): Do this without an explicit reference to grpc.
-    callback({code:grpc.status.NOT_FOUND});
+    callback({ code: NOT_FOUND_STATUS_CODE });
   } else {
     var response = new health_messages.HealthCheckResponse();
     response.setStatus(status);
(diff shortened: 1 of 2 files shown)


murgatroid99:
This is something I've been meaning to address for a while but I haven't had the chance. I think the right way to do this is to entirely remove the `grpc` dependency by removing the `Client` export. However, this has knock-on effects. The `service` export also comes from `health_grpc_pb.js`, and we want to keep the `service` export. So, I think the simplest solution is to switch protobuf implementations to `@grpc/proto-loader`, which would require rewriting some of the implementation code. That would also result in removing the `messages` export.

mic-max:
> This is something I've been meaning to address for a while but I haven't had the chance. I think the right way to do this is to entirely remove the `grpc` dependency by removing the `Client` export. However, this has knock-on effects. The `service` export also comes from `health_grpc_pb.js`, and we want to keep the `service` export. So, I think the simplest solution is to switch protobuf implementations to `@grpc/proto-loader`, which would require rewriting some of the implementation code. That would also result in removing the `messages` export.

Removing the `Client` export would require a major version bump I guess right? It's not something that my program uses at all so I'd be fine with that change 😅

> switch protobuf implementations to `@grpc/proto-loader`

Would this mean the `./v1/` directory be deleted and the `.proto` parsing and code generation would be done at application startup? Also, how would I test/compile this project - I might work on this improvement and would want the next PR update to be working - this one I just did by hand on GitHub.com 😅

murgatroid99:
> Removing the Client export would require a major version bump I guess right? It's not something that my program uses at all so I'd be fine with that change 😅

Yes, the change I suggest would need to be a major version bump. Practically speaking, it's just one line in either gRPC implementation to create the `Client` class from the `service`.

> Would this mean the ./v1/ directory be deleted and the .proto parsing and code generation would be done at application startup?

Yes, that's what I would do.

> Also, how would I test/compile this project - I might work on this improvement and would want the next PR update to be working - this one I just did by hand on GitHub.com 😅

The existing test setup will also have to be updated to handle not having the `grpc` dependency. I would add a dev dependency `"@grpc/grpc-js": "../grpc-js"` and change the import. The test code also uses proto messages, so that code will also need to be updated to handle the new code generation.

PR #2155, opened 2022-07-07
Bump protobuf dep to v21.1

rhinodavid:
Update `grpc-tool`'s `protobuf` submodule to `protobuf` [v21.2](https://github.com/protocolbuffers/protobuf/releases/tag/v21.2).

Currently, installing the `grpc-tools` package fails on M1/M2 Macs because previous versions of `protobuf` did not have precompiled ARM binaries. This was fixed [in an April release](https://github.com/protocolbuffers/protobuf/releases/tag/v3.20.1).


murgatroid99:
We don't use the `protobuf` precompiled binaries. And even if we did, that wouldn't solve the whole problem because we also need to build our own binary for the plugin that lives in this repo that generates the service code. In addition, Protobuf v21 removed all JavaScript generation code from the main protoc compiler and moved it to a plugin in https://github.com/protocolbuffers/protobuf-javascript. So, if we actually want to upgrade to protobuf v21, we will have to rework the build here to handle that separate plugin. And based on the issues in that repo, it seems like that plugin may not even work.

rhinodavid:
Gotcha -- thanks for the feedback.

I'm just really struggling to figure out how to get gRPC working with an M1 Mac. Seems like all the examples (including one I made in 2020) are broken and it's tough finding a path forward.

PR #2156, opened 2022-07-07
Enable xDS tracing in k8s framework tests

murgatroid99:
I noticed that this is the strategy the C++ interop docker image uses to enable tracing.


sergiitk (approved):

sergiitk:
As long as node knows how to handle these env vars.

Issue #2170, opened 2022-07-21
Missing source maps in `proto-loader` npm releases


chm-diederichs:
Hi I'm running into a DevTools error regarding missing source maps in the `proto-loader` package.

DevTools expects the following source maps: `build/src/index.js.map` & `build/src/util.js.map` to be present, however `.map` files are not included in the release.

Would it be possible to do either of the following:

1) Include `.map` files in the npm release

2) set `compilerOptions.sourceMap = false` so that `// # sourceMappingURL`s are omitted and DevTools does not load the missing map files

Thanks!

chm-diederichs:
For convenience, I have opened a PR for disabling source maps by default.

Let me know f you are interested/have any feedback.

PR #2171, opened 2022-07-21
Disable source maps in build

chm-diederichs:
This PR disables source maps by default when building `grpc-js` and `proto-loader` packages, by setting `compilerOptions.sourceMap = false` in the packages' `tsconfig.json` files.

Since `.map` files are not bundled in the `npm` release, this would solve missing source map errors when using the package. (See issue #2170)

Previous `tsconfig.json` files are moved to `tsconfig.dev.json` and a new script: `npm run dev` is added that performs a build with source maps.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -46,6 +46,7 @@
     "build": "npm run compile",
     "clean": "rimraf ./build",
     "compile": "tsc -p .",
+    "dev": "tsc -p . --project tsconfig.dev.json",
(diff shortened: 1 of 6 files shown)


murgatroid99:
Source maps *are* included in the npm release of `@grpc/grpc-js`. The error is that they aren't included in the `@grpc/proto-loader` package. That is a one-line fix.

chm-diederichs:
@murgatroid99 understood, thanks for pointing that out. See #2172 for the correct fix.

PR #2172, opened 2022-07-21
include `.map` files in proto-loader npm package

### LINKED ISSUE #2170: Missing source maps in `proto-loader` npm releases

chm-diederichs:
This PR updates `proto-loader` package.json to include `.js.map` files.

Fixes #2170

### DIFF
--- a/packages/proto-loader/package.json
+++ b/packages/proto-loader/package.json
@@ -38,7 +38,7 @@
   "files": [
     "LICENSE",
     "build/src/*.d.ts",
-    "build/src/*.js",
+    "build/src/*.{js,js.map}",
     "build/bin/*.js"
   ],
   "bin": {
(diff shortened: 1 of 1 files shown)


murgatroid99:
Looking at this again, I notice that `build/bin/proto-loader-gen-types.js` also has a `sourceMappingURL` comment, so you should also update line 42.

murgatroid99 (approved):

PR #2179, opened 2022-08-04
grpc-js: Implement getConnectivityState in subchannel wrapper

murgatroid99:
The `ejected` state overrides the reported connectivity state. The existing implementation implements that correctly when handling the connectivity state update event, but it does not implement `getConnectivityState`, so that falls back to the default implementation which just proxies the result from the `childSubchannel`. This matters because the `round_robin` LB policy uses `getConnectivityState` instead of just tracking the state update events.


gnossen (approved):
:+1:

PR #2183, opened 2022-08-05
proto-loader-gen-types Typename templates

install:
- Allow for customizing the naming pattern for both restricted and permissive types

Enums and filenames are left untouched

### DIFF
--- a/packages/proto-loader/bin/proto-loader-gen-types.ts
+++ b/packages/proto-loader/bin/proto-loader-gen-types.ts
@@ -26,12 +26,25 @@ import * as yargs from 'yargs';
 import camelCase = require('lodash.camelcase');
 import { loadProtosWithOptions, addCommonProtos } from '../src/util';
 
+const templateStr = "%s";
+const useNameFmter = ({outputTemplate, inputTemplate}: GeneratorOptions) => {
+  if (outputTemplate === inputTemplate) {
+    throw new Error('inputTemplate and outputTemplate must differ')
+  }
+  return {
+    outputName: (n: string) => outputTemplate.replace(templateStr, n),
+    inputName: (n: string) => inputTemplate.replace(templateStr, n)
+  };
+}
+
 type GeneratorOptions = Protobuf.IParseOptions & Protobuf.IConversionOptions & {
   includeDirs?: string[];
   grpcLib: string;
   outDir: string;
   verbose?: boolean;
   includeComments?: boolean;
+  inputTemplate: string;
+  outputTemplate: string;
 }
 
 class TextFormatter {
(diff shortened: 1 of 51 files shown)


murgatroid99:
Can you expand some more on what your use case is for this, and also why you excluded enums?

install:
@murgatroid99 
I've found that for the majority of my use cases I'm consuming the `__Output` types, so it'd be nice if the consumer can determine the naming format for inputs vs outputs.

The enum is a special case because it's both a type and a value. Parent messages seem to reference the enum directly.

```ts
export interface TestMessage {
  'enum'?: (test_v1_Enum | keyof typeof test_v1_Enum);
}

export interface TestMessage__Output {
  'enum': (keyof typeof test_v1_Enum);
}

```

It might be worth treating enums like messages in that we create both an input and an output version, and the input/output parent message would point to the corresponding one. I wanted to leave that for a different PR for now.

install:
Hey @murgatroid99 have you had a chance to look at this? I'm open to any feedback and suggestions :) thanks

packages/proto-loader/bin/proto-loader-gen-types.ts
```diff
@@ -114,15 +127,16 @@ function getTypeInterfaceName(type: Protobuf.Type | Protobuf.Enum | Protobuf.Ser
...
   const filePath = from === undefined ? './' + getImportPath(dependency) : getRelativeImportPath(from, dependency);
+  const {outputName, inputName} = useNameFmter(options);
   const typeInterfaceName = getTypeInterfaceName(dependency);
   let importedTypes: string;
   /* If the dependency is defined within a message, it will be generated in that
    * message's file and exported using its typeInterfaceName. */
   if (dependency.parent instanceof Protobuf.Type) {
     if (dependency instanceof Protobuf.Type) {
-      importedTypes = `${typeInterfaceName}, ${typeInterfaceName}__Output`;
+      importedTypes = `${typeInterfaceName}, ${outputName(typeInterfaceName)}`;
```
murgatroid99, line 139: Should the first instance of `typeInterfaceName` on this line be `inputName(typeInterfaceName)`?

packages/proto-loader/bin/proto-loader-gen-types.ts
```diff
@@ -805,7 +826,11 @@ async function runScript() {
         case 'String': return String;
         default: return undefined;
       }
-    }).alias({
+    })
+    .string(['inputTemplate', 'outputTemplate'])
```
murgatroid99, line 830: Please move this into the existing `string` call earlier, and group the new `default` calls with the others.

murgatroid99 (changes requested):
I am not confident that `inputName` is called everywhere it is needed. Since the default behavior is to leave the string unchanged, the current tests will not catch if it is omitted anywhere. Can you add a test with a different `--inputTemplate` option?

Please also create a PR to add information about these new options to [this file](https://github.com/grpc/proposal/blob/master/L70-node-proto-loader-type-generator.md).

install:
@murgatroid99 Thanks for the feedback! I've addressed your comments and updated the golden tests to use the new flags. 
Here's the corresponding proposal PR: https://github.com/grpc/proposal/pull/326

packages/proto-loader/bin/proto-loader-gen-types.ts
install: Good catch. Updating the golden tests found a couple other places that are now covered

snarky-puppy:
FYI - prefixing interfaces with "I" conflicts with Hungarian notation commonly misused in other parts of the TS ecosystem. Suggest prefixing with "In"/"Out" rather than "I/O"

install:
@murgatroid99 How can I help push this forward?

murgatroid99:
Sorry about the delay here.

murgatroid99:
This is out in version 0.7.3

Issue #2184, opened 2022-08-06
Leaked ClientHttp2Stream when canceling duplex streams


Matt-Esch:
I am upgrading a codebase from native grpc to grpc-js. When using grpc-js I see a memory leak affecting duplex streams which is caused by unclosed `ClientHttp2Stream` instances on the caller side of duplex streams. My best guess at the cause of the problem is that the client cancels the request and the underlying `ClientHttp2Stream` waits indefinitely (or until the socket is closed) for trailers that never arrive. So while the the leak appears on the client side, it's possibly caused by buggy server side behavior.

  - Is there an option to configure how long to wait for trailers?
  - Any suggestions for debugging if the server did send trailers, and if not why not?
 
I don't have a deterministic reproduction, some fraction of calls leak when cancelling them, so I don't know what state the connection needs to be in to cause it to leak.

### Environment
 - Ubuntu 20.04.4 LTS (GNU/Linux 5.13.0-1019-aws aarch64)
 - Node v12.22.6 and v16.16.0
 - @grpc/grpc-js@1.6.8

### Additional context
Related to https://github.com/grpc/grpc-node/issues/1339

Heapdump shows leaked `Http2CallStreams` (shown with distance -)
<img width="1217" alt="image" src="https://user-images.githubusercontent.com/1058839/183240009-e255ba8b-e5b4-4b29-976b-edd28d9555f6.png">

`Http2CallStream` instances are leaked through the bound `this` on the the `ClientHttp2Stream` trailers event listener
<img width="1214" alt="image" src="https://user-images.githubusercontent.com/1058839/183240067-4b1a100b-6733-4f5c-ae27-982d1092ea40.png">

We can see that the calls were cancelled on the client side looking at the `finalStatus`
<img width="378" alt="image" src="https://user-images.githubusercontent.com/1058839/183240507-70f7ec70-4388-42f2-b3c3-dacb3f9902dc.png">

Matt-Esch:
@murgatroid99 I can't see how https://github.com/grpc/grpc-node/pull/2187 would cause the listener to not be holding a reference, what am I missing? I understand it's worthwhile to be consistent with the other listeners, but wouldn't it be better to replace `.on` with `.once` so that the listener is removed once the trailers are received? This would only go so far because the retainer is ultimately the `ClientHttp2Stream` which maintains a list of all of the open streams.

[...]

PR #2190, opened 2022-08-08
grpc-js: Destroy http2 stream after closing

murgatroid99:
This will probably fix the `ClientHttp2Stream` leak reported in #2184. `close` tries to close an http2 stream, but it waits for incoming data to be read before finalizing that destruction. `destroy` guarantees destruction without waiting for data to be read. We still want to call close to chose the RST_STREAM code to send. By my own reading of that part of the Node source code, even if `destroy` is called before the RST_STREAM is sent, it will ensure that the RST_STREAM is sent, and it will use the code set by the call to `close`.


gnossen (approved):

murgatroid99:
This is not the right approach. It broke a lot of tests.

PR #2193, opened 2022-08-12
grpc-js: Drain incoming http2 data after outputting status

murgatroid99:
This is an attempt to fix the `ClientHttp2Stream` leak described in #2184. The likely cause of the leak is that `stream.close()` is called but it still has incoming data waiting to be read so it is never destroyed. This fixes that by putting the stream into flowing mode and discarding all incoming messages after outputting a status. This should have no user-visible effect because the existing code already does not output any messages after outputting a status.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/grpc-js",
-  "version": "1.6.9",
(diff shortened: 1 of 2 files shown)


gnossen (approved):

MercanAda:
I observe that someone leaking through my mobile while nearby apps during using. But who is?

PR #2192, opened 2022-08-09
fix: update `long` to v5

SimenB:
`protobufjs@7` (#2167) comes with it, so aligning seems sensible to me

PR #2197, opened 2022-08-17
proto-loader: Undo upgrade of 'long' dependency

murgatroid99:
This reverts #2192. We thought it wouldn't break Cloud API client libraries, but it did. We'll try this again in the future after they have made some changes.


gnossen (approved):

SimenB:
Is there an issue we can track for "they have made some change"?

murgatroid99:
That statement was the result of an internal conversation. I am not aware of a corresponding issue.

PR #2213, opened 2022-08-30
Fix `grpc-message` typo in call-stream.ts

clww:
In PR https://github.com/grpc/grpc-node/pull/2210, error handling code was added for `decodeURI(metadataMap['grpc-message'])`, but there is a typo in the `catch` block where `grpc-message` was written as `grpc-messages` with an extra "**s**".

### DIFF
--- a/packages/grpc-js/src/call-stream.ts
+++ b/packages/grpc-js/src/call-stream.ts
@@ -491,7 +491,7 @@ export class Http2CallStream implements Call {
       try {
         details = decodeURI(metadataMap['grpc-message']);
       } catch (e) {
-        details = metadataMap['grpc-message'] as string;
+        details = metadataMap['grpc-message'];
       }
       metadata.remove('grpc-message');
       this.trace(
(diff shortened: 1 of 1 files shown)


packages/grpc-js/src/call-stream.ts
```diff
@@ -491,7 +491,7 @@ export class Http2CallStream implements Call {
       try {
         details = decodeURI(metadataMap['grpc-message']);
       } catch (e) {
-        details = metadataMap['grpc-messages'] as string;
+        details = metadataMap['grpc-message'] as string;
```
murgatroid99, line 494:
```suggestion
        details = metadataMap['grpc-message'];
```

The `as string` type assertion is only needed because the key didn't match the one in the type check in the surrounding `if` statement, so it can be removed.

murgatroid99 (changes requested):
Thank you for the correction. I can't merge this unless you sign the CLA.

clww:
The EasyCLA status hasn't been updated after I signed (probably related to multiple email associtaed with my GH account?). Could you proceed with another PR? I can close this PR.

murgatroid99:
Unfortunately, the change needs to be against the v1.6.x branch for me to publish it, and switching the branch target for this PR pulls in the commits from master. I will make the change myself in #2216.

murgatroid99:
I published the updated fix in version 1.6.12

clww:
thank you!

PR #2221, opened 2022-09-07
grpc-js-xds: Enable outlier detection by default

murgatroid99:
I updated the check in the same way as the previous line, and I updated the change I previously made in #2212 to be the same as this one.


sergiitk (approved):
Makes sense.

PR #2226, opened 2022-09-12
grpc-js: Restrict control-plane status codes

murgatroid99:
This is an implementation of [gRFC A54](https://github.com/grpc/proposal/blob/master/A54-restrict-control-plane-status-codes.md).

### DIFF
--- a/packages/grpc-js/src/channel.ts
+++ b/packages/grpc-js/src/channel.ts
@@ -68,6 +68,28 @@ function getNewCallNumber(): number {
   return callNumber;
 }
 
+const INAPPPROPRIATE_CONTROL_PLANE_CODES: Status[] = [
+  Status.OK,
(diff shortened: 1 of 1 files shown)


packages/grpc-js/src/channel.ts
```diff
@@ -68,6 +68,28 @@ function getNewCallNumber(): number {
   return callNumber;
 }
 
+const INAPPPROPRIATE_CONTROL_PLANE_CODES: Status[] = [
```
gnossen, line 71: Nit: 3 'P's.

gnossen (approved):

packages/grpc-js/src/channel.ts
murgatroid99: Fixed

Issue #2231, opened 2022-09-20
Performance loss since @grpc/grpc-js version 1.6.8


antonstefer:
### Problem description
I use @google-cloud/bigtable which depends on @grpc/grpc-js through grpc-gcp and google-gax (maybe in more ways). When updating from 1.6.7 to 1.6.8 or any version above that, bigtable calls (especially ReadRows queries) get significantly slower (about 100% on average).

### Reproduction steps
Use bigtable with both of the above versions and perform ReadRows queries on tables.

### Environment
 - Google Cloud, Kubernetes Cluster, Container-Optimized OS from Google 
 - node:16-alpine docker image
 - @grpc/grpc-js@1.6.8 or above
 - google managed bigtable instance

### Additional context
As this is only a performance issue we do not see any error logs. We have created trace analysis reports using the google cloud platform which show the difference in execution time.

before:
<img width="935" alt="before" src="https://user-images.githubusercontent.com/59652072/191322163-bc8878e9-3394-4684-9f83-4f8defe78b81.png">

after:
<img width="935" alt="after" src="https://user-images.githubusercontent.com/59652072/191322159-2759f75a-b68f-4a46-9b60-9be2b75f74cf.png">

murgatroid99:
I published grpc-js version 1.7.1 with a change that will hopefully improve this situation. Please try it out.

antonstefer:
It works, thank you so much!

PR #2232, opened 2022-09-20
grpc-js: Defer evaluating caller stack until an error

murgatroid99:
This is an update to the change I made in #2109 that added caller stack traces to call errors. It turns out that getting the `stack` field of an error is a relatively expensive operation, so we don't do it in the hot path, and instead we wait until an error actually occurs. The goal is to fix the performance issues reported in #2231. #2109 is the only change that looks like it could possibly impact latency between 1.6.7 and 1.6.8.

### DIFF
--- a/packages/grpc-js/src/client.ts
+++ b/packages/grpc-js/src/client.ts
@@ -108,6 +108,10 @@ export type ClientOptions = Partial<ChannelOptions> & {
   callInvocationTransformer?: CallInvocationTransformer;
 };
 
+function getErrorStackString(error: Error): string {
+  return error.stack!.split('\n').slice(1).join('\n');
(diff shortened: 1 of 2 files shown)


packages/grpc-js/src/client.ts
```diff
@@ -340,6 +340,7 @@ export class Client {
         receivedStatus = true;
         if (status.code === Status.OK) {
           if (responseMessage === null) {
+            const callerStack = callerStackError.stack!.split('\n').slice(1).join('\n');
```
gnossen, line 343: Nit: This is repeated quite a bit. Maybe factor out a function?

gnossen (approved):

PR #2244, opened 2022-10-12
grpc-js-xds: Drop xDS v2 support

murgatroid99:
This deletes all of the code for handling v2 and all of the branches for different v2/v3 behavior, and renames a variety of symbols to not specify "v3".


sergiitk (approved):
LGTM

PR #2249, opened 2022-10-17
fix: track aborted status, do not stringify for tracing

AVVS:
I've been working on improving perf of grpc-js and a few things stood out to me:

1. Each `Http2ServerCallStream` has an empty timer created, which results in overhead, I assume this was done in an effort to have the same backing map, but performance is gained when not cleaning / creating timeouts needlessly
2. Track `aborted` event and in case stream is `aborted` emit `cancelled`, otherwise graceful stream end shouldn't cause what ultimately is exception. This allows for improved performance around cancellation logic
3. Wrap call to tracing as JSON.stringify + toJSON on metadata is an expensive operation and is performed regardless of tracing actually be on

Locally I've had all the tests pass on node 18, fingers crossed they pass everywhere else. I'm still mostly concerned about (2) as I don't yet fully grasp the logic around this call (checked blame and it was like that for 3 years)

Anyhow, hopefully these changes are useful or could be a starting base for further improvements

### DIFF
--- a/packages/grpc-js/src/server-call.ts
+++ b/packages/grpc-js/src/server-call.ts
@@ -443,9 +443,6 @@ export class Http2ServerCallStream<
           metadata: null,
         });
       }
-
-      // to compensate for a fact that cancelled is not always called
-      this.emit('close');
     });
 
     this.stream.on('drain', () => {
@@ -776,9 +773,6 @@ export class Http2ServerCallStream<
       call.cancelled = true;
       call.emit('cancelled', reason);
     });
-
-    // to compensate for the fact that cancelled is no longer always called
-    this.once('close', () => call.emit('close'))
 
     this.once('callEnd', (status) => call.emit('callEnd', status));
   }
--- a/packages/grpc-js/src/server-call.ts
+++ b/packages/grpc-js/src/server-call.ts
@@ -19,8 +19,9 @@ import { EventEmitter } from 'events';
 import * as http2 from 'http2';
 import { Duplex, Readable, Writable } from 'stream';
 import * as zlib from 'zlib';
+import { promisify } from 'util';
 
-import { Deadline, StatusObject } from './call-stream';
+import { Deadline, StatusObject, PartialStatusObject } from './call-stream';
 import {
   Status,
   DEFAULT_MAX_SEND_MESSAGE_LENGTH,
@@ -35,6 +36,8 @@ import { ChannelOptions } from './channel-options';
 import * as logging from './logging';
 
 const TRACER_NAME = 'server_call';
+const unzip = promisify(zlib.unzip);
+const inflate = promisify(zlib.inflate);
 
 function trace(text: string): void {
   logging.trace(LogVerbosity.DEBUG, TRACER_NAME, text);
@@ -86,25 +89,22 @@ export type ServerSurfaceCall = {
 export type ServerUnaryCall<RequestType, ResponseType> = ServerSurfaceCall & {
   request: RequestType;
 };
-export type ServerReadableStream<
-  RequestType,
-  ResponseType
-> = ServerSurfaceCall & ObjectReadable<RequestType>;
-export type ServerWritableStream<
-  RequestType,
-  ResponseType
-> = ServerSurfaceCall &
-  ObjectWritable<ResponseType> & {
-    request: RequestType;
-    end: (metadata?: Metadata) => void;
-  };
+export type ServerReadableStream<RequestType, ResponseType> =
+  ServerSurfaceCall & ObjectReadable<RequestType>;
+export type ServerWritableStream<RequestType, ResponseType> =
+  ServerSurfaceCall &
+    ObjectWritable<ResponseType> & {
+      request: RequestType;
+      end: (metadata?: Metadata) => void;
+    };
 export type ServerDuplexStream<RequestType, ResponseType> = ServerSurfaceCall &
   ObjectReadable<RequestType> &
   ObjectWritable<ResponseType> & { end: (metadata?: Metadata) => void };
 
 export class ServerUnaryCallImpl<RequestType, ResponseType>
   extends EventEmitter
-  implements ServerUnaryCall<RequestType, ResponseType> {
+  implements ServerUnaryCall<RequestType, ResponseType>
+{
   cancelled: boolean;
 
   constructor(
@@ -136,7 +136,8 @@ export class ServerUnaryCallImpl<RequestType, ResponseType>
 
 export class ServerReadableStreamImpl<RequestType, ResponseType>
   extends Readable
-  implements ServerReadableStream<RequestType, ResponseType> {
+  implements ServerReadableStream<RequestType, ResponseType>
+{
   cancelled: boolean;
 
   constructor(
@@ -178,7 +179,8 @@ export class ServerReadableStreamImpl<RequestType, ResponseType>
 
 export class ServerWritableStreamImpl<RequestType, ResponseType>
   extends Writable
-  implements ServerWritableStream<RequestType, ResponseType> {
+  implements ServerWritableStream<RequestType, ResponseType>
+{
   cancelled: boolean;
   private trailingMetadata: Metadata;
 
@@ -257,7 +259,8 @@ export class ServerWritableStreamImpl<RequestType, ResponseType>
 
 export class ServerDuplexStreamImpl<RequestType, ResponseType>
   extends Duplex
-  implements ServerDuplexStream<RequestType, ResponseType> {
+  implements ServerDuplexStream<RequestType, ResponseType>
+{
   cancelled: boolean;
   private trailingMetadata: Metadata;
 
@@ -395,7 +398,8 @@ export class Http2ServerCallStream<
   ResponseType
 > extends EventEmitter {
   cancelled = false;
-  deadlineTimer: NodeJS.Timer = setTimeout(() => {}, 0);
+  deadlineTimer: NodeJS.Timer | null = null;
+  private statusSent = false;
   private deadline: Deadline = Infinity;
   private wantTrailers = false;
   private metadataSent = false;
@@ -428,10 +432,20 @@ export class Http2ServerCallStream<
           ' stream closed with rstCode ' +
           this.stream.rstCode
       );
-      this.cancelled = true;
-      this.emit('cancelled', 'cancelled');
-      this.emit('streamEnd', false);
-      this.sendStatus({code: Status.CANCELLED, details: 'Cancelled by client', metadata: new Metadata()});
+
+      if (!this.statusSent) {
+        this.cancelled = true;
+        this.emit('cancelled', 'cancelled');
+        this.emit('streamEnd', false);
+        this.sendStatus({
+          code: Status.CANCELLED,
+          details: 'Cancelled by client',
+          metadata: null,
+        });
+      }
+
+      // to compensate for a fact that cancelled is not always called
+      this.emit('close');
     });
 
     this.stream.on('drain', () => {
@@ -444,9 +458,6 @@ export class Http2ServerCallStream<
     if ('grpc.max_receive_message_length' in options) {
       this.maxReceiveMessageSize = options['grpc.max_receive_message_length']!;
     }
-
-    // Clear noop timer
-    clearTimeout(this.deadlineTimer);
   }
 
   private checkCancelled(): boolean {
@@ -458,52 +469,22 @@ export class Http2ServerCallStream<
     return this.cancelled;
   }
 
-  private getDecompressedMessage(message: Buffer, encoding: string) {
-    switch (encoding) {
-      case 'deflate': {
-        return new Promise<Buffer | undefined>((resolve, reject) => {
-          zlib.inflate(message.slice(5), (err, output) => {
-            if (err) {
-              this.sendError({
-                code: Status.INTERNAL,
-                details: `Received "grpc-encoding" header "${encoding}" but ${encoding} decompression failed`,
-              });
-              resolve();
-            } else {
-              resolve(output);
-            }
-          });
-        });
-      }
-  
-      case 'gzip': {
-        return new Promise<Buffer | undefined>((resolve, reject) => {
-          zlib.unzip(message.slice(5), (err, output) => {
-            if (err) {
-              this.sendError({
-                code: Status.INTERNAL,
-                details: `Received "grpc-encoding" header "${encoding}" but ${encoding} decompression failed`,
-              });
-              resolve();
-            } else {
-              resolve(output);
-            }
-          });
-        });
-      }
-
-      case 'identity': {
-        return Promise.resolve(message.slice(5));
-      }
-  
-      default: {
-        this.sendError({
-          code: Status.UNIMPLEMENTED,
-          details: `Received message compressed with unsupported encoding "${encoding}"`,
-        });
-        return Promise.resolve();
-      }
+  private getDecompressedMessage(
+    message: Buffer,
+    encoding: string
+  ): Buffer | Promise<Buffer> {
+    if (encoding === 'deflate') {
+      return inflate(message.subarray(5));
+    } else if (encoding === 'gzip') {
+      return unzip(message.subarray(5));
+    } else if (encoding === 'identity') {
+      return message.subarray(5);
     }
+
+    return Promise.reject({
+      code: Status.UNIMPLEMENTED,
+      details: `Received message compressed with unsupported encoding "${encoding}"`,
+    });
   }
 
   sendMetadata(customMetadata?: Metadata) {
@@ -518,12 +499,21 @@ export class Http2ServerCallStream<
     this.metadataSent = true;
     const custom = customMetadata ? customMetadata.toHttp2Headers() : null;
     // TODO(cjihrig): Include compression headers.
-    const headers = Object.assign({}, defaultResponseHeaders, custom);
+    const headers = { ...defaultResponseHeaders, ...custom };
     this.stream.respond(headers, defaultResponseOptions);
   }
 
   receiveMetadata(headers: http2.IncomingHttpHeaders) {
     const metadata = Metadata.fromHttp2Headers(headers);
+
+    if (logging.isTracerEnabled(TRACER_NAME)) {
+      trace(
+        'Request to ' +
+          this.handler.path +
+          ' received headers ' +
+          JSON.stringify(metadata.toJSON())
+      );
+    }
 
     // TODO(cjihrig): Receive compression metadata.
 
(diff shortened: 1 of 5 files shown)


packages/grpc-js/src/server-call.ts
```diff
@@ -526,7 +531,9 @@ export class Http2ServerCallStream<
   receiveMetadata(headers: http2.IncomingHttpHeaders) {
     const metadata = Metadata.fromHttp2Headers(headers);
 
-    trace('Request to ' + this.handler.path + ' received headers ' + JSON.stringify(metadata.toJSON()));
+    if (logging.isTracerEnabled('trace')) {
```
murgatroid99, line 534: The tracer name for this file is `server_call`, not `trace`. This is stored in the `TRACER_NAME` constant in this file, so use `logging.isTracerEnabled(TRACER_NAME)`.

packages/grpc-js/src/server-call.ts
```diff
@@ -396,7 +396,8 @@ export class Http2ServerCallStream<
   ResponseType
 > extends EventEmitter {
   cancelled = false;
-  deadlineTimer: NodeJS.Timer = setTimeout(() => {}, 0);
+  deadlineTimer: NodeJS.Timer | null = null
```
murgatroid99, line 399: The whole point of having a noop timer instead of `null` here is to simplify the types, to avoid needing to check `if (this.deadlineTimer)` when acting on it. What exactly is the improvement gained by making this change?

packages/grpc-js/src/server-call.ts
```diff
@@ -422,17 +423,24 @@ export class Http2ServerCallStream<
...
           this.handler?.path +
           ' stream closed with rstCode ' +
           this.stream.rstCode
       );
-      this.cancelled = true;
-      this.emit('cancelled', 'cancelled');
-      this.emit('streamEnd', false);
-      this.sendStatus({code: Status.CANCELLED, details: 'Cancelled by client', metadata: new Metadata()});
+
+      if (this.aborted) {
```
murgatroid99, line 438: I don't see the reason to have this. At a high level, the purpose of the `cancelled` event is to notify the server that it no longer needs to work on this operation because the client no longer needs the result. That is just as true if the TCP connection disappeared as if the client explicitly sent a cancellation.

murgatroid99 (changes requested):

packages/grpc-js/src/server-call.ts
AVVS: Point is to increase performance. by creating an empty timer we allocate underlaying resources that need to be further cleaned up. This is in the hot path and by removing the empty timer I was able to get ~5% perf increase on unary request throughput

packages/grpc-js/src/server-call.ts
AVVS: `close` event will fire on stream regardless of whether it was cancelled or gracefully closed. performance benefits of this is especially noticeable on unary requests - we are avoiding emitting multiple events and triggering associated events for respectful listeners.

for instance, one may have an abort controller listen to 'cancelled' events in case of server or client initiated stream cancels to schedule cleanup events, but the way it stands right now - regardless of how the request has finished the `cancelled` event is always fired and needless work would be triggered - as it stands right now `cancelled` is renamed `close` event, which isn't surfaced

so ideally cancelled would have it's own semantics to notify the user about a cancelled (meaning unfinished) stream and you could have `close` being emitted alongside of it to have the old semantics

AVVS:
to give more perspective - grpc-js unary typically benches around 23-25k rps for unary calls (I know it's not the best use of it and streams are much more efficient) with 1.7ms avg latency and wildly ranging min/max latency between 0.4ms to 25ms(! loop being blocked by GC and other native side code), but when compared to fastify being able to handle around 100k rps on my machine with avg latency of 0 and max latency of 0.1ms there is a lot of room for improvements

packages/grpc-js/src/server-call.ts
AVVS: This is using `nice-grpc` wrapper around `@grpc/grpc-js`:

Screen Shot 2022-10-17 at 15 42 41

This is how things look like with "selective" `cancelled` event:

Screen Shot 2022-10-17 at 15 47 11

```
ghz --insecure --proto ./src/hello.proto --call helloworld.Greeter.SayHello -d '{"who":"nodeconf"}' -n 300000 --connections 1 --count-errors --timeout 0 0.0.0.0:8080

Summary:
  Count:	300000
  Total:	10.66 s
  Slowest:	22.14 ms
  Fastest:	0.12 ms
  Average:	1.43 ms
  Requests/sec:	28151.58

Response time histogram:
  0.125  [1]      |
  2.326  [271660] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  4.527  [25860]  |∎∎∎∎
  6.728  [1801]   |
  8.929  [311]    |
  11.131 [82]     |
  13.332 [135]    |
  15.533 [16]     |
  17.734 [65]     |
  19.935 [19]     |
  22.136 [50]     |

Latency distribution:
  10 % in 0.85 ms
  25 % in 1.08 ms
  50 % in 1.26 ms
  75 % in 1.52 ms
  90 % in 2.28 ms
  95 % in 2.66 ms
  99 % in 4.20 ms

Status code distribution:
  [OK]   300000 responses
```

And this is how they look like with always enabled `cancelled` event:

Screen Shot 2022-10-17 at 15 49 16

```
ghz --insecure --proto ./src/hello.proto --call helloworld.Greeter.SayHello -d '{"who":"nodeconf"}' -n 300000 --connections 1 --count-errors --timeout 0 0.0.0.0:8080

Summary:
  Count:	300000
  Total:	10.94 s
  Slowest:	23.66 ms
  Fastest:	0.15 ms
  Average:	1.50 ms
  Requests/sec:	27428.62

Response time histogram:
  0.147  [1]      |
  2.498  [281007] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  4.850  [16552]  |∎∎
  7.201  [1660]   |
  9.553  [431]    |
  11.904 [160]    |
  14.256 [32]     |
  16.607 [7]      |
  18.959 [35]     |
  21.311 [15]     |
  23.662 [100]    |

Latency distribution:
  10 % in 0.98 ms
  25 % in 1.16 ms
  50 % in 1.33 ms
  75 % in 1.59 ms
  90 % in 2.24 ms
  95 % in 2.62 ms
  99 % in 4.45 ms

Status code distribution:
  [OK]   300000 responses
```

I understand that this isn't strictly grpc-js making things slow, but being unable to differentiate whether `cancelled` event is an actual ungraceful termination from either end or a graceful one makes it hard to create further conditional logic

murgatroid99: I think the right approach here, instead of looking for the `aborted` event, is to track whether the stream has finished gracefully and avoid emitting the `cancelled` event if it has. I think the right thing to check for here is whether the server has sent the status (i.e. whether `sendStatus` has been called).

AVVS: definitely not opposed to the idea, though I believe we need to take into account both client and server cancelling the call, this is a little beyond my current understanding of the codebase and would likely require quite a bit more time to ensure nothing breaks, though if you can point me to specific places where such tracking makes sense it would be great!

murgatroid99: My suggestion is just to add a boolean field (call it `statusSent` or something) to `Http2ServerCallStream`, with the default value `false`, and set it to `true` in `sendStatus`, and change this line to `if (!statusSent)`.

AVVS: Done what we've discussed in the new commit

AVVS:
In the latest commit there are several pieces that has been refactors:

1. channelz - select associated session stream handler on startup, this removes a lot of checks we either know to be true or false ahead of time. some code is duplicated, but not too much
2. commented out validate checks on get / remove - respective keys would not be there in the first place, no reason to validate as it's also in a hot path with plenty of .remove calls
3. receiveUnaryMessage refactored to have fewer microtasks queued and in case compression isnt used completely avoid it. this is a little messy as we introduce a few try/catch blocks and work on a promise directly in case of decompression, but errors are handled correctly and overall its much more performant

with everything in here I managed to get this to about 43k RPS with unary requests reliably

```
ghz --insecure --proto ./src/hello.proto --call helloworld.Greeter.SayHello -d '{"who":"nodeconf"}' -n 3000000 --connections 2 -c 400 --count-errors --timeout 0 0.0.0.0:8080 --cpus=8 --skipFirst=10000

Summary:
  Count:	2990000
  Total:	68.92 s
  Slowest:	37.58 ms
  Fastest:	0.09 ms
  Average:	7.02 ms
  Requests/sec:	43383.58

Response time histogram:
  0.094  [1]      |
  3.842  [135888] |∎∎∎∎∎∎∎∎∎∎∎∎∎
  7.590  [427175] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  11.338 [375947] |∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎
  15.087 [49140]  |∎∎∎∎∎
  18.835 [10126]  |∎
  22.583 [789]    |
  26.331 [526]    |
  30.079 [11]     |
  33.827 [183]    |
  37.575 [214]    |

Latency distribution:
  10 % in 3.38 ms
  25 % in 5.13 ms
  50 % in 7.17 ms
  75 % in 8.89 ms
  90 % in 10.45 ms
  95 % in 11.77 ms
  99 % in 15.43 ms

Status code distribution:
  [OK]   2990000 responses
```

Associated flame graph:

Screen Shot 2022-10-18 at 13 30 45

There are 2 areas of concern: 
* Node.js `stream.respond` is relatively slow
* Lots of transforms between http2 headers <-> grpc.Metadata

Not sure what exactly could be done here except for trying out micro optimizations

AVVS:
One of the tests failed due to `node:` prefix, reverted this

AVVS:
in terms of `switch` vs `if` I believe if is faster up to 10 clauses based on what I've heard, but that shouldn't matter much here, still used that

packages/grpc-js/src/server.ts
```diff
@@ -724,150 +733,233 @@ export class Server {
...
+        channelzSessionInfo.messagesSent += 1;
+        channelzSessionInfo.lastMessageSentTimestamp = new Date();
+      });
+      call.on('receiveMessage', () => {
+        channelzSessionInfo.messagesReceived += 1;
+        channelzSessionInfo.lastMessageReceivedTimestamp = new Date();
+      });
+    }
+
+    const metadata = call.receiveMetadata(headers);
```
murgatroid99, line 847: As far as I can tell, starting at this point this function is the same as `_streamHandler`, except for the last branch of the `if` statements. I would prefer that this section of code be factored into a single function that both handler functions use. That last branch is not expected to be reachable in normal operation so I think it would be reasonable for that new function to conditionally perform the channelz operations in that branch at call handling time.

packages/grpc-js/src/server.ts
```diff
@@ -165,6 +170,10 @@ export class Server {
     if (this.channelzEnabled) {
       this.channelzTrace.addTrace('CT_INFO', 'Server created');
     }
+
+    this._channelzHandler = this._channelzHandler.bind(this)
+    this._streamHandler = this._streamHandler.bind(this)
```
murgatroid99, line 175: I don't like this. There's a kind of spooky action at a distance here, where the references to these functions in `_setupHandlers` work differently than normal because these lines are here. I would prefer to do this explicitly at that point where these functions need to be referenced.

murgatroid99 (changes requested):

packages/grpc-js/src/server-call.ts
```diff
@@ -559,52 +544,89 @@ export class Http2ServerCallStream<
...
+    const limit = this.maxReceiveMessageSize;
+
+    stream.on('data', onData);
+    stream.on('end', onEnd);
+    stream.on('error', onEnd);
+
+    function onData(chunk: Buffer) {
+      receivedLength += chunk.length;
+
+      if (receivedLength > limit) {
```
murgatroid99, line 568: This doesn't handle the case where the limit is `-1`, which means "unlimited". You can see how this is handled on line 577 of the current implementation.

packages/grpc-js/src/server-call.ts
```diff
@@ -559,52 +544,89 @@ export class Http2ServerCallStream<
...
+    const call = this;
+    const body: Buffer[] = [];
+    const limit = this.maxReceiveMessageSize;
+
+    stream.on('data', onData);
+    stream.on('end', onEnd);
+    stream.on('error', onEnd);
+
+    function onData(chunk: Buffer) {
+      receivedLength += chunk.length;
```
murgatroid99, line 566: This changes `byteLength` to `length`. I'm not sure exactly what the difference is, but I think this may be causing the out of bounds access error in one of the tests.

packages/grpc-js/src/server-call.ts
AVVS: yeah, already noticed the bug I've created - will add `receivedLength !== -1`

murgatroid99: No, it should be `limit !== -1`.

AVVS: right, of course :)

AVVS:
@murgatroid99 updated based on the suggestions, lets see how the tests do!

AVVS:
seems like its crashing at `const compressed = requestBytes.readUInt8(0) === 1;` meaning message is empty, will see what changed and why this isnt handled anymore

AVVS:
Seems like tests are generally ok, in terms of test targets - kind of odd that its running node 10/12 as they are both EOL. Any plans to update that?

murgatroid99:
Those test results look pretty good. The Windows failure is old and unrelated to this change. #2246 is an open PR to test on more recent versions. Also, it's only Windows that currently doesn't test on any version newer than 12.

If you would like this to be published as a patch release on 1.7.x, you'll have to rebase it onto the 1.7.x branch and then you or I can change the branch target. Otherwise it looks good to go.

AVVS:
I have one more experiment with https://www.npmjs.com/package/reusify - its used in underlaying modules for fastify and should yield significant improvements if I can wrap Metadata in a way that doesn't break things. It's kind of tricky to get this done as I don't always know when it's safe to release an object. Maybe would be a good idea to have non-wrapped implementation for client calls and wrapper implementation for server side as we can generally release once the request has ended

I'll work on that tomorrow and will rebase to 1.7.x in either case

AVVS:
Screen Shot 2022-10-18 at 16 51 04

if you look at the bottom right - this is with kernel tracing on - you can see how much time GC is taking

murgatroid99:
It is never safe to release objects that have been passed to user application code. We never know if they have stored it for later reference, and there is no API for them to release it.

murgatroid99:
There is quite a bit of potential complexity here (How should we implement this? Should it only cover `Metadata`? Should we also consider the client? etc.), so we should definitely address that in a different PR. It would also be a good idea to have a [gRFC proposal](https://github.com/grpc/proposal) to flesh out the details, because it seems like this will require a new API somewhere.

AVVS:
eh gotta rebase now ❌

murgatroid99:
I don't think you want to keep that merge commit. It's probably not going to make the rebase easier.

AVVS:
eh, had to squash it all up. there is a new part around `.emit('close')` + `.sendStatus` accepting null as Metadata. This is to avoid creating these objects as there were quite a few places where they aren't needed.

packages/grpc-js/src/server-call.ts
```diff
@@ -86,25 +89,22 @@ export type ServerSurfaceCall = {
...
-> = ServerSurfaceCall & ObjectReadable<RequestType>;
-export type ServerWritableStream<
-  RequestType,
-  ResponseType
-> = ServerSurfaceCall &
-  ObjectWritable<ResponseType> & {
-    request: RequestType;
-    end: (metadata?: Metadata) => void;
-  };
+export type ServerReadableStream<RequestType, ResponseType> =
```
AVVS, line 92: this got reformatted, I think its prettier from the project that forced the styling

packages/grpc-js/src/server-call.ts
```diff
@@ -428,10 +432,20 @@ export class Http2ServerCallStream<
...
+        this.emit('streamEnd', false);
+        this.sendStatus({
+          code: Status.CANCELLED,
+          details: 'Cancelled by client',
+          metadata: null,
+        });
+      }
+
+      // to compensate for a fact that cancelled is not always called
+      this.emit('close');
```
murgatroid99, line 448: I don't understand this change. I am not aware of any code that is listening for this `close` event. In addition, in what situation would an application care about this event if it has already sent a status.

AVVS: Screen Shot 2022-10-19 at 15 21 00

Screen Shot 2022-10-19 at 15 21 28

AVVS: Reverted this and if needed we can do a follow up on this - so it's just a separate unrelated issue

murgatroid99: If you're handling a stream of messages from the remote end, you should be listening for the `end` event to know when you have received all messages. And conversely, if you are sending a stream of messages, you should call the `end` event to indicate that you have finished writing. You shouldn't generally need to listen for the `cancelled` event unless you are doing a long asynchronous operation. I don't know why you would need to delete the `call` at all.

AVVS:
Its a long async operation. Clients can disconnect at any moment due to network change, app being closed, etc.

And once client disconnects stream needs to be cleaned up. There is never an ‘end’ event from either side, always abrupt disconnect that needs to be cleaned up

murgatroid99:
In that case you should still always see the `cancelled` event after your change.

Anyway, it looks like the PR is in a good state now, and the tests are passing. Is there anything else you would like to do before we merge it?

AVVS:
LGTM in this current state ;) if you could add `hacktoberfest-accepted` label after that - would be grand, too. t-shirts <3

murgatroid99 (approved):
Thank you for the contribution. I am happy to see some performance improvements in this library.

Regarding Hacktoberfest, repositories opt in to that event this year, and this repository was not opted in until just now, so you almost didn't get credit for this PR. According to the website, you don't actually need that label. It is also sufficient for the PR to be approved or merged.

AVVS:
@murgatroid99 pretty much everything is in there, you are likely going to be the most interested in this:

https://github.com/AVVS/talks/blob/main/nodeconfeu2022/scaling-nodejs/src/grpc-raw.ts

for serialization I used ts-proto + buf to generate (commited to repo), for flamegraphs I'm using clinic.js - there are a bunch of scripts you can meddle with (ie enable / disable kernel tracing)

for load producer I've used ghz, but I think you can also use something like nghttp2, but its a little bit more involved

PR #2259, opened 2022-11-03
grpc-js: Update to newest typescript compiler

murgatroid99:
A recent change to `@types/lodash` broke compatibility with TypeScript 3.9, because that version is outside of the support range. Future changes to types packages will probably cause this again, so the best remedy is to use a newer version of TypeScript. There are two main changes requiring code changes in this library:

 - When creating a Promise, `resolve` can only be called with no parameters if the Promise explicitly has the `void` type parameter.
 - When catching a thrown error, the error now has the `unknown` type, so we need to explicitly check that type.


murgatroid99:
I'm going to close this and try it again after merging #2260.

PR #2268, opened 2022-11-08
grpc-js-xds: interop: output CPU profile logs in old framework tests

murgatroid99:
I am doing this with the old framework because I am unsure how this would work in the new framework when the client is in a docker container. The config file grabs artifacts from `github/grpc/reports/**`, so hopefully this should work.

### DIFF
--- a/packages/grpc-js-xds/scripts/xds.sh
+++ b/packages/grpc-js-xds/scripts/xds.sh
@@ -59,7 +59,7 @@ GRPC_NODE_TRACE=xds_client,xds_resolver,cds_balancer,eds_balancer,priority,weigh
     --gcp_suffix=$(date '+%s') \
     --verbose \
     ${XDS_V3_OPT-} \
-    --client_cmd="$(which node) --enable-source-maps --prof --logfile=grpc/reports/prof.log grpc-node/packages/grpc-js-xds/build/interop/xds-interop-client \
+    --client_cmd="$(which node) --enable-source-maps --prof --logfile=${KOKORO_ARTIFACTS_DIR}/grpc/reports/prof.log grpc-node/packages/grpc-js-xds/build/interop/xds-interop-client \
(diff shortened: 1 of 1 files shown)


packages/grpc-js-xds/scripts/xds.sh
```diff
@@ -59,7 +59,7 @@ GRPC_NODE_TRACE=xds_client,xds_resolver,cds_balancer,eds_balancer,priority,weigh
     --gcp_suffix=$(date '+%s') \
     --verbose \
     ${XDS_V3_OPT-} \
-    --client_cmd="$(which node) --enable-source-maps grpc-node/packages/grpc-js-xds/build/interop/xds-interop-client \
+    --client_cmd="$(which node) --enable-source-maps --prof --logfile=grpc/reports/prof.log grpc-node/packages/grpc-js-xds/build/interop/xds-interop-client \
```
sergiitk, line 62:
```suggestion
    --client_cmd="$(which node) --enable-source-maps --prof --logfile=${KOKORO_ARTIFACTS_DIR}/grpc/reports/prof.log grpc-node/packages/grpc-js-xds/build/interop/xds-interop-client \
```

sergiitk (approved):
I'm not sure if relative path is going to work correctly.

Only files from the artifact folder are collected. I'm not sure the relative path from the `cwd` is pointing to the artifacts folder.
It might be better to use the absolute path to the artifacts folder.

packages/grpc-js-xds/scripts/xds.sh
sergiitk: Also, you might want to ensure the folder exists with `mkdir -p`.

murgatroid99: Yes, that looks like a good idea: #2269

PR #2269, opened 2022-11-08
grpc-js-xds: interop: mkdir artifact directory before running tests

murgatroid99:
As suggested


packages/grpc-js-xds/scripts/xds.sh
```diff
@@ -48,6 +48,8 @@ git clone -b master --single-branch --depth=1 https://github.com/grpc/grpc.git
 
 grpc/tools/run_tests/helper_scripts/prep_xds.sh
 
+mkdir -p "${KOKORO_ARTIFACTS_DIR}/grpc/reports/prof.log"
```
sergiitk, line 51:
```suggestion
mkdir -p "${KOKORO_ARTIFACTS_DIR}/grpc/reports"
```

sergiitk (changes requested):

Issue #2265, opened 2022-11-06
Feature Request: Force grpc client to resolve DNS


CristianBB:
### Is your feature request related to a problem?
When using round_robin load balancing on GRPC client on a kubernetes cluster with autoscaling rules, clients never pick new replicas if current channels are always available, so autoscaling has no real effectivity here.

### Describe the solution you'd like
There is two different possible alternatives:

- Based on GRPC specs, make params `grpc.max_connection_age_ms`  and `grpc.max_connection_age_grace_ms` available in `@grpc/grpc-js` server config. In that way clients will receive the GOAWAY signal and will be forced to resolve again DNS addresses.
- Not defined into GRPC specs but I think would be better and easier to implement an interval on clients to force them to resolve DNS. In that way there is no need to force channels to be disconnected (that could affect in multiple ways to processes) and we have the possibility to define the interval when DNS will be resolved again

murgatroid99:
You are correct that the recommended way to trigger resolution updates is to use the `grpc.max_connection_age_ms` setting, and also that that feature is not currently implemented in the `@grpc/grpc-js` library. I will treat this issue as a feature request to implement that feature.

CristianBB:
Thank you @murgatroid99 ! In that case I've updated the issue title and description to have it properly tracked.

I really think that second choice would be better and easier to implement, but as it's not defined on GRPC specs I'm not sure if would be possible

murgatroid99:
The `grpc.max_connection_age_ms` and `grpc.max_connection_age_grace_ms` options are now supported in the v1.8.0 release, so I am considering this issue resolved. If you still want a different behavior that is not currently specified, please file a separate feature request.

PR #2272, opened 2022-11-10
grpc-js: Implement server connection management

murgatroid99:
This is an implementation of [gRFC A9: Server-side Connection Management](https://github.com/grpc/proposal/blob/master/A9-server-side-conn-mgt.md), except for `MAX_CONNECTION_IDLE` because I don't have an established setup for tracking open streams. This change was prompted by a request in #2265.

### DIFF
--- a/packages/grpc-js/src/server.ts
+++ b/packages/grpc-js/src/server.ts
@@ -63,6 +63,10 @@ import { ChannelzCallTracker, ChannelzChildrenTracker, ChannelzTrace, registerCh
 import { CipherNameAndProtocol, TLSSocket } from 'tls';
 import { getErrorCode, getErrorMessage } from './error';
 
+const UNLIMITED_CONNECTION_AGE_MS = ~(1<<31);
+const KEEPALIVE_MAX_TIME_MS = ~(1<<31);
+const KEEPALIVE_TIMEOUT_MS = 20000;
+
 const {
   HTTP2_HEADER_PATH
 } = http2.constants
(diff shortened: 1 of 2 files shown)


packages/grpc-js/src/server.ts
```diff
@@ -970,12 +983,69 @@ export class Server {
         this.channelzTrace.addTrace('CT_INFO', 'Connection established by client ' + clientAddress);
         this.sessionChildrenTracker.refChild(channelzRef);
       }
+      let connectionAgeTimer: NodeJS.Timer | null = null;
+      let connectionAgeGraceTimer: NodeJS.Timer | null = null;
+      let sessionClosedByServer = false;
+      if (this.maxConnectionAgeMs !== UNLIMITED_CONNECTION_AGE_MS) {
+        // Apply a random jitter within a +/-10% range
+        const jitterMagnitude = this.maxConnectionAgeMs / 10;
```
gnossen, line 991: Is this formula part of a protocol? Or just a one-off heuristic?

gnossen (approved):

packages/grpc-js/src/server.ts
gnossen: Ah, nevermind. I see it inside of A9.

murgatroid99: It's in the spec I linked:

> A per-connection random jitter of +/-10% will be added to MAX_CONNECTION_AGE to spread out connection storms.

PR #2276, opened 2022-11-15
proto-loader-gen-types Avoid TS enums

install:
Avoid using Typescript `enums`.

Instead of a TS enum that are both a value and a type, define a js object that will only hold either the `Number` or `String` (Depending on the configuration) representation of an `enum`.

For backwards compatibility, have the permissive type accept both the `Number` and `String` representations, and the restrictive type be a union of the `String` or `Number` representations.

These changes have the added benefit of simplifying the generator code, because `Message` and `Enum` types are now treated the same.

### DIFF
--- a/packages/proto-loader/bin/proto-loader-gen-types.ts
+++ b/packages/proto-loader/bin/proto-loader-gen-types.ts
@@ -209,14 +209,15 @@ function getTypeNamePermissive(fieldType: string, resolvedType: Protobuf.Type | 
         throw new Error('Found field with no usable type');
       }
       const typeInterfaceName = getTypeInterfaceName(resolvedType);
-      if (resolvedType instanceof Protobuf.Type || resolvedType instanceof Protobuf.Enum) {
+      if (resolvedType instanceof Protobuf.Type) {
         if (repeated || map) {
           return inputName(typeInterfaceName);
         } else {
           return `${inputName(typeInterfaceName)} | null`;
         }
       } else {
-        return '';
+        // Enum
+        return inputName(typeInterfaceName);
       }
   }
 }
@@ -311,7 +312,7 @@ function getTypeNameRestricted(fieldType: string, resolvedType: Protobuf.Type | 
         throw new Error('Found field with no usable type');
       }
       const typeInterfaceName = getTypeInterfaceName(resolvedType);
-      if (resolvedType instanceof Protobuf.Type || resolvedType instanceof Protobuf.Enum) {
+      if (resolvedType instanceof Protobuf.Type) {
         /* null is only used to represent absent message values if the defaults
          * option is set, and only for non-repeated, non-map fields. */
         if (options.defaults && !repeated && !map) {
@@ -320,7 +321,8 @@ function getTypeNameRestricted(fieldType: string, resolvedType: Protobuf.Type | 
           return `${outputName(typeInterfaceName)}`;
         }
       } else {
-        return '';
+        // Enum
+        return outputName(typeInterfaceName);
       }
   }
 }
--- a/packages/proto-loader/bin/proto-loader-gen-types.ts
+++ b/packages/proto-loader/bin/proto-loader-gen-types.ts
@@ -135,20 +135,16 @@ function getImportLine(dependency: Protobuf.Type | Protobuf.Enum | Protobuf.Serv
   /* If the dependency is defined within a message, it will be generated in that
    * message's file and exported using its typeInterfaceName. */
   if (dependency.parent instanceof Protobuf.Type) {
-    if (dependency instanceof Protobuf.Type) {
+    if (dependency instanceof Protobuf.Type || dependency instanceof Protobuf.Enum) {
       importedTypes = `${inputName(typeInterfaceName)}, ${outputName(typeInterfaceName)}`;
-    } else if (dependency instanceof Protobuf.Enum) {
-      importedTypes = `${typeInterfaceName}`;
     } else if (dependency instanceof Protobuf.Service) {
       importedTypes = `${typeInterfaceName}Client, ${typeInterfaceName}Definition`;
     } else {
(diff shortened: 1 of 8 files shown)


packages/proto-loader/golden-generated/google/showcase/v1beta1/Severity.ts
```diff
@@ -3,9 +3,27 @@
 /**
  * A severity enum used to test enum capabilities in GAPIC surfaces
  */
-export enum Severity {
```
install, line 6: Example of how the generated code for an `Enum` changes

packages/proto-loader/bin/proto-loader-gen-types.ts
```diff
@@ -455,21 +447,46 @@ function generateMessageInterfaces(formatter: TextFormatter, messageType: Protob
...
 function generateEnumInterface(formatter: TextFormatter, enumType: Protobuf.Enum, options: GeneratorOptions, nameOverride?: string) {
+  const {inputName, outputName} = useNameFmter(options);
+  const name = nameOverride ?? enumType.name;
   formatter.writeLine(`// Original file: ${(enumType.filename ?? 'null')?.replace(/\\/g, '/')}`);
   formatter.writeLine('');
   if (options.includeComments) {
     formatComment(formatter, enumType.comment);
   }
-  formatter.writeLine(`export enum ${nameOverride ?? enumType.name} {`);
+  formatter.writeLine(`export const ${name} = {`);
```
murgatroid99, line 457: This can easily cause conflicts. There is no guarantee that `name` is different from `inputName` or `outputName`, and in fact by default `inputName` is equal to `name`.

Also, how do you expect this object to be used by the user of the generated code?

packages/proto-loader/bin/proto-loader-gen-types.ts
```diff
@@ -213,14 +209,14 @@ function getTypeNamePermissive(fieldType: string, resolvedType: Protobuf.Type |
         throw new Error('Found field with no usable type');
       }
       const typeInterfaceName = getTypeInterfaceName(resolvedType);
-      if (resolvedType instanceof Protobuf.Type) {
+      if (resolvedType instanceof Protobuf.Type || resolvedType instanceof Protobuf.Enum) {
         if (repeated || map) {
           return inputName(typeInterfaceName);
         } else {
           return `${inputName(typeInterfaceName)} | null`;
```
murgatroid99, line 216: Does Protobuf.js accept `null` as an input for an enum field?

packages/proto-loader/bin/proto-loader-gen-types.ts
```diff
@@ -315,7 +311,7 @@ function getTypeNameRestricted(fieldType: string, resolvedType: Protobuf.Type |
         throw new Error('Found field with no usable type');
       }
       const typeInterfaceName = getTypeInterfaceName(resolvedType);
-      if (resolvedType instanceof Protobuf.Type) {
+      if (resolvedType instanceof Protobuf.Type || resolvedType instanceof Protobuf.Enum) {
         /* null is only used to represent absent message values if the defaults
          * option is set, and only for non-repeated, non-map fields. */
         if (options.defaults && !repeated && !map) {
```
murgatroid99, line 317: Does Protobuf.js ever output `null` for an enum field?

packages/proto-loader/bin/proto-loader-gen-types.ts
install: Good catch. fixed to not allow `null` and match the previous types.

packages/proto-loader/bin/proto-loader-gen-types.ts
install: @murgatroid99 Because it's no longer using a TS enum, which is both a value and a type, we're allowed to have a type override for that value.

Here's an example output with the default input/outputNames and is valid TS:

```ts
// Original file: deps/gapic-showcase/schema/google/showcase/v1beta1/echo.proto

/**
 * A severity enum used to test enum capabilities in GAPIC surfaces
 */
export const Severity = {
  UNNECESSARY: 'UNNECESSARY',
  NECESSARY: 'NECESSARY',
  URGENT: 'URGENT',
  CRITICAL: 'CRITICAL',
} as const;

/**
 * A severity enum used to test enum capabilities in GAPIC surfaces
 */
export type Severity =
  | 'UNNECESSARY'
  | 0
  | 'NECESSARY'
  | 1
  | 'URGENT'
  | 2
  | 'CRITICAL'
  | 3

/**
 * A severity enum used to test enum capabilities in GAPIC surfaces
 */
export type Severity__Output = typeof Severity[keyof typeof Severity]
```

And a consuming message also is still valid:

```ts
// Original file: deps/gapic-showcase/schema/google/showcase/v1beta1/echo.proto

import type { Severity as _google_showcase_v1beta1_Severity, Severity__Output as _google_showcase_v1beta1_Severity__Output } from '../../../google/showcase/v1beta1/Severity';

/**
 * The response message for the Echo methods.
 */
export interface EchoResponse {
  /**
   * The content specified in the request.
   */
  'content'?: (string);
  /**
   * The severity specified in the request.
   */
  'severity'?: (_google_showcase_v1beta1_Severity);
}

/**
 * The response message for the Echo methods.
 */
export interface EchoResponse__Output {
  /**
   * The content specified in the request.
   */
  'content': (string);
  /**
   * The severity specified in the request.
   */
  'severity': (_google_showcase_v1beta1_Severity__Output);
}

```

The `const` is still there mostly for backwards compatibility. Right now the Enum name isn't modified at all by the `input` or `output` name formatting. So current users that import the old TS enum will now get the JS object with the values actually matching the configured `enum` type.

murgatroid99 (approved):
Thank you for your contribution. This will most likely go out in a release after Thanksgiving.

PR #2278, opened 2022-11-16
grpc-js: Implement retries

murgatroid99:
This implements the actual functionality of [gRFC A6: Client Retries](https://github.com/grpc/proposal/blob/master/A6-client-retries.md). The main changes are:

 - Add the `RetryingCall` class, which implements the retry functionality.
 - Modify the `SubchannelCall` class to report what RST_STREAM code ended a stream, if any, and the `LoadBalancingCall` class to use that information to change what call progress it reports with the status.
 - Add the channel arguments for configuring retries.
 - Modify the `ResolvingLoadBalancer` to pass the service config to the channel along with the config selector when name resolution completes.
 - Modify the server call class to enable it to send trailers-only responses, which are necessary to trigger configured retries on the client.

The tests mainly just test the basic functionality: that the client can make requests in various configurations, and that when retries or hedging are configured, the client makes the appropriate number of attempts per call.

### DIFF
--- a/packages/grpc-js/src/service-config.ts
+++ b/packages/grpc-js/src/service-config.ts
@@ -133,14 +133,14 @@ function validateRetryPolicy(obj: any): RetryPolicy {
   for (const value of obj.retryableStatusCodes) {
     if (typeof value === 'number') {
       if (!Object.values(Status).includes(value)) {
-        throw new Error('Invlid method config retry policy: retryableStatusCodes value not in status code range');
+        throw new Error('Invalid method config retry policy: retryableStatusCodes value not in status code range');
       }
     } else if (typeof value === 'string') {
       if (!Object.values(Status).includes(value.toUpperCase())) {
-        throw new Error('Invlid method config retry policy: retryableStatusCodes value not a status code name');
+        throw new Error('Invalid method config retry policy: retryableStatusCodes value not a status code name');
       }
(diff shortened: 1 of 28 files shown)


packages/grpc-js/src/service-config.ts
```diff
@@ -111,6 +111,81 @@ function validateName(obj: any): MethodConfigName {
...
+    throw new Error('Invalid method config hedging policy: maxAttempts must be an integer at least 2');
+  }
+  if (('hedgingDelay' in obj) && (typeof obj.hedgingDelay !== 'string' || !DURATION_REGEX.test(obj.hedgingDelay))) {
+    throw new Error('Invalid method config hedging policy: hedgingDelay must be a string consisting of a positive integer followed by s');
+  }
+  if (('nonFatalStatusCodes' in obj) && Array.isArray(obj.nonFatalStatusCodes)) {
+    for (const value of obj.nonFatalStatusCodes) {
+      if (typeof value === 'number') {
+        if (!Object.values(Status).includes(value)) {
+          throw new Error('Invlid method config hedging policy: nonFatalStatusCodes value not in status code range');
```
gnossen, line 166: Nit: `s/Invlid/Invalid/g`

It appears in three places in this file.

gnossen (approved):

PR #2280, opened 2022-11-29
grpc-js-xds: Implement retry support

murgatroid99:
This is an implementation of [gRFC A44: gRPC xDS Retry Support](https://github.com/grpc/proposal/blob/master/A44-xds-retry.md).


gnossen (approved):
Any accompanying tests? Or is the idea to just use the xDS interop tests for validation?

murgatroid99:
The plan is to use the xDS interop tests for validation. That is what I have done for the other xDS features. I don't currently have a good setup for testing those features locally.

PR #2295, opened 2022-12-08
grpc-js: Add support for grpc.service_config_disable_resolution

### LINKED ISSUE #1518: Add support for "grpc.service_config_disable_resolution" option in grpc-js
Any news on when grpc-js will support the option to disable service config resolution? We are observing dns lookups take much longer than expected (up to 500ms) when the grpc library makes a simultaneous TXT DNS query (the TXT queries always fail). Based on issues I found in repos of other grpc implementations the general advice is to turn off service config resolution in order to make the library stop making TXT queries. We are observing the same thing happening both in case of node-firestore and node-pubsub. 

node-firestore version: 3.5.5
node-pubsub version: 2.0.0

We also tried using the c++ based node grpc library, however, firestore does not seem to be compatible with it.

Thanks in advance!

kleinsch:
Adds support for `grpc.service_config_disable_resolution` to fix #1518 .

### DIFF
--- a/packages/grpc-js/README.md
+++ b/packages/grpc-js/README.md
@@ -62,6 +62,7 @@ Many channel arguments supported in `grpc` are not supported in `@grpc/grpc-js`.
   - `grpc.enable_retries`
   - `grpc.per_rpc_retry_buffer_size`
   - `grpc.retry_buffer_size`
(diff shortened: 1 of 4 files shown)


kleinsch:
Opened a second PR to try to get EasyCLA to wake up....

PR #2315, opened 2023-01-08
grpc-js: delays maybeOutputStatus in end event handler a bit

chessl:
Fix some edge cases that the process.nextTick in `push` function hasn't been processed yet while the end stream event kicks in, which updates `statusOutput` causing the last chunk of data desserted.


murgatroid99:
The change I made in the PR I linked has been published in version 1.8.2. That change supersedes this one so I'm closing this PR.

PR #2320, opened 2023-01-11
grpc-js: Add tracing for resolver results

murgatroid99:
This adds tracing for all success and failure results from resolvers. The main goal is to investigate #2318 by logging the service config that is created from a TXT DNS record, if there is one.


murgatroid99:
I don't think this is needed anymore.

PR #2323, opened 2023-01-13
xds interop: Fix buildscripts not continuing on a failed test suite

sergiitk:
Apparently there's a difference between bash 3 and bash 4. OSX comes with bash 3 out-of-box, so for whoever wrote this logic it "worked on my machine".

### DIFF
--- a/packages/grpc-js-xds/scripts/xds_k8s_lb.sh
+++ b/packages/grpc-js-xds/scripts/xds_k8s_lb.sh
@@ -167,7 +167,7 @@ main() {
   local failed_tests=0
   test_suites=("baseline_test" "api_listener_test" "change_backend_service_test" "failover_test" "remove_neg_test" "round_robin_test" "outlier_detection_test")
   for test in "${test_suites[@]}"; do
-    run_test $test || (( failed_tests++ ))
+    run_test $test || (( failed_tests++ )) && true
   done
   echo "Failed test suites: ${failed_tests}"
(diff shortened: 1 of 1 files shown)


sergiitk:
Same as https://github.com/grpc/grpc/pull/32093.
Proof of concept fix: https://source.cloud.google.com/results/invocations/f076d25c-105b-48ee-a722-b17126b83604

murgatroid99 (commented):
I think I understand why the current code might not be working: the `((` construct returns a 0 exit code if the value is non-zero. Since the value starts at 0 and we do a post-increment, it will always fail the first time. If that is the problem, changing it to a pre-increment should fix it.

I don't understand how adding `&& true` could ever possibly do anything. It's like writing `&& true` in other languages.

sergiitk:
`&& true` ensures the exit code of this op is always true. It's like `|| true` used in bash to suppress a command that failed.https://stackoverflow.com/questions/11231937/bash-ignoring-error-for-a-particular-command

In this case, we're doing the same, just after the increment op. 

But good idea with the pre-increment. I'll give it a try.

sergiitk:
Yep, you're right. Pre-increment works. However, the `&& true` change's already merged to the other repos. @murgatroid99  Would you be ok accepting this as is? Or you feel strongly about it?

sergiitk (commented):
Address the feedback: use pre-increment instead of `&& true`

murgatroid99 (approved):

Issue #1405, opened 2020-05-04
missing pre-compiled arm binary


shotor:
### Problem description

Can't install grpc-tools on arm architecture (raspberry pi) because of missing precompiled binaries.

```
node-pre-gyp ERR! stack Error: 404 status code downloading tarball https://node-precompiled-binaries.grpc.io/grpc-tools/v1.8.1/linux-arm64.tar.gz
```

Is it possible to build these at all?

### Reproduction steps

npm init
npm install grpc-tools

### Environment
- Ubuntu 20.04 arm64
 - 10-12-14
 - n
 - grpc-tools, possibly more grpc packages - haven't checked yet

### Additional context

```
error /home/repo/grpc/node_modules/grpc-tools: Command failed.
Exit code: 1
Command: node-pre-gyp install
Arguments:
Directory: /home/repo/grpc/node_modules/grpc-tools
Output:
node-pre-gyp info it worked if it ends with ok
node-pre-gyp info using node-pre-gyp@0.12.0
node-pre-gyp info using node@10.20.1 | linux | arm64
node-pre-gyp WARN Using request for node-pre-gyp https download
node-pre-gyp info check checked for "/home/repo/grpc/node_modules/grpc-tools/bin/grpc_tools.node" (not found)
node-pre-gyp http GET https://node-precompiled-binaries.grpc.io/grpc-tools/v1.8.1/linux-arm64.tar.gz
node-pre-gyp http 404 https://node-precompiled-binaries.grpc.io/grpc-tools/v1.8.1/linux-arm64.tar.gz
node-pre-gyp ERR! install error
node-pre-gyp ERR! stack Error: 404 status code downloading tarball https://node-precompiled-binaries.grpc.io/grpc-tools/v1.8.1/linux-arm64.tar.gz
node-pre-gyp ERR! stack     at Request.<anonymous> (/home/repo/grpc/node_modules/node-pre-gyp/lib/install.js:149:27)
node-pre-gyp ERR! stack     at Request.emit (events.js:203:15)
node-pre-gyp ERR! stack     at Request.onRequestResponse (/home/repo/grpc/node_modules/request/request.js:1059:10)
node-pre-gyp ERR! stack     at ClientRequest.emit (events.js:198:13)
node-pre-gyp ERR! stack     at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:565:21)
[...]

PR #2333, opened 2023-01-22
Add support for arm64 linux.

RedstoneWizard08:
Closes issue #1405 

Info:
- Merges toolchain files into `linux.toolchain.cmake`
- Updates build image to use ubuntu and fixes issues with x86-only packages
- Updates build script
- Adds general linux/arm64 support to `grpc-node`


murgatroid99:
I'm concerned about updating the base image to `ubuntu:22.04`. Historically, we have purposely run builds in an older environment to get compatibility with a wider range of glibc versions.

murgatroid99:
OK, here are the options as I see them:

 - You get this to work with Debian Stretch
 - We use this change as is, and if there are issues with glibc compatibility I will revert this change.

RedstoneWizard08:
I'll work on using debian stretch, or some other distro with the equivalent glibc version.

RedstoneWizard08:
The main problem is finding the gcc binaries for debian stretch. I might end up pulling from the ubuntu repo and manually installing them.

RedstoneWizard08:
@murgatroid99 I have chosen a better third option - using static linking. This allows it to run on hopefully any glibc version (as my research says) and it compiles well with cross. The cross-compilers are unavailable for debian stretch, and they work really well on Ubuntu Jammy. Older Ubuntu versions with older or the same glibc versions (I.E. Artful) don't have cross compilers.

RedstoneWizard08:
@murgatroid99 I just fixed that, hopefully.

murgatroid99 (approved):

PR #2336, opened 2023-01-25
grpc-js: Hold a reference to transport in SubchannelCall

murgatroid99:
https://github.com/grpc/grpc-node/issues/2318#issuecomment-1403124466 reported a regression in that previously fixed issue. I suspect that the problem is that once a subchannel stops using a transport, nothing has a reference to that transport, which makes it eligible to be garbage collected. To avoid this, the subchannel call now holds a reference to the transport, so it stays alive as long as at least one active call is using it.

I also removed the `disconnectListener` because it became unused in the previous refactor.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/grpc-js",
-  "version": "1.8.5",
+  "version": "1.8.6",
   "description": "gRPC Library for Node - pure JS implementation",
   "homepage": "https://grpc.io/",
   "repository": "https://github.com/grpc/grpc-node/tree/master/packages/grpc-js",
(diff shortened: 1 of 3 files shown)


gnossen (approved):
Is there a way to reproduce this issue in a unit test?

murgatroid99:
The problem with bugs like this is that the test framework code seems to change the behavior. I do have a set of local interop tests that run the client in a separate process in order to catch some bugs like this, but they don't catch this one and I don't know why. In this specific case, if I'm right and it's tied to garbage collection, that makes it even more unpredictable.

yamen:
Any open "ref" in my scripts would cause the bug not to appear (presumably because GC isn't triggered at the same time). But it appears 100% consistently in scripts that are very simple and guarantee no other open refs.

PR #2344, opened 2023-02-03
feat: use Record utility type for generated types

Rvtea:
`Record` is being introduced into TS since v2.1 (per https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type ) and it just released v5.0 recently, I think it is time for us to embrace the utility type to avoid directly object type writing.


Rvtea:
> What specific benefits do you think this change will provide? From what I can find, this will be identical from a type checking perspective.

To me, this will be more clear when reading the generated files based on the Utility type. Such as when you encounter a wrapped object, I think `Record<K, Record<U, T>>` is more clear than using index signature to display it.

murgatroid99:
> Such as when you encounter a wrapped object, I think `Record<K, Record<U, T>>` is more clear than using index signature to display it.

This never happens. As your change shows, the only input that results in an indexed object type is a map field. A map field cannot have a map type as its value type, so so the code generator will never generate nested indexed object types. If the field's value type is a message type, the resulting indexed object types will have that message type's interface name as the value type.

If there is not any functional difference, then this change just adds a layer of indirection, and I do not see that as a positive change.

Rvtea:
I got your point, thx for the explanation. I will close this out.

PR #2357, opened 2023-02-10
Write @deprecated jsdoc annotation if includeComments is enabled

### LINKED ISSUE #2348: Add JSDoc comment to mark fields as deprecated in output interfaces
It would be useful to get JSDoc comments when marking a field as deprecated like this:
```proto
uint32 count = 1 [deprecated = true];
```

The output interface would get the `@deprecated` comment added to the field.
```ts
/** @deprecated */
count: number;
```

It could be an option when generating if these comments should be included or not.

danielronnkvist:
By adding the JSDoc `@deprecated` annotation to the output comments to interface fields IDE's can pick it up and deliver a better developer experience. This annotation is only added when the `includeComments` option is enabled. Since protobuf.js already parses if a field is deprecated or not the options object with that information is passed to the `formatComment` function.

Update golden generated with @deprecated annotation.

Resolves #2348

### DIFF
--- a/packages/proto-loader/bin/proto-loader-gen-types.ts
+++ b/packages/proto-loader/bin/proto-loader-gen-types.ts
@@ -189,7 +189,7 @@ const typeBrandHint = `This field is a type brand and is not populated at runtim
 https://github.com/grpc/grpc-node/pull/2281`;
 
 function formatTypeBrand(formatter: TextFormatter, messageType: Protobuf.Type) {
-  formatComment(formatter, typeBrandHint, messageType.options);
+  formatComment(formatter, typeBrandHint);
   formatter.writeLine(`__type: '${messageType.fullName}'`);
 }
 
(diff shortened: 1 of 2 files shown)


packages/proto-loader/bin/proto-loader-gen-types.ts
```diff
@@ -184,7 +189,7 @@ const typeBrandHint = `This field is a type brand and is not populated at runtim
 https://github.com/grpc/grpc-node/pull/2281`;
 
 function formatTypeBrand(formatter: TextFormatter, messageType: Protobuf.Type) {
-  formatComment(formatter, typeBrandHint);
+  formatComment(formatter, typeBrandHint, messageType.options);
```
murgatroid99, line 192: I don't think this one is right. This comment isn't on the message interface itself, but rather on a synthetic field in the interface. So it doesn't make sense for this field to inherit the message type's deprecated status.

danielronnkvist: @murgatroid99 I removed the options argument for this format call.

murgatroid99 (approved):

murgatroid99:
This is out in version 0.7.5.

PR #2370, opened 2023-02-22
proto-loader-gen-types Narrow field Long check

install:
- Explicitly list the primitive field types that use Long, instead of searching for `64` in the type name.

Currently the check for when to import `Long` just looks to see if any field type has `64` in the name.
This potentially leads to adding unused imports that causes a typescript error, when using something like `google.protobuf.Int64Value`

```
src/__generated__/Example.ts:6:1 - error TS6133: 'Long' is declared but its value is never read.

6 import type { Long } from '@grpc/proto-loader';
```


murgatroid99 (approved):

install:
@murgatroid99 any chance you can help cut a release for this?

murgatroid99:
Sorry for the delay. This is out now in version 0.7.6.

PR #2373, opened 2023-02-24
Fix loading protofiles on Windows when using includeDirs

icodragon:
In the process of loading protofiles, I noticed that errors appear
Error: duplicate name 'module' in Namespace ...

The protofiles were fixed and I specified the path to them through includeDirs.
On UNIX systems, file uploads worked correctly. There is a problem on Windows

After analyzing, I realized that the error is in path.join. On Windows the path is returned with \ , on UNIX systems it is /

### DIFF
--- a/packages/proto-loader/src/util.ts
+++ b/packages/proto-loader/src/util.ts
@@ -27,7 +27,7 @@ function addIncludePathResolver(root: Protobuf.Root, includePaths: string[]) {
       return target;
     }
     for (const directory of includePaths) {
-      const fullPath: string = path.join(directory, target);
+      const fullPath: string = path.join(directory, target).replace(/\\/g, '/');
       try {
         fs.accessSync(fullPath, fs.constants.R_OK);
         return fullPath;


murgatroid99:
`path.join` uses `\` as the path separator on Windows because that is the correct path separator on Windows. I don't believe this is a correct change. Please file an issue with details of the problem you are encountering.

icodragon:
It will probably be easier for me to show you with an example with code:
https://github.com/icodragon/cosmos-proto
Try to run demo.js on Windows platform and you will get an error. Works well for me on MacOS.

Perhaps my fix is ​​not quite correct, I think the reason is not in the proto-loader itself, but in protobufjs, how it saves the loaded proto-files. But still, I provided a fix, maybe my fix will suit someone.

murgatroid99:
The issue is that we have tests that use this basic functionality, and they pass on Windows. So we know the basic functionality works, and I am not going to change it without a clear understanding of what the problem is, and why the change fixes that problem, and most importantly that the change doesn't break anything else. There are two things that are unusual about your usage: you are passing a large number of files as the `loadSync` argument, and you are generating that list using `glob`. Both of those make it hard to tell what about the input is causing the problem. Can you reproduce this problem with a single or small number of input files, and with hardcoded filenames instead of `glob` output? Whether or not you can, I again request that you file an issue with details including information about the input to `loadSync`, the expected output, and the actual output.

murgatroid99:
I still do not believe this change is correct. Please file an issue with the details of the problem and a clear, simple reproduction.

Issue #2375, opened 2023-02-26
Stream closed before all message are received


phoenix741:
### Problem description
The client send multiple message through a stream, then close the stream and wait a response. The server receive all the element, but the end of stream can be receive before all the message and then last message are ignored.

### Reproduction steps
I have a client that write to the stream :

```ts
      for await (const y of x) {
        console.log('write', y.toString('utf-8'));
        writer.write(y);
      }
      console.log('end');
      writer.end();
```

The server use requestStream: ServerReadableStream<RefreshCacheRequest, RefreshCacheReply> :

```ts
    for await (const x of request) {
      console.log('read', x.toString('utf8'));
    }
```

I have some missing event in read. If i had a setTimeout 5s arroud writer.end() i have all the message.

In the log i have :

```
[w:client] D 2023-02-25T22:18:39.562Z | server_call | Received message of length 361
[w:client] D 2023-02-25T22:18:39.562Z | server_call | Received message of length 859
[w:client] D 2023-02-25T22:18:39.562Z | server_call | Received message of length 1028
[w:client] D 2023-02-25T22:18:39.562Z | server_call | Received message of length 1267
[w:client] D 2023-02-25T22:18:39.562Z | server_call | Received message of length 1311
[w:client] D 2023-02-25T22:18:39.562Z | server_call | Received message of length 1437
[w:client] D 2023-02-25T22:18:39.563Z | server_call | Received message of length 1431
[w:client] D 2023-02-25T22:18:39.563Z | server_call | Received end of stream
[w:client] D 2023-02-25T22:18:39.563Z | server_call | Request to method /woodstock.WoodstockClientService/RefreshCache ended with status code: OK details: OK
[w:client] D 2023-02-25T22:18:39.563Z | server_call | Received message of length 3390
[w:client] D 2023-02-25T22:18:39.563Z | server_call | Received message of length 3937
[w:client] D 2023-02-25T22:18:39.565Z | server_call | Request to method /woodstock.WoodstockClientService/RefreshCache stream closed with rstCode 0
```

And the last two message are lost. The Request to method send the response because the stream is ended.

If i read with a setTimeout, all message are processed and the event are in the good order:

```ts
    for await (const x of request) {
      console.log('read', x.toString('utf8'));
      await setTimeout(0);
    }
```

Log:
```
[w:client] D 2023-02-26T11:16:58.697Z | server_call | Received message of length 859
[...]

PR #2376, opened 2023-02-26
grpc-js: add await/async on method that return promise

phoenix741:
add await/async on method that return promise to ensure that the order of message (and of the end of stream) are preserved

The goal is to fix the ticket #2375 where the end of stream can arrive before last messages and the message can be lost.

### DIFF
--- a/packages/grpc-js/src/server-call.ts
+++ b/packages/grpc-js/src/server-call.ts
@@ -808,10 +808,10 @@ export class Http2ServerCallStream<
 
     let pushedEnd = false;
 
-    const maybePushEnd = () => {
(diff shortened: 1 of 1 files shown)


phoenix741:
Hello,

Can I afford to ask what is the rest of the process on the PullRequest? (Are we still waiting for something from me to continue)

murgatroid99:
I'm sorry about the delay. I'm not waiting on anything from you, I've just been focused on other things.

murgatroid99 (approved):

murgatroid99:
I published this in version 1.8.12

phoenix741:
Thanks

PR #2394, opened 2023-03-10
grpc-js: Unregister socket from channelz when closing transport

### LINKED ISSUE #2393: Creating multiple clients leaks memory
### Problem description

Creating new clients and calling `close()` in a loop results in a memory leak. Originally reported as https://github.com/googleapis/gapic-generator-typescript/issues/1350.

### Reproduction steps

`package.json`:
```
{
  "dependencies": {
    "@google-cloud/spanner": "^6.7.2",
    "@grpc/grpc-js": "^1.8.12",
    "@grpc/proto-loader": "^0.7.5",
    "google-auth-library": "^8.7.0"
  }
}
```

`index.js`:
```
const auth = require('google-auth-library');
const grpc = require('@grpc/grpc-js');
const loader = require('@grpc/proto-loader');
const spannerProtoJson = require('@google-cloud/spanner/build/protos/protos.json');

async function test(ctr, credentials, project) {
  return new Promise((resolve, reject) => {

    const serviceClient = new ctr('spanner.googleapis.com', credentials);
    serviceClient.listInstanceConfigs({
      parent: `projects/${project}`,
    }, (err, result) => {
      serviceClient.close();
      if (err) {
        reject(err);
      } else {
        resolve(result);
      }
    });
  });
}

async function main() {
  const options = {
      keepCase: false,
      longs: String,
      enums: String,
      defaults: true,
      oneofs: true,
  };

  const packageDefinition = loader.fromJSON(spannerProtoJson, options);

  const packageObject = grpc.loadPackageDefinition(packageDefinition);
  const authClient = new auth.GoogleAuth();
  const project = await authClient.getProjectId();
  const credentials = grpc.credentials.combineChannelCredentials(
      grpc.credentials.createSsl(),
      grpc.credentials.createFromGoogleCredential(authClient)
  );

  const ctr = packageObject.google.spanner.admin.instance.v1.InstanceAdmin;

  for (let iter = 0; iter < 20; ++iter) {

    await test(ctr, credentials, project);

    global.gc();
    const used = process.memoryUsage().heapUsed / 1024 / 1024;
    console.log(`After iteration ${iter}: Current mem usage ${Math.round(used * 100) / 100} MB`);
  }
}

main();
```

You are expected to be authenticated with `gcloud` and have Spanner API enabled. If not, I'm assuming the same will work for any other gRPC backend.

Note that `@google-cloud/spanner` is only imported for its proto JSON and is never used in the repro.

### Environment
 - OS name, version and architecture: [e.g. Linux Ubuntu 18.04 amd64] gLinux x86_64
 - Node version [e.g. 8.10.0] v16.15.0
 - Node installation method [e.g. nvm] `nvm`
 - If applicable, compiler version [e.g. clang 3.8.0-2ubuntu4] N/A
 - Package name and version [e.g. gRPC@1.12.0] `@grpc/grpc-js@1.8.12`

### Additional context

N/A

murgatroid99:
This fixes #2393. Manual testing shows that without this change, after a channel is closed, the channelz registry reference keeps the transport alive, and that in turn keeps the subchannel alive.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/grpc-js",
-  "version": "1.8.12",
+  "version": "1.8.13",
   "description": "gRPC Library for Node - pure JS implementation",
   "homepage": "https://grpc.io/",
   "repository": "https://github.com/grpc/grpc-node/tree/master/packages/grpc-js",
--- a/packages/grpc-js/src/transport.ts
+++ b/packages/grpc-js/src/transport.ts
@@ -20,7 +20,7 @@ import { checkServerIdentity, CipherNameAndProtocol, ConnectionOptions, PeerCert
 import { StatusObject } from './call-interface';
 import { ChannelCredentials } from './channel-credentials';
 import { ChannelOptions } from './channel-options';
-import { ChannelzCallTracker, registerChannelzSocket, SocketInfo, SocketRef, TlsInfo } from './channelz';
+import { ChannelzCallTracker, registerChannelzSocket, SocketInfo, SocketRef, TlsInfo, unregisterChannelzRef } from './channelz';
 import { LogVerbosity } from './constants';
 import { getProxiedConnection, ProxyConnectionResult } from './http_proxy';
 import * as logging from './logging';
@@ -471,6 +471,7 @@ class Http2Transport implements Transport {
 
   shutdown() {
     this.session.close();
+    unregisterChannelzRef(this.channelzRef);
   }
 }
 


Matt-Esch:
Does the grpc spec specify that the default configuration should be to enable this? I was fairly convinced something related to channelz was leaking when I reported the memory issues. I disabled it just to reduce the memory overhead of having it there anyway, I wonder if it would be best to make this an opt in feature? I also think the same about retries, unless I'm mistaken it looked as if I was opted into retries by upgrading the version. That change is a bit more problematic.

murgatroid99:
The spec doesn't say anything about enabling or disabling data collection, but the corresponding [channel option in the C core](https://grpc.github.io/grpc/cpp/group__grpc__arg__keys.html#ga169f04dc1fa795c27b0daeda33c16999) defaults to enabled. At this point, we can't change it because that would potentially break any code that is using it. And once the leak is fixed, I don't think it would be enough of an issue that we would want to do that anyway.

The retry functionality that is enabled by default is [transparent retries](https://github.com/grpc/proposal/blob/master/A6-client-retries.md#transparent-retries), which are specified that way.

gnossen (approved):

PR #2403, opened 2023-03-30
Replace invalid link in grpc-tools README

pietrodn:
The provided link leads to a 404 since Google pulled the documentation page for the JS library and left the poor developers without any documentation.

The original page can be accessed via [Wayback Machine](http://web.archive.org/web/20220422095626/https://developers.google.com/protocol-buffers/docs/reference/javascript-generated).

The replacement link provides the same original content.


pietrodn:
I signed the CLA.

PR #2419, opened 2023-04-09
Skip the remaining callbacks on connectivity state change

Hexta:
Skip the remaining callbacks in `subchannel.transitionToState()` if connectivity state has been changed before callbacks called.

Connectivity state could be changed by calling `client.close()` inside a callback.
As a result - infinite busy loop in `LoadBalancingCall.doPick()`.

It affected [etcd3 client library](https://github.com/microsoft/etcd3/blob/00bfb8bf3a26e176c48b2f483d70df0605a76006/src/connection-pool.ts#L213]).

PR #2421, opened 2023-04-11
grpc-js: Fix connectivity state change event sequencing

murgatroid99:
The test in #2419 revealed a couple of bugs related to the order of connectivity state change events:

 - `PickFirstLoadBalancer#pickSubchannel` informs its parent of the state change before resolving its own internal state, and as a result arbitrary code can run at that point.
 - The calls to the list of listeners in `Subchannel#transitionToState` can eventually result in API calls that cause additional state changes, causing potential reentrancy in `transitionToState` which can result in out-of-order state change notifications.

The changes to fix this are:

 - Move the state reporting in `PickFirstLoadBalancer#pickSubchannel` to the end.
 - Refactor the `listener` list in `Subchannel` to a set to simplify interacting with it and to allow removing listeners during the iteration.
 - Ensure all calls to `transitionToState` are asynchronous with respect to surface API calls, to avoid reentrant `transitionToState` calls.

I adapted the test from #2419 to test this combination of bugs.

### DIFF
--- a/packages/grpc-js/src/subchannel.ts
+++ b/packages/grpc-js/src/subchannel.ts
@@ -60,7 +60,7 @@ export class Subchannel {
    * state changes. Will be modified by `addConnectivityStateListener` and
    * `removeConnectivityStateListener`
    */
-  private stateListeners: ConnectivityStateListener[] = [];
+  private stateListeners: Set<ConnectivityStateListener> = new Set();
 
   private backoffTimeout: BackoffTimeout;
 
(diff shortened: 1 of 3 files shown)


gnossen (approved):

packages/grpc-js/src/subchannel.ts
```diff
@@ -227,6 +227,8 @@ export class Subchannel {
     }
     const previousState = this.connectivityState;
     this.connectivityState = newState;
+    process.nextTick(() => {
```
bolt-juri-gavshin, line 230: What's the purpose of this call?
It looks like some debugging leftover. It doesn't make anything asynchronous, it just adds an empty function to be called by Event Loop after synchronous part of the code is executed.

murgatroid99: You're right. That was part of a change that I was trying out before I went in a different direction.

PR #2465, opened 2023-06-15
Convert to eslint/prettier

dancrumb:
This change is limited to `grpc-js` for now for a few reasons:

 - That's really the scope of my immediate interest :)
 - Updating _every_ package at once would make the PR much larger
 - This is the package that contains the most JS and TS

If you'd prefer to have every package updated, I'd suggest either:

 - I update this PR accordingly
 - I/we create an issue for each package and knock them out one at a time


murgatroid99:
FYI the Windows Test failure is not a concern here. It is a previously known failure unrelated to this change.

packages/grpc-js/test/common.ts
```diff
@@ -16,7 +16,7 @@
  */
 
 import * as loader from '@grpc/proto-loader';
-import * as assert from 'assert';
+import * as assert2 from './assert2';
```
dancrumb, line 19: @murgatroid99 this is one material change I made (although it's really just a simple refactor).

It just removes the exporting of a namespace.

packages/grpc-js/test/test-call-credentials.ts
```diff
@@ -86,21 +86,16 @@ describe('CallCredentials', () => {
...
       );
-      let metadata: Metadata;
-      try {
-        metadata = await callCredentials.generateMetadata({
-          service_url: 'foo',
-        });
-      } catch (err) {
-        throw err;
-      }
+      const metadata: Metadata = await callCredentials.generateMetadata({
```
dancrumb, line 89: @murgatroid99  this is another material change; I removed the redundant try/catch (since the catch block just rethrows the caught error).

This was causing a lint error and seems redundant anyway.

packages/grpc-js/test/test-call-credentials.ts
```diff
@@ -147,12 +138,10 @@ describe('CallCredentials', () => {
       await Promise.all(
         testCases.map(async testCase => {
           const { credentials, expected } = testCase;
-          let metadata: Metadata;
-          try {
```
dancrumb, line 151: @murgatroid99  this is another material change; I removed the redundant try/catch (since the catch block just rethrows the caught error).

This was causing a lint error and seems redundant anyway.

packages/grpc-js/apache-notice.md
```diff
@@ -0,0 +1,4 @@
+The following files contain configuration settings that were derived from [the this commit](https://github.com/google/gts/commit/3b9ab6dd59691f77f5c5c632a44c6762ba4ef7c6) in the Google GTS repository:
```
dancrumb, line 1: @murgatroid99 is this helpful, necessary, overkill? 🤷

murgatroid99: I don't know if it's necessary, but I have no objection to including it.

murgatroid99 (approved):
Thank you for doing this work.

PR #2501, opened 2023-07-12
Fix premature leaving of context due to improper Http2ServerCallStream handling

CedricKassen:
To provide context and metadata in our internal gRPC handling framework, we utilize nodejs's [asynchronous context tracking](https://nodejs.org/api/async_context.html), specifically metadata namespaces similar to `cls-hooked`, implemented with `AsyncLocalStorage`.

Due to the asynchronous nature of the `Http2ServerCallStream`, the synchronous handling of receiving unary messages introduced in #2249 causes unexpected behavior. Handling the stream outside of a `Promise` makes node leave the context prematurely before continuing the message handling, therefore rendering access to the information stored with `AsyncLocalStorage` impossible.

### DIFF
--- a/packages/grpc-js/src/server-call.ts
+++ b/packages/grpc-js/src/server-call.ts
@@ -615,19 +615,15 @@ export class Http2ServerCallStream<
         );
 
         if (Buffer.isBuffer(decompressedMessage)) {
-          call
-            .deserializeMessageWithInternalError(decompressedMessage)
-            .then(resolve)
-            .catch(reject);
+          resolve(
+            call.deserializeMessageWithInternalError(decompressedMessage)
+          );
           return;
         }
 
         decompressedMessage.then(
           decompressed =>
-            call
-              .deserializeMessageWithInternalError(decompressed)
-              .then(resolve)
-              .catch(reject),
+            resolve(call.deserializeMessageWithInternalError(decompressed)),
           (err: any) =>
             reject(
               err.code
--- a/packages/grpc-js/src/server-call.ts
+++ b/packages/grpc-js/src/server-call.ts
@@ -553,102 +553,100 @@ export class Http2ServerCallStream<
     return metadata;
   }
 
-  receiveUnaryMessage(
-    encoding: string,
-    next: (
-      err: Partial<ServerStatusResponse> | null,
-      request?: RequestType
-    ) => void
-  ): void {
-    const { stream } = this;
+  receiveUnaryMessage(encoding: string): Promise<RequestType | void> {
+    return new Promise((resolve, reject) => {
+      const { stream } = this;
 
-    let receivedLength = 0;
+      let receivedLength = 0;
 
-    // eslint-disable-next-line @typescript-eslint/no-this-alias
-    const call = this;
-    const body: Buffer[] = [];
-    const limit = this.maxReceiveMessageSize;
+      // eslint-disable-next-line @typescript-eslint/no-this-alias
+      const call = this;
+      const body: Buffer[] = [];
+      const limit = this.maxReceiveMessageSize;
 
-    stream.on('data', onData);
-    stream.on('end', onEnd);
-    stream.on('error', onEnd);
+      this.stream.on('data', onData);
+      this.stream.on('end', onEnd);
+      this.stream.on('error', onEnd);
 
-    function onData(chunk: Buffer) {
-      receivedLength += chunk.byteLength;
+      async function onData(chunk: Buffer) {
+        receivedLength += chunk.byteLength;
 
-      if (limit !== -1 && receivedLength > limit) {
+        if (limit !== -1 && receivedLength > limit) {
+          stream.removeListener('data', onData);
+          stream.removeListener('end', onEnd);
+          stream.removeListener('error', onEnd);
+
+          reject({
+            code: Status.RESOURCE_EXHAUSTED,
+            details: `Received message larger than max (${receivedLength} vs. ${limit})`,
+          });
+          return;
+        }
+
+        body.push(chunk);
+      }
+
+      async function onEnd(err?: Error) {
         stream.removeListener('data', onData);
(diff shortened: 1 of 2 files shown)


packages/grpc-js/src/server-call.ts
```diff
@@ -553,102 +553,100 @@ export class Http2ServerCallStream<
...
-    let receivedLength = 0;
+      this.stream.on('data', onData);
+      this.stream.on('end', onEnd);
+      this.stream.on('error', onEnd);
 
-    // eslint-disable-next-line @typescript-eslint/no-this-alias
-    const call = this;
-    const body: Buffer[] = [];
-    const limit = this.maxReceiveMessageSize;
+      async function onData(chunk: Buffer) {
```
murgatroid99, line 571: Making a function `async` has two main effects: it can `await` things and it returns a promise. This function does not `await` anything and it is used as an event listener so its return value is not used. So, I don't see any reason to make that change.

`onEnd` has the same issue.

packages/grpc-js/src/server-call.ts
```diff
@@ -553,102 +553,100 @@ export class Http2ServerCallStream<
...
   private safeDeserializeMessage(
     buffer: Buffer,
-    next: (
-      err: Partial<ServerStatusResponse> | null,
-      request?: RequestType
-    ) => void
+    resolve: (
+      value: void | RequestType | PromiseLike<void | RequestType>
+    ) => void,
+    reject: (reason: any) => void
```
murgatroid99, line 644: I don't like passing `resolve` and `reject` in here. It's not a normal async flow and it makes the code logic more complicated. It would be better for this function to return a promise itself. In fact, I think that if you make this function `async` and replace the entire function body with `return this.deserializeMessage(buffer)`, any error will bubble up as a promise rejection, which can be handled by the caller. Alternatively, you could keep the `try...catch`, but `throw` the object that is currently passed to `reject`.

packages/grpc-js/src/server.ts
```diff
@@ -96,6 +96,7 @@ function getUnimplementedStatusResponse(
   return {
     code: Status.UNIMPLEMENTED,
     details: `The server does not implement the method ${methodName}`,
+    metadata: new Metadata(),
```
murgatroid99, line 99: The return type of this function is specifically written to not need this field. I don't think it is necessary to add it.

murgatroid99 (changes requested):

packages/grpc-js/src/server-call.ts
CedricKassen: Sorry, this was a slight oversight. I halfway through opted to make this change as little as possible and overlooked this

packages/grpc-js/src/server-call.ts
CedricKassen: This was in an attempt to make this function behave is the same way as before. I can see how this definitely makes it less understandable. I opted to the second option, keeping the `try...catch`, because I want to keep the Error mapping, and otherwise I'd have to do it twice

packages/grpc-js/src/server-call.ts
```diff
@@ -553,105 +553,97 @@ export class Http2ServerCallStream<
...
 
-      const requestBytes = Buffer.concat(body, receivedLength);
-      const compressed = requestBytes.readUInt8(0) === 1;
-      const compressedMessageEncoding = compressed ? encoding : 'identity';
-      const decompressedMessage = call.getDecompressedMessage(
-        requestBytes,
-        compressedMessageEncoding
-      );
+        if (Buffer.isBuffer(decompressedMessage)) {
+          call.deserializeMessageWithInternalError(decompressedMessage);
```
murgatroid99, line 618: Now that this method returns the value, you have to do something with that value here and at the other call site. The way it currently works, I think you need another try/catch block here to pass the result into `resolve` or `reject` as appropriate. However, I think if you make `deserializeMessageWithInternalError` into an `async` function, it will automatically translate the `throw` there into a rejected promise. Then, since `resolve` here accepts a `PromiseLike<RequestType | void>`, you can pass the result of that call directly to `resolve` and it will handle both cases.

CedricKassen: Oh yeah you're right, I neglected the resolve there. I now turned it into an async function and specifically specified the resolve and catch at the call accordingly to make it very clear what is happening to the result. Otherwise it might become unclear to some, why `deserializeMessageWithInternalError` is async.

I thought if awaiting it might matter, because it would make the handling more complicated. But I don't think it makes a difference, because we're inside of another promise that only resolves after this call is finished anyway.

packages/grpc-js/src/server-call.ts
```diff
@@ -553,105 +553,103 @@ export class Http2ServerCallStream<
...
-      const compressedMessageEncoding = compressed ? encoding : 'identity';
-      const decompressedMessage = call.getDecompressedMessage(
-        requestBytes,
-        compressedMessageEncoding
-      );
+        if (Buffer.isBuffer(decompressedMessage)) {
+          call
+            .deserializeMessageWithInternalError(decompressedMessage)
+            .then(resolve)
+            .catch(reject);
```
murgatroid99, line 621: I'm pretty sure this can just be `resolve(call.deserializeMessageWithInternalError(decompressedMessage))`. The output of that function is now a promise-like object, which can be passed directly to `resolve`.

CedricKassen: I actually preferred the readability of explicitly passing resolve and reject, to actually show why we might want a Promise here, but I can also understand why you would want to keep this as simple as possible.
I'll push a change for that as well.

CedricKassen: I'm not really concerned, it's still very readable. I just personally prefer writing explicit code, but as I said I also understand why keeping it simplistic has its benefits, and I definitely want to adhere to the code style this repository aims to maintain.

murgatroid99 (approved):

PR #2504, opened 2023-07-13
proto-loader: Update long dependency to match protobufjs

haines:
protobufjs 7.x depends on long ^5.0.0. This PR updates the dependency in proto-loader to match, so that we avoid installing multiple versions of long.


murgatroid99:
The last attempt at this was in #2192. I'm trying to find out if it will be OK this time, but the person most likely to know will not be available for a week.

murgatroid99:
This change is now out in version 0.7.9.

PR #2538, opened 2023-08-08
grpc-js: Return LB policy configs from resolvers in JSON form

murgatroid99:
More generally, the principle here is that LB policy configs should be created and passed around as plain objects until they need to be passed to an LB policy, at which time they are parsed into a typed object.

In addition, an LB policy config for a non-leaf/petiole policy has a list of possible configs for each child policy, and the first usable config in the list is used. Prior to this change, that selection took place at child policy creation time. Now, it takes place at LB policy config parse time, recursively.

The purpose of this PR is to make it easier to implement [custom LB policies](https://github.com/grpc/proposal/blob/master/A52-xds-custom-lb-policies.md). Specifically, that spec requires generating arbitrary LB policy configs from `Struct` messages.

API changes:

 - The interface `ServiceConfig` and interfaces for its members `MethodConfig`, `LoadBalancingConfig`, and `RetryPolicy` have been moved out of the `experimental` namespace, into the stable public API.
 - The `LoadBalancingConfig` type now refers to simple raw objects instead of class instances.
 - `experimental.getFirstUsableConfig`, `experimental.validateLoadBalancingConfig`, and `experimental.OutlierDetectionLoadBalancingConfig` have been removed.
 - `experimental.selectLbConfigFromList`, `experimental.parseLoadBalancingConfig`, and `experimental.OutlierDetectionRawConfig` have been added.

### DIFF
--- a/packages/grpc-js-xds/src/duration.ts
+++ b/packages/grpc-js-xds/src/duration.ts
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2023 gRPC authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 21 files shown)


gnossen (approved):
TIL the word petiole.

Issue #2537, opened 2023-08-04
Cannot read properties of undefined (reading 'startConnecting')


akosyakov:
After upgrading 1.8.17 we see lot of

```
TypeError: Cannot read properties of undefined (reading 'startConnecting')
    at /app/node_modules/@grpc/grpc-js/src/load-balancer-pick-first.ts:308:43
    at processTicksAndRejections (node:internal/process/task_queues:78:11)
```

We tried to downgrade a bit to 1.8.8 but it still happens from time to time.

murgatroid99:
Can you check if the same thing happens with version 1.9.0?

svenefftinge:
I believe it can potentially still happen as [the code](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/load-balancer-pick-first.ts#L309) is executed on next tick and there's a on of other code pathes potentially being executed that would potentially change/clear the `children` property. 
So I'd say either capturing the `children` array in the lambda or checking for `undefined` would be important.

murgatroid99:
The fix for this has been published in version 1.9.1.

PR #2539, opened 2023-08-08
grpc-js: Add null check in pick_first array access

murgatroid99:
This addresses a possible cause of #2537. TypeScript doesn't require this check because the type system does not account for array out-of-bounds access.


gnossen (approved):

svenefftinge:
Thank you! That will fix the problem 🎉

PR #2559, opened 2023-08-29
Fix pick-first-load-balancer pick subchannel lost state listener

gusumuzhe:
If subchannel wrap isn't equals and wrap's real subchannel is same, will result subchannel‘s state listener loss. When grpc server down, pick first can't change connectivity state to TRANSIENT_FAILURE, and load-balancing-call doPick function will infinite call.


murgatroid99 (approved):
Thank you for the fix.

murgatroid99:
This has been published in version 1.9.2.

PR #2568, opened 2023-09-08
grpc-js-xds: Implement ring_hash LB policy

murgatroid99:
This implements [gRFC A42: xDS Ring Hash LB Policy](https://github.com/grpc/proposal/blob/master/A42-xds-ring-hash-lb-policy.md), with modifications described in gRFC A61 (grpc/proposal#356).

In addition to the changes specified in A42, I had to make the following changes to get ring_hash to work:

 1. The config selector now takes an additional `channelId` argument, to be hashed in the "CHANNEL_ID" hash policy.
 2. The `xxhash-wasm` dependency has been added.
 3. The TypeScript dependency has been updated to 4.x, to get access to the `Awaited` utility type used in the loader code for `xxhash-wasm`.
 4. `ResolvingLoadBalancer`, `XdsClusterManagerLoadBalancer`, and `PriorityLoadBalancer` now consistently delegate to the child when in a connectivity state other than `READY`.

Note: Hash generation only works on Node 15+, because that is when support for `bigint` representations of 64 bit integers from WASM modules was introduced. If the feature is enabled by the environment variable, the library will not work at all on Node 14 and below. For reference, Node 16 is currently the lowest supported LTS version.

Experimental API changes:

 - Added `EndpointMap`, an implementation of several `Map` APIs that considers `Endpoint` equality based on the unordered set of addresses. Not efficient.
 - The `ConfigSelector` function signature has an additional argument `channelId: number`, which is randomly generated for each channel, and has a consistent value for the lifetime of the channel. Implementations that don't use it can ignore it.
 - The `QueuePicker` class has an additional optional argument `childPicker: Picker`, which the `QueuePicker` instance will delegate to if provided.

### DIFF
--- a/packages/grpc-js-xds/gulpfile.ts
+++ b/packages/grpc-js-xds/gulpfile.ts
@@ -63,7 +63,9 @@ const compile = checkTask(() => execNpmCommand('compile'));
 const runTests = checkTask(() => {
   process.env.GRPC_EXPERIMENTAL_XDS_FEDERATION = 'true';
   process.env.GRPC_EXPERIMENTAL_XDS_CUSTOM_LB_CONFIG = 'true';
-  process.env.GRPC_XDS_EXPERIMENTAL_ENABLE_RING_HASH = 'true';
+  if (Number(process.versions.node.split('.')[0]) > 14) {
+    process.env.GRPC_XDS_EXPERIMENTAL_ENABLE_RING_HASH = 'true';
+  }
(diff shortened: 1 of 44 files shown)


murgatroid99:
The interop test passed in [this run](https://source.cloud.google.com/results/invocations/43f87149-edea-4f8b-a607-d093596da1b4/targets;collapsed=)

gnossen (approved):

PR #2586, opened 2023-09-26
grpc-js: Unref backoff timer in subchannel

murgatroid99:
This was making the tests taking a long time to exit after finishing. Our general principle is that grpc should keep the process open if and only if there is a pending request or other surface API event (e.g. channel state change). If a call is not assigned to a connection, the channel has a timer that keeps the process open, and if the call is assigned to a connection, the corresponding transport keeps the connection open. Either way, the subchannel is never responsible for that, so it should unref any timers or other handles that it owns.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/grpc-js",
-  "version": "1.9.3",
+  "version": "1.9.4",
   "description": "gRPC Library for Node - pure JS implementation",
   "homepage": "https://grpc.io/",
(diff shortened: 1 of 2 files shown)


gnossen (approved):

silentroach:
We found that in 1.9.4 with this unref if you write a script that makes a gRPC call and displays the output, the script might terminate (successfully) before the actual execution of the request, because it no longer waits for this timer.

murgatroid99:
@silentroach Please test this with the latest version (1.9.12) and then if the problem persists, file an issue.

silentroach:
Yep, the same problem in 1.9.12, will file an issue, thank you

PR #2599, opened 2023-10-17
grpc-js: Include library version and PID in all trace logs

### LINKED ISSUE #2594: Logs are all written to `console.error` and overriding the logger only partially works
### Problem description

In this library's `DEFAULT_LOGGER`, all logs are written to `console.error`, no matter whether it's an `error`, `info`, or `debug` log:

https://github.com/grpc/grpc-node/blob/83743646cf69baf9ae1294015de5ffed33339154/packages/grpc-js/src/logging.ts#L20-L30

I would expect the library to write to the appropriate channel instead (`error`, `info`, or `debug`).

### Reproduction steps

Just set `GRPC_TRACE` to `all` and `GRPC_VERBOSITY` to `DEBUG`. Tools like Datadog will indicate that **all** messages coming from `grpc-js` are errors, while they aren't.

### Environment
 - OS name, version and architecture: macOS 14.0 x86_64
 - Node version [e.g. 8.10.0] 18.16.0
 - Node installation method [e.g. nvm] nvm
 - If applicable, compiler version [e.g. clang 3.8.0-2ubuntu4] N/A
 - Package name and version [e.g. gRPC@1.12.0] grpc-js@1.9.5

### Additional context

I've overridden the logger as follows:

```TS
import {setLogger} from '@grpc/grpc-js'

setLogger({
    error: (message?: any, ...optionalParams: any[]) => {
        console.error('E ' + message, ...optionalParams);
    },
    info: (message?: any, ...optionalParams: any[]) => {
        console.info('I ' + message, ...optionalParams);
    },
    debug: (message?: any, ...optionalParams: any[]) => {
        console.debug('D ' + message, ...optionalParams);
    },
})
```

However, even when overriding the logger, the `Loading @grpc/grpc-js version XXX` message is still written to `console.error`:

```
2023-10-10T09:10:03.469Z	undefined	ERROR	D 2023-10-10T09:10:03.469Z | index | Loading @grpc/grpc-js version 1.9.5
```

https://github.com/grpc/grpc-node/blob/83743646cf69baf9ae1294015de5ffed33339154/packages/grpc-js/src/index.ts#L281-L285

This is because the TypeScript compilation process turns our code into `commonjs`, meaning that the log line above is triggered before we are even able to override the logger:

```JS
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });

// This call triggers the "Loading @grpc/grpc-js version XXX" message before the logger is overridden
const grpc_js_1 = require("@grpc/grpc-js");

(0, grpc_js_1.setLogger)({
    error: (message, ...optionalParams) => {
        console.error('E ' + message, ...optionalParams);
    },
    info: (message, ...optionalParams) => {
        console.info('I ' + message, ...optionalParams);
    },
    debug: (message, ...optionalParams) => {
        console.debug('D ' + message, ...optionalParams);
    },
});
```

Could you please consider:
- Setting the correct console log type for the corresponding log levels (`console.error`, `console.info`, `console.debug`)
- Updating the code in such way that the `Loading @grpc/grpc-js version XXX` message isn't triggered before we get a chance to override the logger

Thanks! Let me know if I can help with anything.

murgatroid99:
This fixes #2594 by removing the `index` tracer and moving the relevant information (the library version) into the common prefix of all trace log messages. That will also help confirm the version in the frequent case when people provide snippets of trace logs that don't include the first log, or don't have the `index` tracer enabled. I also included the PID because sometimes people provide logs with output from multiple processes interspersed, and this will help separate them.

### DIFF
--- a/doc/environment_variables.md
+++ b/doc/environment_variables.md
@@ -42,7 +42,6 @@ can be set.
   - `subchannel_internals` - Traces HTTP/2 session state. Includes per-call logs.
   - `channel_stacktrace` - Traces channel construction events with stack traces.
   - `keepalive` - Traces gRPC keepalive pings
-  - `index` - Traces module loading
   - `outlier_detection` - Traces outlier detection events
 
   The following tracers are added by the `@grpc/grpc-js-xds` library:
--- a/packages/grpc-js/src/index.ts
+++ b/packages/grpc-js/src/index.ts
@@ -273,14 +273,7 @@ import * as load_balancer_outlier_detection from './load-balancer-outlier-detect
 import * as channelz from './channelz';
 import { Deadline } from './deadline';
 
-const clientVersion = require('../../package.json').version;
-
 (() => {
-  logging.trace(
-    LogVerbosity.DEBUG,
-    'index',
-    'Loading @grpc/grpc-js version ' + clientVersion
-  );
   resolver_dns.setup();
   resolver_uds.setup();
   resolver_ip.setup();
--- a/packages/grpc-js/src/logging.ts
+++ b/packages/grpc-js/src/logging.ts
@@ -16,6 +16,9 @@
  */
 
 import { LogVerbosity } from './constants';
+import { pid } from 'process';
+
+const clientVersion = require('../../package.json').version;
 
 const DEFAULT_LOGGER: Partial<Console> = {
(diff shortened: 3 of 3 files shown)


gnossen (approved):
:+1: Really nice. We should consider this for Python

Issue #2448, opened 2023-05-19
@grpc/grpc-js has memory leak issue with latest version


alanzhao:
### Problem description
We use spanner node js for our api, and the latest grpc-js 1.8.14 has memory leak issues, same as 1.8.13.

### Reproduction steps
We use spanner node library for our api. We recently upgraded to node 18. so in package-lock json, grpc-js gets upgraded to 1.18.14.  Before upgrade, we were not having the memory leak issue.  After upgrade, we have a slow memory leak, manifested in 4-7 days.  And our pods restarted due to low memory.

Our stop gap solution is to downgrade @grpc/grpc-js to 1.6.2 and that seems to fix the issue.
Thanks for looking into this. Much appreciated.

### Environment
 - x86_64 Linux
 - 18.15
 - nvm
 - @grpc/grpc-js@1.8.14
 - @google-cloud/spanner@5.18.0

### Additional context
We have about 100 pods in prod, and all of them exhibit the symptom after the library gets updated.

murgatroid99:
This isn't really enough information to find the problem. Can you provide code that demonstrates this memory leak, preferably without needing to access an external service? Can you provide heap dumps from at least two different points in a process's lifetime that show the memory growth?

Please also check whether you are creating multiple separate client objects. If you are, either change your code to not do that or make sure you are closing them when you are done with them. See #2436, #2402, and #2003 for previous examples of this.

alanzhao:
```
async function getValue({ userId }) {
  const sql = `
    SELECT *
    FRROM ${table}
    WHERE userId = @userId
  `;

  const query = {
    sql,
    params: { userId },
    json: true,
  };

  const [[total]] = await clients.spanner.clientA.run(query);
  return total;
}

module.exports = getActivationTotalByUserId;
```

Here is a sample query that we use.  We do have a wrapper around spanner and have 3 clients: ` clients.spanner.clientA`, ` clients.spanner.clientB`, and ` clients.spanner.clientC`.


```
function allClose() {
  const closePromises = [];

  if (clients.spanner) {
    if (clients.spanner.clientA) {
      closePromises.push(clients.spanner.clientA.close());
    }

    if (clients.spanner.clientB) {
      closePromises.push(clients.spanner.clientB.close());
    }

    if (clients.spanner.clientC) {
      closePromises.push(clients.spanner.clientC.close());
    }
  }
```

which is used in the graphql server graceful shutdown here:

```
return startApp(app, ready, () => allClose());
```

[...]

PR #2606, opened 2023-10-27
grpc-js: Cancel and don't start idle timer on shutdown

murgatroid99:
This fixes the memory leak reported in https://github.com/grpc/grpc-node/issues/2448#issuecomment-1782690279, according to my manual testing.


Thibault2ss (approved):

Thibault2ss:
Thanks for the fix 👍 That was a fast response 😄

gnossen (approved):

Issue #2631, opened 2023-12-14
[@grpc/grpc-reflection] Issue with proto file dependencies only on second or next request


gawsoftpl:
# Resolution
I found issue and create pull request #2632 :

```typescript
private getFileDependencies(file: IFileDescriptorProto): IFileDescriptorProto[] {
    const visited: Set<IFileDescriptorProto> = new Set();
    
    # Correct:
    const toVisit: IFileDescriptorProto[] = [...(this.fileDependencies.get(file) || [])];
    
    # was 
    # const toVisit: IFileDescriptorProto[] = this.fileDependencies.get(file) || [];
    
    while (toVisit.length > 0) {
      # Here loop get item and remove from array but if you 
      # dont spreed fileDendepencies, all depedencies will be removed from original array
      const current = toVisit.pop();

      if (!current || visited.has(current)) {
        continue;
      }

      visited.add(current);
      toVisit.push(...this.fileDependencies.get(current)?.filter((dep) => !visited.has(dep)) || []);
    }

    return Array.from(visited);
  }
```
### Problem description
When I send first request to grpc reflection server everythings works fine, I can generate client. 
But when I reset my client and send second request to grpc server I received error:

```js
protobufjs@7.2.5/node_modules/protobufjs/src/namespace.js:410
        throw Error("no such Type or Enum '" + path + "' in " + this);
              ^
Error: no such Type or Enum 'google.protobuf.Timestamp' in Type .requestshistory.RequestCreateRequestItem

```

### Reproduction steps
```proto
# requestshistory.proto
syntax = "proto3";

package requestshistory;

import "google/protobuf/timestamp.proto";

service RequestsHistoryService {
    rpc Create(RequestCreateRequest) returns (RequestsCreateResponse) {};
}


message RequestsCreateResponse{
    bool success = 1;
}

message RequestCreateRequestItem {
    string user_id = 1;
    string api_server_job_id=2;
    google.protobuf.Timestamp date=3;

}
```
```typescript
# Grpc reflection client
import { GrpcReflection } from 'grpc-js-reflection-client';

const client = new GrpcReflection(host, ChannelCredentials.createInsecure());

// Get services without proto file for specific symbol or file name
const descriptor = await client.getDescriptorByFileName('requestshistory.proto');

// Create package services
const packageObject: any = descriptor.getPackageObject({
  keepCase: true,
  longs: String,
  enums: String,
  defaults: true,
  oneofs: true,
});
```

```typescript
[...]

PR #2632, opened 2023-12-14
Fix issue #2631 [@grpc/grpc-reflection] Issue with proto file dependencies only on second or next request

gawsoftpl:
I found issue:

More info about issue #2631 

```typescript
    
    # Correct:
    const toVisit: IFileDescriptorProto[] = [...(this.fileDependencies.get(file) || [])];
    
    # was 
    # const toVisit: IFileDescriptorProto[] = this.fileDependencies.get(file) || [];
    
    while (toVisit.length > 0) {
      # Here loop get item and remove from array but if you 
      # dont spreed fileDendepencies, all depedencies will be removed from original array
      const current = toVisit.pop();

      if (!current || visited.has(current)) {
        continue;
      }

      visited.add(current);
      toVisit.push(...this.fileDependencies.get(current)?.filter((dep) => !visited.has(dep)) || []);
    }

    return Array.from(visited);
  }
```

### DIFF
--- a/packages/grpc-reflection/src/implementations/reflection-v1.ts
+++ b/packages/grpc-reflection/src/implementations/reflection-v1.ts
@@ -314,8 +314,8 @@ export class ReflectionV1Implementation {
 
   private getFileDependencies(file: IFileDescriptorProto): IFileDescriptorProto[] {
     const visited: Set<IFileDescriptorProto> = new Set();
-    const toVisit: IFileDescriptorProto[] = this.fileDependencies.get(file) || [];
-
+    const toVisit: IFileDescriptorProto[] = [...(this.fileDependencies.get(file) || [])];
+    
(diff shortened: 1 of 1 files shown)


gawsoftpl:
Hi @jtimmons can you review this?

murgatroid99 (approved):
Thank you for the fix.

gawsoftpl:
@murgatroid99 @jtimmons Do you plan create new release for @grpc/reflection 1.0.1 with this fix? Without this server with @grpc/reflection is broken.

murgatroid99:
I merged the PR 5 minutes ago. Just have some patience and the release will go out in a reasonable amount of time.

murgatroid99:
This change has been published in @grpc/reflection 1.0.1

PR #2637, opened 2024-01-04
fix: export type VerifyOptions

chakhsu:
image

image

I think `VerifyOptions` should be export.


murgatroid99 (approved):

murgatroid99:
Thank you for the fix.

PR #2651, opened 2024-01-30
grpc-health-check: Move `typescript` as a dev dependency

satazor:
Currently, `typescript` is installed as a dependency in `grpc-health-check` package, but it's only used during build. This pull-request moves it to a dev dependency.


murgatroid99 (approved):
Thanks for the fix.

satazor:
Tests on windows are failing. Is there anything I need to do?

murgatroid99:
This has been released in grpc-health-check 2.0.1

PR #2653, opened 2024-01-31
Fix memory leak in grpc-js bidi streams

mcollina:
This PR fixes a relatively bad memory leak when the client sends a huge stream of messages to the server, which is slow in processing them. In the current implementation, messages where kept buffered inside `.messageToPush` array. This is pausing the underlining H2 stream, so that the memory leak does not occur.

### DIFF
--- a/packages/grpc-js/src/server-call.ts
+++ b/packages/grpc-js/src/server-call.ts
@@ -863,7 +863,9 @@ export class Http2ServerCallStream<
         await this.pushOrBufferMessage(readable, decompressedMessage);
       }
       pendingMessageProcessing = false;
-      this.stream.resume();
+      if (this.canPush) {
+        this.stream.resume();
+      }
       await maybePushEnd();
     });
 
--- a/packages/grpc-js/test/test-server.ts
+++ b/packages/grpc-js/test/test-server.ts
@@ -228,6 +228,63 @@ describe('Server', () => {
     });
   });
 
+  describe('reading backpressure', () => {
+    let client: ServiceClient;
+    const protoFile = path.join(__dirname, 'fixtures', 'echo_service.proto');
+    const echoService = loadProtoFile(protoFile)
+      .EchoService as ServiceClientConstructor;
+
+    const serviceImplementation = {
+      echo(call: ServerUnaryCall<any, any>, callback: sendUnaryData<any>) {
+        callback(null, call.request);
+      },
+      echoBidiStream(call: ServerDuplexStream<any, any>) {
+        setTimeout(() => {
+          call.resume()
+          call.end();
+
+          const http2Stream = (call as any).call as http2.ServerHttp2Stream;
+          const messagesToPush = (http2Stream as any).messagesToPush as Array<any>;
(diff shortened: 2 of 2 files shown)


ttessarolo (approved):
I've tested the fix in my environment and works like a charm!!

murgatroid99:
#2650 rewrites a lot of this server code, and it includes a change that should fix this bug.

mcollina:
Did you run the test? I don't think that PR overlaps with this area of code.

I think the streaming logic could be simplified further.

murgatroid99:
That PR completely deletes and replaces the `Http2ServerCallStream` class, and the replacement does contain logic that is similar to the logic in this change. However, that means that this test does not work with that PR, because the test code digs into internals that will no longer exist.

murgatroid99:
Specifically, I am referring to [this section](https://github.com/grpc/grpc-node/pull/2650/files#diff-0f49ed72ddd0e55ef565b872be6720238ced5c4547477f8351d24458fa839efdR835-R848)

mcollina:
> However, that means that this test does not work with that PR, because the test code digs into internals that will no longer exist.

I think we can prepare a test that fails in that PR too.

Specifically, there is no backpressure handled when calling [`stream.push()`](https://github.com/grpc/grpc-node/pull/2650/files#diff-20fe25e7ac7f191151695915965b6c4aeb0867fb8ffcd4ff82a54cc9afa83fa1R1573)

murgatroid99:
There is backpressure, but there's a bit of action at a distance. At the interceptor API level, a single call to `startRead` eventually produces a single message or `end` event. `startRead` is called by the stream's `_read` method, which will only be called when the application is ready to read another message.

mcollina:
You are right!

I would put a comment in https://github.com/murgatroid99/grpc-node/blob/eeaa6c0e6e360b4d884f4654c07796f7149645f3/packages/grpc-js/src/subchannel-call.ts#L386, because I think that's the `pause()` that makes it not leak.
Note that this only works because it "automatically" pause after one message and it waits for a subsequent `call.startRead()`.

After taking a look at that whole logic, I think it could be greatly simplified by using `stream.read()` and `on('readable')` instead of `on('data')`. It's clear that you want to process one message at a time, and `stream.read()` will give you that (or null if you need to wait for a readable event). It would likely be significantly faster, and you would not have to  worry about leaks.

---

What's the timeline for the interceptor PR? Might be worthwhile shipping this in a patch.

murgatroid99:
That link points to client code. The code there is similar, but it's not actually involved here. The pause call at a similar level on the server side is [here](https://github.com/grpc/grpc-node/pull/2650/files#diff-0f49ed72ddd0e55ef565b872be6720238ced5c4547477f8351d24458fa839efdR689).

It's not really clear to me that using `read()` and `on('readable')` instead of `pause()`, `resume()`, and `on('data')` would actually result in simpler code, but I'd be happy to consider it as a change after the initial implementation goes out.

I'm hoping to get the server interceptor change out within the next week.

murgatroid99:
Release 1.10.0 includes the server interceptors change.

PR #2658, opened 2024-02-02
grpc-js: Optimizing buffer allocation for server server and bidi streams

mishimastar:
Logical continuation of PR #2652 

Reduced allocated memory while making gRPC grame from **2N+5** bytes to **N+5** bytes (where N is length of serialized message)

P.S. Sorry for big diff, [this file](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/server-interceptors.ts)  does not meet the [prettier conditions](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/prettier.config.js) in `master` branch

### DIFF
--- a/packages/grpc-js/src/server-interceptors.ts
+++ b/packages/grpc-js/src/server-interceptors.ts
@@ -856,14 +856,9 @@ export class BaseServerInterceptingCall
         frame.payload.length
     );
 
-    this.stream.write(
-      frame.infoBytes,
-      this.createFrameWriteErrorCallback(callback)
-    );
-    this.stream.write(
-      frame.payload,
-      this.createFrameWriteErrorCallback(callback)
-    );
+    const cb = this.createFrameWriteErrorCallback(callback);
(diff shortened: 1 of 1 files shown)


mishimastar:
Hope it's more readable at now

packages/grpc-js/src/server-interceptors.ts
```diff
@@ -39,6 +39,11 @@ function trace(text: string) {
   logging.trace(LogVerbosity.DEBUG, TRACER_NAME, text);
 }
 
+export interface GrpcFrame {
```
murgatroid99, line 42: This shouldn't be exported. It's only used in internal APIs in this file.

mishimastar: Fair

murgatroid99:
I'm concerned about the Linux test failures. I don't know why this change would cause some calls to be cancelled.

PR #2663, opened 2024-02-12
Fix deprecated code

Ileriayo:


murgatroid99:
I agree that removing that method call would be good, but leaving a do-nothing statement has the potential to create more confusion. I would remove that line entirely.

Ileriayo:
Please review @murgatroid99

murgatroid99 (approved):

murgatroid99:
Thanks for the improvements

Issue #2671, opened 2024-02-21
@grpc/reflection writes wanings to console for .proto files with no package declaration


paulish:
### Problem description

If proto file has not `package` declaration then call to ReflectionService writes unwanted warnings to console.

### Reproduction steps

Use attached .proto file from archive [dbdata.zip](https://github.com/grpc/grpc-node/files/14358378/dbdata.zip) and the following code
```const PROTO_PATH = path.join(__dirname, 'dbdata.proto');

    const loadOptions = {
        keepCase: true
    };
    const packageDefinition = protoLoader.loadSync(PROTO_PATH, loadOptions);
    const dbDataProto = grpc.loadPackageDefinition(packageDefinition);
    const reflection = new ReflectionService(packageDefinition, loadOptions);
```

### Environment

"@grpc/reflection": "^1.0.1"

### Additional context

The problem is in the `reflection-v1.js` addReference function.
```
            if (ref.startsWith('.')) {
                // absolute reference -- just remove the leading '.' and use the ref directly
                referencedFile = this.symbols[ref.replace(/^\./, '')];
            }
```

It expects to remove the prefix before the package but regular messages has no package prefix and dot should not be removed for them.

paulish:
I have not tried your tests but perhaps a possible solution could be to replace that piece of code with

```            if (ref.startsWith('.') && ref.includes('.', 1)) {
                // absolute reference -- just remove the leading '.' and use the ref directly
                referencedFile = this.symbols[ref.slice(1)];
            }
```
I also changed `ref.replace(/^\./, '')` with `ref.slice(1)` since removing a first dot with a regular expression looks like overkill :)

murgatroid99:
I think that looks reasonable, though I'm guessing that `ref.includes('.', 1)` could just be `ref.length > 1`, and I think `ref.substring(1)` would be preferred over `ref.slice(1)`. I would also think that the part where the relevant entry in `this.symbols` is added might have the same problem, so a similar change would probably be needed there.

@jtimmons You wrote this code, does this change look reasonable to you, or is there something we're missing here?

jtimmons:
> does this change look reasonable to you, or is there something we're missing here?

I think the problem here is actually in the "seek upwards in package scope to find symbol" logic in the second if statement. I managed a pretty minimal reproduction of @paulish's issue by paring his example down to:
```proto
syntax = "proto3";
message TopLevelMessage {
  bool value = 1;
}
message ProcessRequest {
  string key = 1;
[...]

PR #2673, opened 2024-02-22
fix(grpc-reflection): [#2671] handle references to root-level message types in default package

jtimmons:
Bugfix to handle references to root-level message types in the default (unspecified) package from issue #2671. This was causing warnings about failed lookups to any message type that:
  1. Was in the default package (eg. no `package` directive in the proto file), and
  2. Was at the top-level of the package, not nested in another message

The root cause of this was a bug in the `scope` utility where, when we were seeking up the package scope for where the symbol reference lived, we skipped over the root (`.`) package. Relevant snippet from bug report:
> we end up with `TopLevelMessage` being indexed as `.TopLevelMessage` and the reference to it from `ProcessRequest` being sourced from the scope `.ProcessRequest`. When checking the index the search goes like this in order:
>   1. Check current scope (`.ProcessRequest.TopLevelMessage`) which does not exist
>   2. Check parent scope (`TopLevelMessage`) which does not exist
>   3. Bail out because we're at the root already

This change includes:
  1. A fix to the `scope` utility to prevent it from skipping over the root (`.`) package
  3. A fix to how we reference message types to avoid cases like `.` + `.TopLevelMessage`
  4. A reproduction of the issue from #2671 in `unscoped.proto` as well as a regression test in `test-reflection-v1-implementation.ts`

### DIFF
--- a/packages/grpc-reflection/proto/sample/sample.proto
+++ b/packages/grpc-reflection/proto/sample/sample.proto
@@ -3,6 +3,7 @@ syntax = "proto3";
 package sample;
 
 import 'vendor.proto';
+import 'unscoped.proto';
 
 service SampleService {
   rpc Hello (HelloRequest) returns (HelloResponse) {}
--- a/packages/grpc-reflection/proto/sample/unscoped.proto
+++ b/packages/grpc-reflection/proto/sample/unscoped.proto
@@ -1,0 +1,10 @@
+syntax = "proto3";
+
+message TopLevelMessage {
+  bool value = 1;
+}
+
+message ProcessRequest {
+  string key = 1;
+  TopLevelMessage message = 2;
+}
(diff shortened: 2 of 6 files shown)


murgatroid99 (approved):
Thanks for the fix.

murgatroid99:
This has been published in version 1.0.2.

murgatroid99:
This change is not entirely correct, because when a file has no package, its entries in the symbol table have a leading `.`, so removing the period causes the failure to find the symbol. Also, this test doesn't seem to correspond to the warnings that were the originally reported problem in #2671: I can trigger new warnings by adding a service to `unscoped.proto`, but I can't see how to make the test fail even with that change.

murgatroid99:
I think the main issue with the test is that when there are no cross-file references, `addReference` doesn't actually do anything. You get the warning about a failure to find a symbol, but the actual outcome is no different than if it did find the symbol.

murgatroid99:
I think I have a fix. I'll send a PR shortly.

PR #2675, opened 2024-02-23
grpc-js: Add Server#createConnectionInjector API

### LINKED ISSUE #2317: Creating a gRPC service listener on an existing TCP socket

The problem we have is that app A needs to invoke app B via gRPC, but app B is behind a firewall and cannot open ports. In order to do that, we have configured app A to create an ephemeral TCP listener: app B makes an outbound TCP connection to app A's listener, and then the connection is reused so app B starts a gRPC server on that connection.

We (Dapr project) need to support this scenario with app B being written in a variety of languages. App A is written in Go.

We have [implemented this in Go](https://github.com/ItalyPaleAle/dapr-firewall-poc/) using go-grpc. We are able to create a [`net.Listener`](https://github.com/ItalyPaleAle/dapr-go-sdk/compare/main...firewall#diff-e2ee70a6966b66d48fc09ca760684f10e8916467226dfa3b31d59b46f5ad365bR49-R58) that returns an existing `net.Conn`, and the gRPC server can "listen" on that.

I am not able to find a way to create a gRPC listener on top of an existing TCP socket in the grpc-node SDK.

I have not been able to find alternatives to the `grpc.bind` method in this SDK, which always starts a new listener.

murgatroid99:
This implements gRFC L114 (currently in review at grpc/proposal#418). So, it will fix #2317.

### DIFF
--- a/packages/grpc-js/src/server.ts
+++ b/packages/grpc-js/src/server.ts
@@ -74,6 +74,7 @@ import { ServerInterceptingCallInterface, ServerInterceptor, getServerIntercepti
 import { PartialStatusObject } from './call-interface';
 import { CallEventTracker } from './transport';
 import { Socket } from 'net';
+import { Duplex } from 'stream';
 
 const UNLIMITED_CONNECTION_AGE_MS = ~(1 << 31);
 const KEEPALIVE_MAX_TIME_MS = ~(1 << 31);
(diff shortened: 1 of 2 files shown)


murgatroid99:
It's a new API, so the plan is to release it in 1.11

schummar:
Is there a plan yet, when 1.11 will be released? That feature is exactly what I needed to bring BuildKit support to dockerode 😉. So thank you!

murgatroid99:
I was planning on completing some other things before releasing it, but knowing that people need it, I can probably get it out soon.

murgatroid99:
This is now out in version 1.11.0.

Issue #2595, opened 2023-10-10
proto-loader not populating file descriptor dependency list


jtimmons:
### Problem description
The `*Definition` structures that proto-loader returns each have a `fileDescriptorProtos` field which include the files needed to resolve each type. The file descriptor includes a "dependency list" of files that they require but this is currently always returning nothing

This was discussed in #79

### Reproduction steps
This can be reproduced using the following two proto file definitions:
**sample.proto**
```protobuf
syntax = "proto3";
package sample;
import 'vendor.proto';

service SampleService {
  rpc Hello (HelloRequest) returns (HelloResponse) {}
}

message HelloRequest {
  string hello = 1;
}

message HelloResponse {
  vendor.HelloStatus status = 2;
}
```
**vendor.proto**
```protobuf
syntax = "proto3";
package vendor;

enum HelloStatus {
  HELLO = 1;
  WORLD = 2;
}
```

this then inspected using the following code:
```ts
import { FileDescriptorProto } from 'google-protobuf/google/protobuf/descriptor_pb';
import * as protoLoader from '@grpc/proto-loader';

const root = protoLoader.loadSync('sample.proto', { includeDirs: [__dirname] });

Object.values(root).forEach(({ fileDescriptorProtos }) => {
  if (Array.isArray(fileDescriptorProtos)) {
    fileDescriptorProtos.forEach((bin) => {
      const proto = FileDescriptorProto.deserializeBinary(bin);
      console.log(`File: ${proto.getName()}\tDependencies: ${proto.getDependencyList()}`);
      /* output:
File: sample.proto      Dependencies: 
File: vendor.proto      Dependencies: 
File: sample.proto      Dependencies: 
File: vendor.proto      Dependencies: 
File: sample.proto      Dependencies: 
File: vendor.proto      Dependencies: 
      */
    });
  }
});
```

we would expect that `sample.proto` would include `vendor.proto` in its dependency list due to the import and usage of it in the `HelloResponse` field

### Environment
 - OS name, version and architecture: [e.g. Linux Ubuntu 18.04 amd64]: Mac OS 13.4.1
 - Node version [e.g. 8.10.0]: v20.8.0
 - Node installation method [e.g. nvm]: nvm
 - If applicable, compiler version [e.g. clang 3.8.0-2ubuntu4]: N/A
 - Package name and version [e.g. gRPC@1.12.0]: `proto-loader@0.6.9`

### Additional context
[...]

PR #2678, opened 2024-02-27
reflection: Fix references to symbols with no package

### LINKED ISSUE #2671: @grpc/reflection writes wanings to console for .proto files with no package declaration

murgatroid99:
This is a follow up to #2673, and fixes #2671 according to my manual testing. The new comment basically explains this change:

> In files with no package, symbols are populated in the symbols table with a leading period in the key. If there is a package, the symbol does not have a leading period in the key. For simplicity, we check without the period, then with it.

CC @jtimmons

### DIFF
--- a/packages/grpc-reflection/package.json
+++ b/packages/grpc-reflection/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/reflection",
-  "version": "1.0.2",
+  "version": "1.0.3",
   "author": {
     "name": "Google Inc."
   },
--- a/packages/grpc-reflection/proto/sample/sample.proto
+++ b/packages/grpc-reflection/proto/sample/sample.proto
@@ -8,6 +8,7 @@ import 'unscoped.proto';
 service SampleService {
   rpc Hello (HelloRequest) returns (HelloResponse) {}
   rpc Hello2 (HelloRequest) returns (CommonMessage) {}
+  rpc Hello3 (ProcessRequest) returns (TopLevelMessage) {}
 }
 
 service IgnoreService {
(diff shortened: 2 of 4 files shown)


gnossen (approved):

jtimmons:
ah good catch @murgatroid99 - the "absolute vs relative" package references are definitely a bit confusing because we need to handle it on both ends (how it's inserted into the index and then how it's accessed afterwards) and it doesn't seem super consistent across those as to which reference type is used. I think #2595 would solve this as we'd be working off the source-of-truth at that point rather than trying to determine the file relationships after load-time

another thing we could do here is add an option to the service to suppress these warnings? They don't always have a material effect on a reflection client working so in some cases it might just be noise for the user

murgatroid99:
> another thing we could do here is add an option to the service to suppress these warnings? They don't always have a material effect on a reflection client working so in some cases it might just be noise for the user

I was actually thinking that maybe they should be errors instead of warnings. These warnings indicates that API responses will have incorrect dependency information, and it probably indicates a programming error, either on our part or on the part of the reflection library user.

PR #2688, opened 2024-03-13
Add keepalive example

murgatroid99:
This adds an example of configuring client and server keepalive options, modeled on [the Go example](https://github.com/grpc/grpc-go/tree/master/examples/features/keepalive). Note that ping enforcement settings are not included because that functionality is not implemented in Node.

I had to bump the minimum required grpc-js version to 1.10.2 because that is the version that added support for the `grpc.max_connection_idle_ms` server option.

### DIFF
--- a/examples/keepalive/README.md
+++ b/examples/keepalive/README.md
@@ -1,0 +1,16 @@
+# Keepalive
+
+This example illustrates how to set up client-side keepalive pings and
+server-side keepalive pings and connection age and idleness settings.
+
+## Start the server
+
+```
+node server.js
+```
+
+## Start the client
+
+```
+GRPC_TRACE=transport,keepalive GRPC_VERBOSITY=DEBUG client.js
+```
--- a/examples/keepalive/README.md
+++ b/examples/keepalive/README.md
@@ -12,5 +12,5 @@ node server.js
 ## Start the client
 
 ```
-GRPC_TRACE=transport,keepalive GRPC_VERBOSITY=DEBUG client.js
+GRPC_TRACE=transport,keepalive GRPC_VERBOSITY=DEBUG node client.js
 ```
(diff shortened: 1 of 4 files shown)


examples/keepalive/README.md
```diff
@@ -0,0 +1,16 @@
...
+## Start the server
+
+```
+node server.js
+```
+
+## Start the client
+
+```
+GRPC_TRACE=transport,keepalive GRPC_VERBOSITY=DEBUG client.js
```
temawi, line 15: need to include `node` before `client.js`, yes?

examples/keepalive/README.md
```diff
@@ -0,0 +1,16 @@
+# Keepalive
+
+This example illustrates how to set up client-side keepalive pings and
+server-side keepalive pings and connection age and idleness settings.
```
temawi, line 4: readability nit: "...set up client and server keepalive pings as well as connection..."

examples/keepalive/client.js
```diff
@@ -0,0 +1,61 @@
...
+  {keepCase: true,
+   longs: String,
+   enums: String,
+   defaults: true,
+   oneofs: true
+  });
+const echoProto = grpc.loadPackageDefinition(packageDefinition).grpc.examples.echo;
+
+const keepaliveOptions = {
+  // Ping the client every 10 seconds to ensure the connection is still active
```
temawi, line 36: "ping the server", right?

examples/keepalive/README.md
murgatroid99: Honestly, I'm not sure. I'm trying to convey that the keepalive ping settings are on both the client and the server, but the connection age and idleness settings are only on the server. The Go example uses this wording:

> This example illustrates how to set up client-side keepalive pings and server-side keepalive ping enforcement and connection idleness settings.

Which is similar, but also omits the fact that it has server-side keepalive pings.

temawi: Oh I see, on initial read this distinction did not come through, but makes sense now that you mentioned it. If you wanted to be super clear you go with:

"...how to set up keepalive pings on the client-side and keepalive pings, connection age and idleness setting on the server-side"

But I'll leave it up to you.

temawi (approved):

PR #2689, opened 2024-03-14
grpc-js: support content-subtype option in http2 transport

noahziheng:
Hi all,

`HTTP2Transport` has set the `Content-Type` Header as `application/grpc` by default, which library users cannot modify.

The gRPC protocol spec allows using of "content-subtype" just like `application/grpc+proto` at
 <https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests>.

Some gRPC server implementation such as [cloudwego/kitex](https://github.com/cloudwego/kitex) depends on this header to detect the payload codec(like thrift, JSON, etc.).

So this PR has added ~~a new channel option `grpc-node.http_content_subtype`~~ to let end-users modify the subtype if they need it.

---

Updated(2023-03-15): move option to `MetadataOptions` as simpler `contentSubtype`.

### DIFF
--- a/packages/grpc-js/src/channel-options.ts
+++ b/packages/grpc-js/src/channel-options.ts
@@ -62,6 +62,14 @@ export interface ChannelOptions {
    */
   'grpc-node.tls_enable_trace'?: number;
   'grpc.lb.ring_hash.ring_size_cap'?: number;
+  /**
+   * This option will set the "content-subtype" for a request.
+   * For example, a "content-subtype" of "proto" will result in a content-type of
+   * "application/grpc+proto". And this will always be lowercase. See
+   * https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests for
+   * more details.
+   */
+  'grpc-node.http_content_subtype'?: string;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   [key: string]: any;
 }
--- a/packages/grpc-js/src/channel-options.ts
+++ b/packages/grpc-js/src/channel-options.ts
@@ -62,14 +62,6 @@ export interface ChannelOptions {
    */
   'grpc-node.tls_enable_trace'?: number;
   'grpc.lb.ring_hash.ring_size_cap'?: number;
-  /**
-   * This option will set the "content-subtype" for a request.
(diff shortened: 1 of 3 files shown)


noahziheng:
By the way, the current `@grpc/grpc-js@1.10.x` branch is not in sync with the master.

Is this expected?

murgatroid99:
The 1.10.x branch being out of sync with master is just a temporary state. It shouldn't impact this change.

This change should be on master, because it's an API addition. I make an exception for channel arguments that are already defined, but aren't implemented, but this doesn't qualify.

Regarding the change itself: as far as I am aware, gRPC Go is the only implementation that currently supports configuring the content subtype, and it does it with a [call option](https://github.com/grpc/grpc-go/blob/master/Documentation/encoding.md). That makes more sense to me, because the content type is really a property of the individual request, not the whole channel. They also have the concept of default call options for the channel is a whole, which we could also consider using in this library to make configuring that behavior simpler.

noahziheng:
@murgatroid99

Great idea! I have moved the `contentSubtype` option to `MetadataOptions`, the end-user can set it per call now.

On the other hand, if this feature needs to be applied to the master. Should I close this PR, and reopen one to merge the master and wait for you or other maintainers to apply it on the `@grpc/grpc-js@v1.10.x` branch?

murgatroid99:
You can just rebase the changes onto the master branch and then change the PR's branch target to master (or I can do the second part)

noahziheng:
@murgatroid99 

I have rebased and changed the base branch of this PR. Please check it~

murgatroid99:
I discussed this change with my team, and we have decided not to accept it. As far as I know, only Go supports content subtypes, and we do not want to add that behavior in other languages. Instead, we would be willing to consider other solutions that involve other, non-standardized headers. A [gRFC](https://github.com/grpc/proposal) would be needed to formalize the behavior for implementation across the different libraries.

ejona86:
For some of the reasoning on why we don't want to do the subtypes, see https://github.com/grpc/grpc-java/issues/10959#issuecomment-2015352359 . (I think this is the first written record of the issue. I've known about the problem since before grpc-go implemented sub-types, but I didn't learn it was being used there until after-the-fact.)

Issue #2687, opened 2024-03-13
Adding more context to error stack traces


pratik151192:
### Problem description
Hey,

We've been investigating `DEADLINE_EXCEEDED` errors that our clients are facing at [Momento](https://www.gomomento.com/). 
Our clients periodically receive this and we have done a few things that has helped reduce the volume of the deadline errors drastically. Most of these improvements had to do around keepalive settings and/or tweaking other connection settings. We've been mostly enabling gRPC traces with `subchannel, dns_resolver, resolving_load_balancer, keepalive` traces which have been very helpful. 

Our clients still see a burst of these errors, lasting for up-to a minute, and then resolving itself. We have been gathering some event loop/CPU metrics from the clients to correlate any resource contention on the nodeJS process. Meanwhile, we are curious if there are other ways to debug the errors apart from the methods I mentioned above. Now we are at a point where we do not see those gRPC trace logs around the requests indicating that no reconnections took place. Since these logs are verbose, we do not want clients to enable `all` for the traces as it impacts their monitoring bills. 

This ticket is primarily for stack traces (similar to [this](https://github.com/grpc/grpc-node/issues/1810)) that I believe can help us figure out at what point the deadline happened. Going via a [ticket](https://github.com/grpc/grpc-java/issues/9072#issuecomment-1096982926) on this Java SDK, it seems like we can answer a few things based on the stack trace. Currently the stack traces stop at the `onReceiveStatus` method of the underlying gRPC library. I'd expect some stack traces from the request lifecycle when the call actually failed.

We'd love to hear any other ideas you may have for us to debug client-side timeouts/deadline exceeded errors better!

### Reproduction steps
- Set a very low timeout value on any gRPC backed service and see the stack trace always looking similar (I'd expect more traces from `grpc-js`):

```
Error: 4 DEADLINE_EXCEEDED: Deadline exceeded
        at callErrorFromStatus (/Users/pratik/sandbox/js2/client-sdk-javascript/packages/client-sdk-nodejs/node_modules/@grpc/grpc-js/src/call.ts:82:17)
        at Object.onReceiveStatus (/Users/pratik/sandbox/js2/client-sdk-javascript/packages/client-sdk-nodejs/node_modules/@grpc/grpc-js/src/client.ts:360:55)
        at /Users/pratik/sandbox/js2/client-sdk-javascript/packages/client-sdk-nodejs/node_modules/@grpc/grpc-js/src/call-interface.ts:149:27
[...]

PR #2692, opened 2024-03-15
grpc-js: Add more info to deadline exceeded errors

murgatroid99:
Currently, the details string in a client-side deadline exceeded error is just "Deadline exceeded". This change adds more information to that, to help narrow down the cause of those errors. Specifically, it shows a breakdown of any time taken before the request was actually sent to a specific server, along with the address of the server if it was sent. For example, an error string from a test with all information shown looks like this:

```
Deadline exceeded after 1.001s,name resolution: 0.000s,metadata filters: 0.002s,previous attempts: 0,time to current attempt start: 0.001s,LB pick: 0.006s,remote_addr=[::1]:45251
```

This is just a first pass to get some useful information into the error string, and to add the framework to add or change individual pieces of information there as necessary.

This should largely address #2687.

### DIFF
--- a/packages/grpc-js/src/internal-channel.ts
+++ b/packages/grpc-js/src/internal-channel.ts
@@ -684,7 +684,7 @@ export class InternalChannel {
     host: string,
     credentials: CallCredentials,
     deadline: Deadline
-  ): Call {
+  ): LoadBalancingCall | RetryingCall {
     // Create a RetryingCall if retries are enabled
     if (this.options['grpc.enable_retries'] === 0) {
       return this.createLoadBalancingCall(
--- a/packages/grpc-js/src/resolving-call.ts
+++ b/packages/grpc-js/src/resolving-call.ts
@@ -19,6 +19,7 @@ import { CallCredentials } from './call-credentials';
 import {
   Call,
   CallStreamOptions,
+  DeadlineInfoProvider,
   InterceptingListener,
   MessageContext,
   StatusObject,
@@ -27,6 +28,7 @@ import { LogVerbosity, Propagate, Status } from './constants';
 import {
   Deadline,
   deadlineToString,
+  formatDateDifference,
   getRelativeTimeout,
   minDeadline,
 } from './deadline';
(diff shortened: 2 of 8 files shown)


packages/grpc-js/src/internal-channel.ts
```diff
@@ -684,7 +684,7 @@ export class InternalChannel {
     host: string,
     credentials: CallCredentials,
     deadline: Deadline
-  ): Call {
+  ): LoadBalancingCall | RetryingCall {
```
pratik151192, line 687: For my knowledge, I often see either of these or `ResolvingCall` on my `grpc` traces. Curious what the difference between the 3 are and whether they indicate any specific pattern? `RetryingCall` is the one that's a bit confusing as we don't have gRPC retries enabled by default. Does this account for the `transparent` retries?

packages/grpc-js/src/resolving-call.ts
```diff
@@ -97,12 +103,37 @@ export class ResolvingCall implements Call {
...
   private runDeadlineTimer() {
     clearTimeout(this.deadlineTimer);
+    this.deadlineStartTime = new Date();
     this.trace('Deadline: ' + deadlineToString(this.deadline));
     const timeout = getRelativeTimeout(this.deadline);
     if (timeout !== Infinity) {
       this.trace('Deadline will be reached in ' + timeout + 'ms');
       const handleDeadline = () => {
-        this.cancelWithStatus(Status.DEADLINE_EXCEEDED, 'Deadline exceeded');
+        if (!this.deadlineStartTime) {
```
pratik151192, line 112: This is okay but we should never reach here right in lieu of line 106?

packages/grpc-js/src/internal-channel.ts
murgatroid99: Those classes do have tracers set up, specifically `load_balancing_call`, `retrying_call`, and `resolving_call`. You should see them if you have them enabled and you make any requests. All of those are part of the call stack that makes up the internal implementation of a call object. As you say, `RetryingCall` is used by default for transparent retries.

packages/grpc-js/src/resolving-call.ts
murgatroid99: That's true but TypeScript doesn't see a guarantee that the value doesn't change before the timer triggers, so it's easier to just have the check for type safety.

pratik151192:
Hey just checking on this PR where a check is failing for it to be merged

murgatroid99:
That check isn't the problem. It's just waiting on a review. That should happen soon.

gnossen (approved):

PR #2693, opened 2024-03-18
`proto-loader-gen-types`: Import internal files with extension

### LINKED ISSUE #2401: Generated proto types are not valid for ES modules
### Environment
 - Mac m1 2020
 - Node 18.14.2
 - nvm
 - @grpc/grpc-js: "^1.8.13",
 - @grpc/proto-loader: "^0.7.6"

### Problem description
The generated file(s) are not compatible with ES modules.

### Reproduction steps
- Package.json with `"type": "module"`
- tsconfig with `"module": "node16",  "moduleResolution": "node16"`
- generate types using command: 
```
proto-loader-gen-types --longs=String --enums=String --defaults --oneofs --grpcLib=@grpc/grpc-js --outDir=src/generated src/.proto
```

### How to Fix 
The generated import should end with "**.js**":  (See image below)

```typescript
import type { BookItem as _bookStorePackage_BookItem, BookItem__Output as _bookStorePackage_BookItem__Output } from '../bookStorePackage/BookItem.js';
```

### Additional context
Current error:
```bash
error TS2835: Relative import paths need explicit file extensions in EcmaScript imports when '--moduleResolution' is 'node16' or 'nodenext'. Did you mean './bookStorePackage/Book.js'?
```

Proto file:
```typescript
syntax = 'proto3';

package bookStorePackage;

service Book {
    rpc createBook (BookItem) returns (BookItem);
}

message BookItem {
    int32 id = 1;
    string book = 2;
}
```
Image with error:
Screenshot 2023-03-29 at 00 24 13

atjn:
Different runtimes have different requirements for which extension you need to put on relative file imports. It is problematic that the type generator only supports bare imports. This is for example not supported in Deno.

This PR adds an option to specify which file extension should be used.

I have tested the changes in my own repo, and everything works for me.
Closes #2401

### DIFF
--- a/packages/proto-loader/README.md
+++ b/packages/proto-loader/README.md
@@ -114,8 +114,8 @@ Consume the types:
 ```ts
 import * as grpc from '@grpc/grpc-js';
 import * as protoLoader from '@grpc/proto-loader';
-import { ProtoGrpcType } from './proto/example';
-import { ExampleHandlers } from './proto/example_package/Example';
+import type { ProtoGrpcType } from './proto/example.ts';
+import type { ExampleHandlers } from './proto/example_package/Example.ts';
 
 const exampleServer: ExampleHandlers = {
   // server handlers implementation...
--- a/packages/proto-loader/bin/proto-loader-gen-types.ts
+++ b/packages/proto-loader/bin/proto-loader-gen-types.ts
@@ -153,7 +153,7 @@ function getImportLine(dependency: Protobuf.Type | Protobuf.Enum | Protobuf.Serv
       throw new Error('Invalid object passed to getImportLine');
     }
   }
-  return `import type { ${importedTypes} } from '${filePath}';`
+  return `import type { ${importedTypes} } from '${filePath}.ts';`
 }
 
 function getChildMessagesAndEnums(namespace: Protobuf.NamespaceBase): (Protobuf.Type | Protobuf.Enum)[] {
--- a/packages/proto-loader/golden-generated/echo.ts
+++ b/packages/proto-loader/golden-generated/echo.ts
@@ -1,8 +1,8 @@
 import type * as grpc from '@grpc/grpc-js';
 import type { EnumTypeDefinition, MessageTypeDefinition } from '@grpc/proto-loader';
 
-import type { OperationsClient as _google_longrunning_OperationsClient, OperationsDefinition as _google_longrunning_OperationsDefinition } from './google/longrunning/Operations';
-import type { EchoClient as _google_showcase_v1beta1_EchoClient, EchoDefinition as _google_showcase_v1beta1_EchoDefinition } from './google/showcase/v1beta1/Echo';
+import type { OperationsClient as _google_longrunning_OperationsClient, OperationsDefinition as _google_longrunning_OperationsDefinition } from './google/longrunning/Operations.ts';
+import type { EchoClient as _google_showcase_v1beta1_EchoClient, EchoDefinition as _google_showcase_v1beta1_EchoDefinition } from './google/showcase/v1beta1/Echo.ts';
 
 type SubtypeConstructor<Constructor extends new (...args: any) => any, Subtype> = {
   new(...args: ConstructorParameters<Constructor>): Subtype;
--- a/packages/proto-loader/golden-generated/google/api/Http.ts
+++ b/packages/proto-loader/golden-generated/google/api/Http.ts
@@ -1,6 +1,6 @@
 // Original file: deps/googleapis/google/api/http.proto
 
-import type { IHttpRule as I_google_api_HttpRule, OHttpRule as O_google_api_HttpRule } from '../../google/api/HttpRule';
+import type { IHttpRule as I_google_api_HttpRule, OHttpRule as O_google_api_HttpRule } from '../../google/api/HttpRule.ts';
 
 /**
  * Defines the HTTP configuration for an API service. It contains a list of
--- a/packages/proto-loader/golden-generated/google/api/HttpRule.ts
+++ b/packages/proto-loader/golden-generated/google/api/HttpRule.ts
@@ -1,7 +1,7 @@
 // Original file: deps/googleapis/google/api/http.proto
 
-import type { ICustomHttpPattern as I_google_api_CustomHttpPattern, OCustomHttpPattern as O_google_api_CustomHttpPattern } from '../../google/api/CustomHttpPattern';
-import type { IHttpRule as I_google_api_HttpRule, OHttpRule as O_google_api_HttpRule } from '../../google/api/HttpRule';
+import type { ICustomHttpPattern as I_google_api_CustomHttpPattern, OCustomHttpPattern as O_google_api_CustomHttpPattern } from '../../google/api/CustomHttpPattern.ts';
+import type { IHttpRule as I_google_api_HttpRule, OHttpRule as O_google_api_HttpRule } from '../../google/api/HttpRule.ts';
 
 /**
  * # gRPC Transcoding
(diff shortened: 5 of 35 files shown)


murgatroid99:
I don't really understand this. TypeScript is a compiled language. The runtime shouldn't care about the raw TypeScript content, and if it doesn't like the compiler output, that should be fixable by changing the compiler options.

atjn:
It is only valid typescript if you use legacy settings. Modern node resolution settings require that you use `.js` extensions for relative import of typescript files.

From the TS docs: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#type-in-packagejson-and-new-extensions
> Relative import paths need full extensions (we have to write `import "./foo.js"` instead of `import "./foo"`).

 `.js` is chosen to be compatible with node, which would not understand what a `.ts` file is. This choice is different in other runtimes that have native typescript support:

From the Deno docs: https://docs.deno.com/runtime/manual/advanced/typescript/overview#determining-the-type-of-file
> When the extension is absent in a local file, it is assumed to be JavaScript.

That quote means if you try to import a typescript file with no extension, it will be read by a javascript runtime which will fail because it cannot read typescript. There is no option to change this configuration, you must use `.ts` at the end of files.

..so TL;DR if you don't support this feature, your code cannot run in modern Node projects and can not run in Deno at all. I suspect other runtimes will be unsupported as well.

murgatroid99:
If I understand that TypeScript update page correctly, the extension should always be `.ts` because the imports are purely type imports. I would support that change, if it doesn't break anything.

atjn:
I am not sure how you could come to that conclusion. If you use the `.ts` extension in Node, you need to significantly change your compiler config to support that use case. As far as I can tell, it is impossible to even run it without importing third party software that can read and understand the `.ts` extension and transform it into something that Node can run.

I proposed this change because it would be a simple change with no breakage and refactoring. However if you are looking for the most correct solution to this, I think it is to use `.d.ts` extensions for all the files, and also importing them with the `.d.ts` extensions. That extension only allows types which are removed at compile time, and therefore typescript does not need to be Node compliant. This will be a breaking change though because I am sure somebody is relying on the extension to be `.ts`, and will have breakage when all the files change to `.d.ts`. I also am not sure if this tool can sometimes be used to produce typescript that is not allowed in a `.d.ts` file.

murgatroid99:
Node should have no involvement here. These are type-only imports of type-only files. They should be completely omitted in any compiled code that Node would see.

atjn:
Yes, but typescript doesn't know that it is type-only imports because if it was type-only imports you are supposed to use `.d.ts` extensions, so it does not allow import of `.ts` types no matter what is in them.

atjn:
You don't have to take my word for it. Just set up a basic Node repo and try to import a typescript file with `.ts`. Typescript won't allow you.

atjn:
I think I understand now. It seems you are correct, if you use the `type` declarator to show that you are only doing type imports, then `.ts` extensions are also supported in Node. I missed the point on that, sorry :)

So what you are saying is that you prefer to always use the `.ts` extension and you don't want an option to choose/toggle the extension? I don't know if that breaks some backwards compatibility, but I doubt it.

murgatroid99:
You can see the `import type` text in the line you changed in this PR.

It looks like it is possible for this code to use a single extension that works everywhere, and if so, we should just do that. Then providing an option to use a different extension is superfluous.

atjn:
Should be ready for merge now :)

murgatroid99:
Please run the `generate-golden` npm script in the proto-loader package.

atjn:
I also updated the example in the docs to use the same import syntax :)

murgatroid99 (approved):

atjn:
Thanks for helping me get this merged!

murgatroid99:
This is now out in version 0.7.11

versat:
Yesaterday we suddenly observed new eslint errors in our (unchanged, only rebuilt) code. It took us some time to find out that `@grpc/proto-loader` 0.7.11 caused it. 
Errors look like this:
```
D:\project\src\grpcClient.ts
  339:9   error  Unsafe assignment of an `any` value              @typescript-eslint/no-unsafe-assignment
  521:36  error  Unsafe member access .version on an `any` value  @typescript-eslint/no-unsafe-member-access

D:\project\src\grpcCommandHandler.ts
  37:15  error  Unsafe member access .serverPort on an `any` value                                            @typescript-eslint/no-unsafe-member-access
  38:55  error  Invalid type "any" of template literal expression                                             @typescript-eslint/restrict-template-expressions
  38:55  error  Unsafe member access .serverPort on an `any` value                                            @typescript-eslint/no-unsafe-member-access
  39:21  error  Unsafe argument of type `any` assigned to a parameter of type `number | PromiseLike<number>`  @typescript-eslint/no-unsafe-argument
  39:21  error  Unsafe member access .serverPort on an `any` value                                            @typescript-eslint/no-unsafe-member-access
  40:22  error  Unsafe member access .errorCode on an `any` value                                             @typescript-eslint/no-unsafe-member-access
  41:55  error  Invalid type "any" of template literal expression                                             @typescript-eslint/restrict-template-expressions
  41:55  error  Unsafe member access .errorCode on an `any` value                                             @typescript-eslint/no-unsafe-member-access
```
This is how we are calling proto-loader-gen-types:
`npx proto-loader-gen-types --grpcLib=@grpc/grpc-js --outDir=src/api/ ../../grpc/*.proto`

These are the most relevant parts for importing the grpc stuff:
```js
import * as protoLoader from '@grpc/proto-loader';

import { ProtoGrpcType as DMErrorCode } from './api/dm_errorcode';
import { ProtoGrpcType as UEDM } from './api/UEDM';
import { UEDMClient } from './api/UE/DM/UEDM';
```
I guess we have to change something so the types are known again? But I am not sure what exactly. Do you have any idea?
Currently our workaround is to lock `@grpc/proto-loader` to 0.7.10, but this is not a solution.

atjn:
@versat have you tried importing the types as types?
```diff
-import { ProtoGrpcType as DMErrorCode } from './api/dm_errorcode';
+import type { ProtoGrpcType as DMErrorCode } from './api/dm_errorcode.ts';
-import { ProtoGrpcType as UEDM } from './api/UEDM';
+import type { ProtoGrpcType as UEDM } from './api/UEDM.ts';
-import { UEDMClient } from './api/UE/DM/UEDM';
+import type { UEDMClient } from './api/UE/DM/UEDM.ts';
```

The long term solution is definitely that this tool should not be producing incorrect types.

versat:
I have adapted the imports, but now I am getting even more of these errors.
Currently we are using typescript 4.9.5.
I have updated it to 5.4.3 because I saw that you created a PR for updating Typescript in this repository to fix something.
And with Typescript 5.4.3 eslint is happy now too :)
So, is it a hard requirement (now) to use Typescript >= 5?
Maybe this can be added as a requirement?

murgatroid99:
I'm reverting this. It shouldn't be a minor change if it breaks compatibility with TypeScript 4. We can re-evaluate what the best approach is after that.

atjn:
@murgatroid99 I can make a boolean that chooses between no extension and `.ts` extension. You can leave it off by default, then enable it by default and deprecate it in the next major version.

murgatroid99:
I think it could also be that a `.js` extension will work everywhere. I would want to get a better understanding of the practical consequences of these import statement variations before choosing a specific solution.

ClayenKitten:
Hey, @murgatroid99, sorry to bother, but is there any progress on that one?

Maybe we could implement `--extension` CLI option with `none`, `js` and `ts` options? `none` would be the default for the time being, with possible change to another default after further investigation on the matter. I’d be happy to work on this if such a solution would be accepted.

Issue #2625, opened 2023-11-29
Calls trigger ERR_HTTP2_SESSION_ERROR with node 20.9.0 


AlbertLlebaria:
### Problem description
When running multiple gRPC requests asynchronously with the same gRPC client, gRPC client at some point starts cancelling thoe calls with the following error message:
_Received RST_STREAM with code 2 triggered by internal client error: Session closed with error code 2_

I've enabled the following ENV variables `GRPC_TRACE=all` and `GRPC_VERBOSITY=DEBUG` and these are the logs for one of the calls that received the RST_STREAM  error.

```
D 2023-11-28T11:52:30.785Z | subchannel_call | [5567] write() called with message of length 433
D 2023-11-28T11:52:30.785Z | subchannel_call | [5567] sending data chunk of length 433
D 2023-11-28T11:52:31.297Z | subchannel_call | [5567] end() called
D 2023-11-28T11:52:31.297Z | subchannel_call | [5567] calling end() on HTTP/2 stream
D 2023-11-28T11:52:35.382Z | subchannel_call | [5567] Node error event: message=Session closed with error code 2 code=ERR_HTTP2_SESSION_ERROR errno=Unknown system error undefined syscall=undefined
D 2023-11-28T11:52:35.393Z | subchannel_call | [5567] HTTP/2 stream closed with code 2
D 2023-11-28T11:52:35.393Z | subchannel_call | [5567] ended with status: code=13 details="Received RST_STREAM with code 2 triggered by internal client error: Session closed with error code 2"
```
Right before the calls start triggering the `ERR_HTTP2_SESSION_ERROR` , i found the following a log explaining the client got closed by GOAWAY  with code 2. `0.0.0.0:3000 connection closed by GOAWAY with code 2`.

```
D 2023-11-28T11:52:35.354Z | resolving_call | [6070] Received status
D 2023-11-28T11:52:35.354Z | resolving_call | [6070] ended with status: code=0 details="OK"
D 2023-11-28T11:52:35.354Z | subchannel_call | [6202] HTTP/2 stream closed with code 0
D 2023-11-28T11:52:35.354Z | subchannel_call | [6424] Received server trailers:
		grpc-status: 0
		grpc-message: OK

D 2023-11-28T11:52:35.354Z | subchannel_call | [6424] received status code 0 from server
D 2023-11-28T11:52:35.354Z | subchannel_call | [6424] received status details string "OK" from server
[...]

PR #2695, opened 2024-03-20
grpc-js: Avoid sending redundant RST_STREAMs from the client

murgatroid99:
Based on the hypothesis that #2625 is caused by RST_STREAM rate limiting (for rapid reset mitigation), this change reduces the chances of that by avoiding sending an RST_STREAM from the client when the server has already half-closed its side of the connection.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/grpc-js",
-  "version": "1.10.3",
+  "version": "1.10.4",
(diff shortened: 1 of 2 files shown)


sergiitk (commented):
Was handling a similar-ish situation in java recently - preventing multiple RSTs when the on server failures. The difference in my case was needing to send a reset when our server transport failed for non-io reasons. https://github.com/grpc/grpc-java/pull/11016

This LGTM, but let's chat for a few minutes before I approve this.

sergiitk (approved):

PR #2696, opened 2024-03-21
grpc-js: Ensure server interceptors work with builder utility classes

murgatroid99:
The idea with the part of the code modified here is that the user can pass in a partially filled in responder or server listener, and the rest of the object will be filled in with defaults so that the rest of the class can handle it more simply. Unfortunately, the specific strategy of doing that using the spread operator (`...`) works with unpopulated object fields, but not object fields set to `undefined`, which is what `ResponderBuilder` and `ServerListenerBuilder` produce.

The test auth interceptor partially populates both a listener and a resolver, so switching it to use the builders should ensure that it catches this problem on both sides.


sergiitk (approved):

sergiitk:
Makes sense

PR #2699, opened 2024-03-27
proto-loader: Revert 'Import internal files with extension'

murgatroid99:
Revert #2693. This change broke xDS interop tests.


murgatroid99:
The test failures are caused by the bug this change is fixing. It's using the released package, not the local one.

sergiitk (approved):

murgatroid99:
Superseded by #2700.

murgatroid99:
I am actually doing this. It broke other people's builds too.

PR #2711, opened 2024-04-08
grpc-loader: Expose method options

n0v1:
This PR adds support for [MethodOptions](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/descriptor.proto#L887-L919) to proto-loader.

It is a continuation of #2230 (originally written by @hiepthai) which tries to address the comments of @murgatroid99.

### DIFF
--- a/packages/proto-loader/src/index.ts
+++ b/packages/proto-loader/src/index.ts
@@ -122,18 +122,18 @@ export enum IdempotencyLevel {
 }
 
 export interface NamePart {
-  namePart: string;
-  isExtension: boolean;
+  name_part: string;
+  is_extension: boolean;
 }
 
 export interface UninterpretedOption {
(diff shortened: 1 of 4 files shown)


packages/proto-loader/src/index.ts
```diff
@@ -126,6 +154,7 @@ export interface MethodDefinition<RequestType, ResponseType, OutputRequestType=R
   originalName?: string;
   requestType: MessageTypeDefinition;
   responseType: MessageTypeDefinition;
+  options?: MethodOptions;
```
murgatroid99, line 157: This does not need to be optional, so it should not be optional.

murgatroid99 (approved):
Thank you for fixing this up. It looks good now.

Issue #2325, opened 2023-01-13
Impossible to set custom User Agent


liamcottle:
### Problem description

It seems that it's not possible to use a custom user agent exactly as it is provided. The library always adds its own user agent to the beginning or end, depending on whether we use `grpc.primary_user_agent` or `grpc.secondary_user_agent` options.

Looking in `transport.ts`, this is the code that decides what the user agent should be:

https://github.com/grpc/grpc-node/blob/%40grpc/grpc-js%401.8.x/packages/grpc-js/src/transport.ts#L128

```
this.userAgent = [
  options['grpc.primary_user_agent'],
  `grpc-node-js/${clientVersion}`,
  options['grpc.secondary_user_agent'],
]
```

It doesn't matter how I provide my user agent, `grpc-node-js/x.x.x` will always be injected into my user agent.

I'm connecting to a service which enforces that user agents must match a specific pattern, or a list of allowed user agents. I'm not able to change this list of user agents, so having this internal user agent injected into to the custom user agent I set makes it impossible to use `grpc-js` to connect.

Is there any way to override this without forking the project, removing the internal user agent, and releasing my own version of the package?

### Reproduction steps

Setting the user agent via any of the following code snippets still results in the internal user agent being injected.

```
// results in user agent being "ExampleUserAgent/1.2.3.4 grpc-node-js/x.x.x"
const client = new ExampleService("api.example.com", credentials, {
    "grpc.primary_user_agent": "ExampleUserAgent/1.2.3.4",
});
```

```
// also results in user agent being "ExampleUserAgent/1.2.3.4 grpc-node-js/x.x.x"
const client = new ExampleService("api.example.com", credentials);
var metadata = new grpc.Metadata();
metadata.add('user-agent', 'ExampleUserAgent/1.2.3.4');
const ret = client.ExampleMethod({}, metadata, (error, result) => {
    // ...
});
```

### Environment

 - OS name, version and architecture: macOS 12.4
 - Node version: v18.12.1
 - Node installation method: unsure
 - Package name and version: @grpc/grpc-js v1.8.4

murgatroid99:
This is intended behavior. From our point of view, the server is either behaving unreasonably or deliberately announcing an incompatibility with this client. Either way, you should talk to the server's owner about changing it.

liamcottle:
No problem!

I feel it's pretty reasonable to allow the developer consuming this library to specify an exact user agent, and have it use the internal user agent as a default if no user agent was specified.

[...]

PR #2728, opened 2024-04-25
grpc-js: Enable custom user-agent configuration in metadata

ronork:
Issue: The current implementation of the grpc-js library imposes a default 'user-agent' header in gRPC calls, restricting users who require customization to identify their clients with specific user-agent information.

Resolution: This PR addresses the issue by introducing functionality to set a custom user-agent header for gRPC calls. It allows users to provide their own user-agent strings in the call metadata, enhancing flexibility and customization options.

### DIFF
--- a/examples/interceptors/README.md
+++ b/examples/interceptors/README.md
@@ -1,0 +1,120 @@
+# Interceptor
+
+Node gRPC provides simple APIs to implement and install interceptors on clients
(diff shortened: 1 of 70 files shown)


murgatroid99:
This is not an option we want to provide to users. See also #2325 and #1668.

PR #2746, opened 2024-05-10
HTTP CONNECT: handle early server packets

### LINKED ISSUE #2744: RST_STREAM with code 2 / system error -505 when using HTTP CONNECT proxy
### Problem description

When using an HTTP CONNECT proxy, it is possible for the server's first HTTP2 packet to reach the client before the latter sends its own first HTTP2 packet. If that happens, then the client terminates the connection (HTTP2 GOAWAY with code 2 and system error -505) and fail. Other gRPC implementations that we tested (Go, Java, and the Tokio third party implementation in Rust) don't have issue in that same scenario.

For example, in the capture below, we can see the server's first packet, at line 19, while the client's first packet, including the HTTP2 magic string is at line 21. Then, at line 37, we can see that the client side is initiating a GOAWAY, with code 2.

Screenshot 2024-05-08 at 13 20 20

With gRPC and node's tracing enabled, we can see the following interesting messages (this is a truncated snippet; see the attached zip file for the complete log and corresponding pcap).

```
Http2Session client (11473) Error 'Remote peer returned unexpected data while we expected SETTINGS frame.  Perhaps, peer does not support HTTP/2 properly.'
HttpStream 1 (11479) [Http2Session client (11473)] sending rst_stream with code 2
Http2Session client (11473) frame type 3 was not sent, code: -530
D 2024-05-09T18:52:34.462Z | v1.10.7 718 | subchannel_call | [571] Node error event: message=Protocol error code=ERR_HTTP2_ERROR errno=Unknown system error -505 syscall=undefined
D 2024-05-09T18:52:34.462Z | v1.10.7 718 | subchannel_call | [571] HTTP/2 stream closed with code 2
D 2024-05-09T18:52:34.462Z | v1.10.7 718 | subchannel_call | [571] ended with status: code=13 details="Received RST_STREAM with code 2 triggered by internal client error: Protocol error"
```

### Reproduction steps

1. Clone the repro from GitHub

```console
git clone https://github.com/mjameswh/grpcjs-proxy-first-packet-repro.git
cd grpcjs-proxy-first-packet-repro
```

2. Install npm dependencies.

```console
$ npm i
```

3. Start the gRPC server and proxy in background; alternatively, you may run these in a distinct tabs.

```console
$ node src/greeter_server.js --port 6233 &
$ go run go/httpproxy.go --port 8888 &
```

4. Run the client 1000 times, without proxy

```console
$ node src/greeter_client.js --target 127.0.0.1:6233 --repeat 1000
Completed without error
```

5. Run again, this time with proxy

```console
$ export grpc_proxy=http://127.0.0.1:8888
$ node src/greeter_client.js --target 127.0.0.1:6233 --repeat 1000
Error: 13 INTERNAL: Received RST_STREAM with code 2 triggered by internal client error: Protocol error
(**redacted for brievety; see full error details in linked gist**)
{
  code: 13,
  details: 'Received RST_STREAM with code 2 triggered by internal client error: Protocol error',
  metadata: Metadata { internalRepr: Map(0) {}, options: {} }
}
Failed at iteration 43
```

6. Restart the proxy, this time with the "wait for client's first packet" patch

```console
$ killall  httpproxy
$ go run go/httpproxy.go --port 8888 --wait-for-first-client-packet &
```

7. Run the client 1000 times, with the patched proxy

```console
$ export grpc_proxy=http://127.0.0.1:8888
$ node src/greeter_client.js --target 127.0.0.1:6233 --repeat 1000
```

### Environment

 - OS name: tested on MacOS 14.2.1 (Apple Silicon M2)
 - Node version: Node 20.12.2
 - Node installation method: volta
 - Package name: @grpc/grpc-js v1.10.7

### Additional context

I only have limited knowledge of the HTTP2 protocol internals, but as far as I understand, though unlikely, the fact that the server's first packet reaches the client before the client its first packet is technically a legit situation, and could also happen in some non-proxied scenarios (e.g. using TCP Fast Open).

Therefore, this might arguably be IMHO an oversight in the underlying HTTP2 implementation, rather than a bug in grpc-node's proxy support; still, I'm reporting this here since I wouldn't be able to debate this further without considerably more digging of my part.

### Attached files

- [grpcjs-proxy-first-packet-repro-capture.zip](https://github.com/grpc/grpc-node/files/15266889/grpcjs-proxy-first-packet-repro-capture.zip)

mjameswh:
When using an HTTP CONNECT proxy, it is possible for the server's first HTTP2 packet to reach the client before the node's HTTP library even starts parsing the HTTP `HTTP/1.1 200 Connected` response.

When that happens, node's HTTP response parser may take these extra bytes (i.e. the server's first HTTP2 packets) out of the socket into its parsing buffer. That means that these extra bytes will not be emitted through the tunneled socket provided on the `connect` event. Instead, these extra bytes are provided through [the `head` parameter of the `connect` event](https://nodejs.org/docs/latest/api/http.html#event-connect_1).

This PR reinjects these extra bytes (if any) into the tunneled socket.

Fixes #2744.

### DIFF
--- a/packages/grpc-js/src/http_proxy.ts
+++ b/packages/grpc-js/src/http_proxy.ts
@@ -233,6 +233,12 @@ export function getProxiedConnection(
             ' through proxy ' +
             proxyAddressString
         );
+        // The HTTP client may have already read a few bytes of the proxied
+        // connection. If that's the case, put them back into the socket.
+        // See https://github.com/grpc/grpc-node/issues/2744.
+        if (head.length > 0) {
+          socket.unshift(head);
+        }
         if ('secureContext' in connectionOptions) {
           /* The proxy is connecting to a TLS server, so upgrade this socket
            * connection to a TLS connection.


murgatroid99:
This change looks good, but I can't merge it until you sign the CLA.

murgatroid99:

PR #2752, opened 2024-05-20
enable esModuleInterop

### LINKED ISSUE #2617: Autogenerated code erro: Cannot use namespace 'Long' as a type
### Problem description
The autogenerated type code from `@grpc/grpc-js` is generating code that is not valid typescript.

For example, the `Long` that is imported causes an error:
CleanShot 2023-11-15 at 07 18 39@2x

The way to fix this is by using `typeof Long` further below in the code like so:
CleanShot 2023-11-15 at 07 19 34@2x

### Reproduction steps
Work in progress

This is what my settings for generating types looks like using @grpc/proto-loader
https://www.npmjs.com/package/@grpc/proto-loader#generating-typescript-types

```sh
yarn run proto-loader-gen-types \
  -I $PROTOS_DESTINATION_DIR/serviceProto \
  -I $PROTOS_DESTINATION_DIR \
  --keepCase \
  --longs=String \
  --enums=String \
  --defaults \
  --oneofs \
  --grpcLib=@grpc/grpc-js \
  --outDir=$PROTOS_DESTINATION_DIR \
  $PROTOS_DESTINATION_DIR/serviceProto/proto/com/linkedin/euler/controlplane/*.proto

```

### Environment
OS Name: Sonomo 14.1.1
Node version 20.5.0
Typrscript: 5.0.4
@grpc/grpc-js": "1.9.9",
@grpc/proto-loader": "0.7.10",

idranme:
Fixes #2617

### DIFF
--- a/packages/proto-loader/src/index.ts
+++ b/packages/proto-loader/src/index.ts
@@ -16,13 +16,13 @@
  *
  */
 
-import camelCase = require('lodash.camelcase');
+import camelCase from 'lodash.camelcase';
 import * as Protobuf from 'protobufjs';
 import * as descriptor from 'protobufjs/ext/descriptor';
 
 import { loadProtosWithOptionsSync, loadProtosWithOptions, Options, addCommonProtos } from './util';
 
-import Long = require('long');
+import Long from 'long';
 
 export { Options, Long };
 
--- a/packages/grpc-health-check/gulpfile.ts
+++ b/packages/grpc-health-check/gulpfile.ts
@@ -16,8 +16,8 @@
  */
 
 import * as gulp from 'gulp';
-import mocha from 'gulp-mocha';
-import execa from 'execa';
+import * as mocha from 'gulp-mocha';
+import * as execa from 'execa';
 import * as path from 'path';
 
 const healthCheckDir = __dirname;
--- a/packages/grpc-health-check/gulpfile.ts
+++ b/packages/grpc-health-check/gulpfile.ts
@@ -16,8 +16,8 @@
  */
 
 import * as gulp from 'gulp';
-import * as mocha from 'gulp-mocha';
-import * as execa from 'execa';
+import mocha from 'gulp-mocha';
+import execa from 'execa';
 import * as path from 'path';
 
 const healthCheckDir = __dirname;
--- a/packages/grpc-health-check/test/test-health.ts
+++ b/packages/grpc-health-check/test/test-health.ts
@@ -16,7 +16,7 @@
  *
  */
 
-import assert from 'assert';
+import * as assert from 'assert';
 import * as grpc from '@grpc/grpc-js';
 import { HealthImplementation, ServingStatusMap, service as healthServiceDefinition } from '../src/health';
 import { HealthClient } from './generated/grpc/health/v1/Health';
--- a/packages/grpc-health-check/test/test-health.ts
+++ b/packages/grpc-health-check/test/test-health.ts
@@ -16,7 +16,7 @@
  *
  */
 
-import * as assert from 'assert';
+import assert from 'assert';
 import * as grpc from '@grpc/grpc-js';
 import { HealthImplementation, ServingStatusMap, service as healthServiceDefinition } from '../src/health';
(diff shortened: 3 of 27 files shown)


packages/proto-loader/src/index.ts
```diff
@@ -16,13 +16,13 @@
...
 
-import camelCase = require('lodash.camelcase');
+import camelCase from 'lodash.camelcase';
 import * as Protobuf from 'protobufjs';
 import * as descriptor from 'protobufjs/ext/descriptor';
 
 import { loadProtosWithOptionsSync, loadProtosWithOptions, Options, addCommonProtos } from './util';
 
-import Long = require('long');
+import Long from 'long';
```
idranme, line 25: This is the key to solving the problem.

murgatroid99:
It looks like the tests are failing because the tsconfig also affects the `gulpfile.ts`.

idranme:
> It looks like the tests are failing because the tsconfig also affects the `gulpfile.ts`.

The issue is affected by https://github.com/grpc/grpc-node/blob/master/tsconfig.json

murgatroid99:
The changes in your last commit are a lot more extensive than I expected. Why did you need to change anything outside of the `packages/proto-loader` directory?

idranme:
> The changes in your last commit are a lot more extensive than I expected. Why did you need to change anything outside of the `packages/proto-loader` directory?

Since CI is run in the project root directory, it is affected by tsconfig.json in the project root directory. When modifying tsconfig.json in the project root directory, tests for other packages are also affected.

murgatroid99:
Why did you modify the tsconfig.json in the root directory? Did you try only changing files in the `packages/proto-loader` directory?

idranme:
> Why did you modify the tsconfig.json in the root directory? Did you try only changing files in the `packages/proto-loader` directory?

At https://github.com/grpc/grpc-node/pull/2752/commits/07711a6fe4c366d1c5464de984a601b658301f21, I only changed `packages/proto-loader`, so https://btx.cloud.google.com/invocations/3548d71a-bb98-4ede-b2a6-746dca856424/targets/grpc%2Fnode%2Fpull_request%2Fwindows;config=default/log

murgatroid99:
The error in that log occurred in `packages/proto-loader/gulpfile.ts`. Can you please try making this change modifying only files in the `packages/proto-loader` directory, including `packages/proto-loader/gulpfile.ts`.

idranme:
> The error in that log occurred in `packages/proto-loader/gulpfile.ts`. Can you please try making this change modifying only files in the `packages/proto-loader` directory, including `packages/proto-loader/gulpfile.ts`.

Previously, the modification of packages/proto-loader/gulpfile.ts was due to changes in packages/proto-loader/tsconfig.json. The error occurred because the CI was running in the root directory and was using the root tsconfig.json, rather than the one in packages/proto-loader/.

idranme:
> The error in that log occurred in `packages/proto-loader/gulpfile.ts`. Can you please try making this change modifying only files in the `packages/proto-loader` directory, including `packages/proto-loader/gulpfile.ts`.

I will try to solve the problem by following the method you suggested.

murgatroid99:
I take back what I said before about changing only files in the `packages/proto-loader` directory. I thought that your original commit didn't change `packages/proto-loader/gulpfile.ts`, so I misunderstood what the problem was.

I understand the problem now: the `packages/proto-loader/gulpfile.ts` can be executed either on its own or from the root gulpfile. As a result it has to be compatible with both of those `tsconfig.json` files, which means that they both need the `esModuleInterop` option. That then propagates to the other gulpfiles.

idranme:
Should I revert those two comments?

murgatroid99:
yes, I think you should.

idranme:
PTAL

murgatroid99:
I did some manual testing and I don't think we can make this change. My testing indicates that if a user of this library doesn't have the `esModuleInterop` option enabled, their own build fails with with this change. That's not apparent with the automated tests because we had to change every local user of the library to enable that option for other reasons.

Issue #2734, opened 2024-05-02
grpc-js servers not sending keepalives


davidfiala:
I've noticed that grpc-js servers do not *seem* to be sending keepalives to gRPC clients when they are connected to a long-lived stream. (But only within my larger applications so far)

I'm very happy to create a tiny PoC and then upload it here, but before I build it up, I wanted to clarify if grpc-js ChannelOptions were in fact intended for gRPC servers to send PINGs to clients. Or if it was only meant to be clients and I misunderstood the docs. I can confirm that clients are happily sendings PINGs as expected in `1.10.6`

For example, I've instrumented the grpc-js server.ts file to log when pings may happen, but I don't see it. Server has long-lived bidi streams from a client and the server is configured with:

```
createServer({
    'grpc.keepalive_time_ms': args.grpc_server_keepalive_ms,
    'grpc.keepalive_timeout_ms': args.grpc_server_keepalive_timeout_ms,
    'grpc.keepalive_permit_without_calls': 1,
  })
```

... and no pings are sent and no keepalive disconnects will occur, in my app so far.

From https://grpc.io/docs/guides/keepalive/ it looks like it has under the column `Availability` both `Client and Server` for keepalives. And https://github.com/grpc/grpc/blob/master/doc/keepalive.md says that if "if the ping is not acknowledged by **the peer**".

I haven't ruled out a bug in my own setup yet, but I would like to verify that the intention is that both either/or the client/server can independently set the keep alives and transmit their own PINGs. I'll be happy to make a PoC and continue this bug with my findings if yes.

Thank you for your time!

murgatroid99:
Here's my guess for why this bug wasn't noticed:

 - The keepalive functionality was added to this server implementation relatively recently. On top of that, keepalives are less impactful on the server than on the client, because the client is responsible for reconnecting when a connection drops. As a result, it's likely that few users have keepalives enabled at all.
 - Node can detect and report connection drops on its own, so failing to detect keepalive errors only matters when the connection drop is also not detected in any other way.
 - The persistence of stale connections and even stale streams should be relatively unimpactful on the server. You might see a bit of memory leakage, or a stream that has become inactive, but often not visible failures. This is as opposed to on the client, where a stale connection can prevent the client from making requests and/or being responsive.

murgatroid99:
[...]

PR #2756, opened 2024-05-28
Serverside keepalive error detection and cleanups

davidfiala:
Improve server-side keepalives, possibly resolve bug where keepalive errors were not being treated as errors.

- Ensure that http2 ping failures are detected at request time and in the callback. Code inspection indicates both cases may have silently been ignored before.
- Ensure keepalive intervals and timeouts are always cleared.
- Bug Fix: keepalive intervals were being cleared with an incorrect clearTimeout before. Not sure if this was causing intervals leaks in some nodejs impls or not. (v20.13.1 seems to accept this mismatch without issue)
- Rename timers/intervals for clarity
- Ensure all paths do contain a channelz keepalive trace message with unique error strings.

Note: I think that in https://github.com/nodejs/node/issues/18447 half of the issue was clarified already, but the code just wasn't updated to reflect the answer there. The other half is checking the return value of `session.ping` for a non-true result, too. Given my experience of keepalives not working in https://github.com/grpc/grpc-node/issues/2734 my current conclusion is that although keepalives are being sent (I inspected via manual logging), failures may not have been properly detected.

### DIFF
--- a/packages/grpc-js/src/server.ts
+++ b/packages/grpc-js/src/server.ts
@@ -1459,6 +1459,26 @@ export class Server {
           session.destroy();
         }
       }, this.keepaliveTimeMs).unref?.();
+      session.once('goaway', (errorCode, opaqueData) => {
+        if (keepaliveInterval) {
+          if (this.channelzEnabled) {
+            if (errorCode === http2.constants.NGHTTP2_ENHANCE_YOUR_CALM) {
+              this.channelzTrace.addTrace(
+                'CT_INFO',
+                `Client sent GOAWAY for too many pings ${clientAddress}`
+              );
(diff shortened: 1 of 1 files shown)


murgatroid99:
Unfortunately, the master branch is actually behind the release branch on this, so your changes are going to conflict with that. How about I merge those changes into master and then we look at this again?

On the plus side, that version does already handle the ping error, so you won't need to do that again. On the minus side, there are two parallel implementations of keepalives for performance reasons, so any changes will need to be made twice.

murgatroid99:
The current release branch is `@grpc/grpc-js@1.10.x`. You can make the change on that branch.

Issue #2139, opened 2022-06-16
grpc session ping when session is destoryed by server


zhtswang:
### Problem description
When a session call ping, if the session is destoryed by the server. it will throw an error.
Error [ERR_HTTP2_INVALID_SESSION]: The session has been destroyed
at new NodeError (internal/errors.js:322:7)
at ClientHttp2Session.ping (internal/http2/core.js:1317:13)
at Subchannel.sendPing (D:\app\node_modules\@grpc\grpc-js\src\subchannel.ts:367:19)
at Timeout._onTimeout (D:\app\node_modules\@grpc\grpc-js\src\subchannel.ts:376:12)
at listOnTimeout (internal/timers.js:557:17)
at processTimers (internal/timers.js:500:7)

emmansun:
How about to check session!.destroyed first?

```typescript
  private sendPing() {
    if (this.session!.destroyed) {
      this.keepaliveTrace('Session was destroyed');
      this.handleDisconnect();
      return;
    }
    if (this.channelzEnabled) {
      this.keepalivesSent += 1;
    }
    this.keepaliveTrace('Sending ping with timeout ' + this.keepaliveTimeoutMs + 'ms');
    this.keepaliveTimeoutId = setTimeout(() => {
      this.keepaliveTrace('Ping timeout passed without response');
      this.handleDisconnect();
    }, this.keepaliveTimeoutMs);
    this.keepaliveTimeoutId.unref?.();
    this.session!.ping(
      (err: Error | null, duration: number, payload: Buffer) => {
        this.keepaliveTrace('Received ping response');
        clearTimeout(this.keepaliveTimeoutId);
      }
    );
  }
```

murgatroid99:
This should be fixed in version 1.6.9. Please try it out.

emmansun:
Thx @murgatroid99 , we'll try it out.

zhtswang:
@murgatroid99, we have applied the version to our production environment and monitor one more week, no such issue happened again. Please close the issue.

PR #2760, opened 2024-05-28
Keepalive bugfixes and unify timers strategies between client and server

davidfiala:
Improve server-side keepalives, possibly resolve bug where keepalive errors were not being treated as errors.

- Bugfix: Ensure that if session.ping returns false we correctly identify fail the keepalive and connection
- Bugfix: Ensure that if the interval between keepalives being sent occurs faster than the prior keepalive's timeout that we do not overwrite the reference to the prior timeout. Before this change: we could have in theory prevented a valid keepalive timeout from clearing itself because it's object reference was replaced with a newer timeout. This rewrite keeps every timeout as a local (vs a shared state per session). Even if the timeout outlives the lifetime of a session, we still guard against errors by checking that the parent interval is not falsy. I reckon this could result in a short-term memory leak per session which is bounded for a maximum of keepaliveTimeoutMs. On the other hand even with that potential for a short reference hold, this implementation proposed here is more correct I think. One alternative we could do is keep a list of pending timeouts.. which is complex for a rare situation that will self resolve anyhow when keepaliveTimeoutMs is reached. So I'm going with the cheaper, more common case.
- Bug Fix: keepalive intervals were being cleared with an incorrect clearTimeout before. Not sure if this was causing intervals leaks in some nodejs impls or not. (v20.13.1 seems to accept this mismatch without issue)
- Rename variables for clarity, to prevent future bugs like swapping clearInterval vs clearTimeout.
- Trace on GOAWAY frames and ensure we drop the connection
- Implementation is repeated in two places, per warning from https://github.com/grpc/grpc-node/pull/2756#issuecomment-2136031256
- This commit supercedes the prior PR on a master branch which was out of date. https://github.com/grpc/grpc-node/pull/2756

Aims to help resolve https://github.com/grpc/grpc-node/issues/2734

### DIFF
--- a/packages/grpc-js/src/server.ts
+++ b/packages/grpc-js/src/server.ts
@@ -18,50 +18,10 @@
 import * as http2 from 'http2';
 import * as util from 'util';
 
+import { CipherNameAndProtocol, TLSSocket } from 'tls';
 import { ServiceError } from './call';
-import { Status, LogVerbosity } from './constants';
-import { Deserialize, Serialize, ServiceDefinition } from './make-client';
-import { Metadata } from './metadata';
-import {
-  BidiStreamingHandler,
-  ClientStreamingHandler,
-  HandleCall,
-  Handler,
-  HandlerType,
-  sendUnaryData,
-  ServerDuplexStream,
-  ServerDuplexStreamImpl,
-  ServerReadableStream,
-  ServerStreamingHandler,
-  ServerUnaryCall,
-  ServerWritableStream,
-  ServerWritableStreamImpl,
-  UnaryHandler,
-  ServerErrorResponse,
-  ServerStatusResponse,
-  serverErrorToStatus,
-} from './server-call';
-import { ServerCredentials } from './server-credentials';
+import { PartialStatusObject } from './call-interface';
 import { ChannelOptions } from './channel-options';
-import {
-  createResolver,
-  ResolverListener,
-  mapUriDefaultScheme,
-} from './resolver';
-import * as logging from './logging';
-import {
-  SubchannelAddress,
-  isTcpSubchannelAddress,
-  subchannelAddressToString,
-  stringToSubchannelAddress,
-} from './subchannel-address';
-import {
-  GrpcUri,
-  combineHostPort,
-  parseUri,
-  splitHostPort,
-  uriToString,
-} from './uri-parser';
 import {
   ChannelzCallTracker,
   ChannelzCallTrackerStub,
@@ -69,23 +29,63 @@ import {
   ChannelzChildrenTrackerStub,
   ChannelzTrace,
   ChannelzTraceStub,
-  registerChannelzServer,
-  registerChannelzSocket,
   ServerInfo,
   ServerRef,
   SocketInfo,
   SocketRef,
   TlsInfo,
+  registerChannelzServer,
+  registerChannelzSocket,
   unregisterChannelzRef,
 } from './channelz';
-import { CipherNameAndProtocol, TLSSocket } from 'tls';
+import { LogVerbosity, Status } from './constants';
+import * as logging from './logging';
+import { Deserialize, Serialize, ServiceDefinition } from './make-client';
+import { Metadata } from './metadata';
+import {
+  ResolverListener,
+  createResolver,
+  mapUriDefaultScheme,
+} from './resolver';
+import {
+  BidiStreamingHandler,
+  ClientStreamingHandler,
+  HandleCall,
+  Handler,
+  HandlerType,
+  ServerDuplexStream,
+  ServerDuplexStreamImpl,
+  ServerErrorResponse,
+  ServerReadableStream,
+  ServerStatusResponse,
+  ServerStreamingHandler,
+  ServerUnaryCall,
+  ServerWritableStream,
+  ServerWritableStreamImpl,
+  UnaryHandler,
+  sendUnaryData,
+  serverErrorToStatus,
+} from './server-call';
+import { ServerCredentials } from './server-credentials';
 import {
   ServerInterceptingCallInterface,
   ServerInterceptor,
   getServerInterceptingCall,
 } from './server-interceptors';
-import { PartialStatusObject } from './call-interface';
+import {
+  SubchannelAddress,
+  isTcpSubchannelAddress,
+  stringToSubchannelAddress,
+  subchannelAddressToString,
+} from './subchannel-address';
 import { CallEventTracker } from './transport';
+import {
+  GrpcUri,
+  combineHostPort,
+  parseUri,
+  splitHostPort,
+  uriToString,
+} from './uri-parser';
 
 const UNLIMITED_CONNECTION_AGE_MS = ~(1 << 31);
 const KEEPALIVE_MAX_TIME_MS = ~(1 << 31);
@@ -1469,6 +1469,12 @@ export class Server {
             clearKeepaliveTimeout();
             if (err) {
               this.keepaliveTrace('Ping failed with error: ' + err.message);
+              this.trace(
+                'Connection dropped due to error of a ping frame ' +
+                  err.message +
+                  ' return in ' +
+                  duration
+              );
               sessionClosedByServer = true;
               session.close();
             } else {
@@ -1480,6 +1486,7 @@ export class Server {
 
         if (!pingSentSuccessfully) {
           this.keepaliveTrace('Ping failed to send');
+          this.trace('Connection dropped due to failure to send ping frame');
           sessionClosedByServer = true;
           session.close();
           return;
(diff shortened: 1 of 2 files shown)


packages/grpc-js/src/server.ts
```diff
@@ -1421,41 +1420,72 @@ export class Server {
...
             session.destroy();
           }
         }, this.keepaliveTimeMs);
-        keeapliveTimeTimer.unref?.();
+        keepaliveInterval.unref?.();
       }
 
+      session.once('goaway', (errorCode, lastStreamID, opaqueData) => {
+        if (errorCode === http2.constants.NGHTTP2_ENHANCE_YOUR_CALM) {
+          this.trace('Connection dropped by client due to ENHANCE_YOUR_CALM');
```
murgatroid99, line 1478: Does the client ever actually send a GOAWAY with this error code? There's a reason for the server to send it: the client can take that information into consideration after reconnecting, and send pings less frequently (and we actually do that). But the server doesn't keep track of which clients have sent this, so it doesn't change behavior.

Also, ENHANCE_YOUR_CALM only means that the connection closed due to excessive pings if the `opaqueData` is equal to the string `too_many_pings`.

In addition, the log lines in both branches are essentially the same. I'd rather be consistent, and either always report the error code, or always report the string name of the error code, and avoid the branch entirely.

packages/grpc-js/src/server.ts
```diff
@@ -1421,41 +1420,72 @@ export class Server {
...
+        } else {
+          this.trace(
+            'Connection dropped by client via GOAWAY with error code ' +
+              errorCode
+          );
+        }
+        sessionClosedByServer = true;
```
murgatroid99, line 1485: The log lines indicate that this event means that the session was closed by the client, but this line says that the session was closed by the server. It shouldn't be both.

packages/grpc-js/src/server.ts
```diff
@@ -1421,41 +1420,72 @@ export class Server {
...
-          keepaliveTimeoutTimer.unref?.();
+          keepaliveTimeout.unref?.();
 
           try {
-            session.ping(
-              (err: Error | null, duration: number, payload: Buffer) => {
-                if (keepaliveTimeoutTimer) {
-                  clearTimeout(keepaliveTimeoutTimer);
+            if (
+              !session.ping(
```
murgatroid99, line 1437: I don't like having a big statement in an `if` condition. I'd rather have a `pingSentSuccessfully` variable or something.

packages/grpc-js/src/server.ts
```diff
@@ -1421,41 +1420,72 @@ export class Server {
...
+            }
           } catch (e) {
-            clearTimeout(keepaliveTimeoutTimer);
-            // The ping can't be sent because the session is already closed
+            // The ping can't be sent because the session is already closed, max outstanding pings reached, etc
+            clearTimeout(keepaliveTimeout);
+            if (keepaliveInterval) {
+              clearInterval(keepaliveInterval);
+              keepaliveInterval = null;
+            }
```
murgatroid99, line 1465: We shouldn't need to do anything with these timeouts here. `session.destroy` should trigger the `close` event, which will handle them.

packages/grpc-js/src/server.ts
```diff
@@ -1421,41 +1420,72 @@ export class Server {
...
-                      ' return in ' +
-                      duration
-                  );
-                  session.close();
-                }
-              }
-            );
+              )
+            ) {
+              throw new Error('Server keepalive ping send failed');
```
murgatroid99, line 1457: I don't like the `throw` immediately into a `catch` block here. If you remove the cleanup stuff in the `catch` block, I think it's reasonable to just have separate `trace` and `session.destroy` calls here.

packages/grpc-js/src/server.ts
```diff
@@ -1421,41 +1420,72 @@ export class Server {
       }
 
       if (this.keepaliveTimeMs < KEEPALIVE_MAX_TIME_MS) {
-        keeapliveTimeTimer = setInterval(() => {
-          keepaliveTimeoutTimer = setTimeout(() => {
-            sessionClosedByServer = true;
-            session.close();
+        keepaliveInterval = setInterval(() => {
+          const keepaliveTimeout = setTimeout(() => {
```
murgatroid99, line 1424: Instead of making these changes to support multiple outstanding pings and corresponding keepalive timeout timers, I think it would be better to take the same approach as on the client, and ensure that there is only one outstanding ping at any given time.

There isn't really a reason to have multiple outstanding pings: in terms of keeping the connection alive, if you need to send a second ping before receiving a previous response, that implies that your round trip latency can be longer than the maximum connection age, which probably means that you have other problems. And in terms of detecting that a connection is no longer working, the timeout for the first ping will end before the timeout for the second ping, so the second ping doesn't let you find out faster.

packages/grpc-js/src/server.ts
```diff
@@ -1421,41 +1420,72 @@ export class Server {
...
+            if (keepaliveInterval) {
+              clearInterval(keepaliveInterval);
+              keepaliveInterval = null;
```
murgatroid99, line 1427: These cleanup actions should be handled in the `close` event handler triggered by the `session.close` call below. We don't also need to do them here.

packages/grpc-js/src/server.ts
```diff
@@ -1421,41 +1420,72 @@ export class Server {
...
-                if (keepaliveTimeoutTimer) {
-                  clearTimeout(keepaliveTimeoutTimer);
+            if (
+              !session.ping(
+                (err: Error | null, duration: number, payload: Buffer) => {
+                  clearTimeout(keepaliveTimeout);
+                  if (err) {
+                    if (keepaliveInterval) {
+                      clearInterval(keepaliveInterval);
+                      keepaliveInterval = null;
```
murgatroid99, line 1443: Again, these cleanup actions shouldn't be needed here.

murgatroid99 (changes requested):
All per-line comments apply to both implementations.

`channelzTrace.addTrace` isn't a replacement for `trace`. `channelzTrace.addTrace` only traces to channelz, so that will only be seen by a channelz client. You still want to call `trace` to log to the console or whatever log file.

Also, keepalive logs may be verbose compared to other server logs, so I think it would be a good idea to have a separate tracer, as the client does. It's probably fine to use the same `keepalive` tracer name that the client uses.

packages/grpc-js/src/server.ts
murgatroid99: #2139 reported that `session.ping` can throw errors.

packages/grpc-js/src/transport.ts
```diff
@@ -101,28 +101,29 @@ class Http2Transport implements Transport {
...
-   * Timer reference tracking when the most recent ping will be considered lost
-   */
-  private keepaliveTimeoutId: NodeJS.Timeout | null = null;
-  /**
-   * Indicates whether keepalive pings should be sent without any active calls
+   * Indicates when keepalives should no longer be performed for this transport. Used to prevent a race where a
+   * latent session.ping(..) callback is called after the transport has been notified to disconnect.
    */
-  private keepaliveWithoutCalls = false;
+  private keepaliveDisabled = false;
```
murgatroid99, line 126: I don't think we need a separate variable for this. We should just be able to check `session.destroyed`.

packages/grpc-js/src/transport.ts
```diff
@@ -101,28 +101,29 @@ class Http2Transport implements Transport {
...
   /**
-   * Timer reference for timeout that indicates when to send the next ping
+   * Indicates whether keepalive pings should be sent without any active calls
+   */
+  private readonly keepaliveWithoutCalls: boolean;
+  /**
+   * Timer reference indicating when to send the next ping or when the most recent ping will be considered lost.
    */
-  private keepaliveTimerId: NodeJS.Timeout | null = null;
+  private keepaliveTimeout: NodeJS.Timeout | null = null;
```
murgatroid99, line 116: This shouldn't be called "keepaliveTimeout" because that is already the name of another concept in this context. `keepaliveTimer` should be fine.

murgatroid99 (changes requested):
I think the channelz traces are fine the way they are.

packages/grpc-js/src/transport.ts
murgatroid99: The main consequence of that sequence of events is that one extra timer is started. And I don't think that's a significant enough problem that we need to track more state just to avoid it.

We could minimize the chances of that by explicitly calling `session.destroy()` at the end of the `setImmediate` block in `handleDisconnect`.

murgatroid99:
Sorry about the delay, I had to focus on https://github.com/grpc/grpc-node/security/advisories/GHSA-7v5v-9h63-cj86. I can get this out soon.

murgatroid99:

PR #2761, opened 2024-05-29
fix: add decoding for url encoded user credentials

brendan-myers:
Added support for URL encoded user credentials.

Previously, if a username/password contained a special character that required encoding, eg `grpc_proxy=http://user%40example.com:password@127.0.0.1:8888`, then the connection would be rejected (40x).


murgatroid99:

PR #2772, opened 2024-06-06
grpc-js: Fix client hang when receiving extra messages for a unary response

### LINKED ISSUE #2768: Cardinality violations should use error code “unimplemented”
The gRPC docs for error codes state that both client and server should use the `unimplemented` code for cardinality violations. See table at the bottom of this doc (you can search for “cardinality violation” in the doc): https://grpc.github.io/grpc/core/md_doc_statuscodes.html.

A cardinality violation is when a stream contains an incorrect number of messages. Specifically, when a response stream for a unary or client-stream RPC contains zero messages with an OK status or more than one message; or when a request stream for a unary or server-stream RPC contains zero or more than one messages.

The server in this repo conforms to the documented behavior, but the client does not.

1. For unary and client-stream RPCs, if no response messages are present in the response, only an “OK” status, the client reports an `internal` error (with a message of “No message received”) instead of  `unimplemented`.
2. For unary and client-stream RPCs, if a second response message is erroneously sent by the server before the trailers/status, the operation appears to hang indefinitely (!!). A `metadata` event is triggered, but nothing else — no callback, no `error` event, no `status` event, and no `end` event.

murgatroid99:
The new `call.startRead()` call is the fix for the main bug. Previously the  I also changed the client to use the `UNIMPLEMENTED` status code for cardinality violations to make it consistent with the server, and because that is what the spec actually says.

This fixes #2768.

### DIFF
--- a/packages/grpc-js/src/client.ts
+++ b/packages/grpc-js/src/client.ts
@@ -330,7 +330,7 @@ export class Client {
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       onReceiveMessage(message: any) {
         if (responseMessage !== null) {
-          call.cancelWithStatus(Status.INTERNAL, 'Too many responses received');
+          call.cancelWithStatus(Status.UNIMPLEMENTED, 'Too many responses received');
         }
         responseMessage = message;
       },
@@ -345,7 +345,7 @@ export class Client {
             callProperties.callback!(
               callErrorFromStatus(
                 {
-                  code: Status.INTERNAL,
+                  code: Status.UNIMPLEMENTED,
                   details: 'No message received',
                   metadata: status.metadata,
                 },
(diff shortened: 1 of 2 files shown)


gnossen (approved):

ejona86:
We should not use UNIMPLEMENTED. It is very bad if it ever gets triggered. Please swap back to INTERNAL. We will change the spec.

murgatroid99:
Is it really that bad? I would say that if a server responds with multiple messages to a method that the client believes is unary or client streaming, the server doesn't implement the method the client is intends to access.

In any case, as long as this is what the spec says, I consider it to be correct behavior. If and when the spec changes, I'll make the change here too. But as far as I know Python and C++ also have this behavior, so it is consistent with (some) other languages.

Issue #2775, opened 2024-06-18
Implement authority overrides in the DNS resolver


atollena:
### Is your feature request related to a problem? Please describe.

We have use cases where we would like to use the DNS resolver with a DNS server that is not the system's DNS server.

### Describe the solution you'd like

Implement DNS server authority parsing as described in https://github.com/grpc/grpc/blob/master/doc/naming.md:

> `authority` indicates the DNS server to use, although this is only supported by some implementations. (In C-core, the default DNS resolver does not support this, but the c-ares based resolver supports specifying this in the form "IP:port".).

- [Golang implementation](https://github.com/grpc/grpc-go/blob/c04b085930ce33ee83cc3f92dbe7632031e127a9/internal/resolver/dns/dns_resolver.go#L91-L107)
- [C-core implementation (only when using `GRPC_DNS_RESOLVER=ares`)](https://github.com/grpc/grpc/blob/master/src/core/resolver/dns/c_ares/dns_resolver_ares.cc)
- [Java: not implemented as customizing DNS server is a bit of a moving target](https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/DnsNameResolver.java#L165-L166)

### Describe alternatives you've considered

Do more customisation to the DNS configuration, not involving gRPC directly.

### Additional context

We have workloads in javascript that run browser tests and use public DNS for most resolutions, and a custom DNS server for our internal service discovery system to report results to our internal systems.

atollena:
We are likely willing to provide a patch for this, if that helps.

murgatroid99:
I don't see how you would do that. The only way I can see to make requests to a different DNS server is with `dns.setServers`, but that is a global setting, as opposed to affecting a single request, so it is inappropriate to use in a library.

gkampitakis:
Something like this where you create a new resolver https://github.com/gkampitakis/grpc-node/pull/1?

murgatroid99:
I see. I didn't realize that the Resolver class had its own `setServers` method. That should do it.

gkampitakis:
I have created the pr with the above changes 👍

atollena:
Fixed in https://github.com/grpc/grpc-node/pull/2776.

PR #2776, opened 2024-06-18
fix: support authority overrides in the DNS resolver

gkampitakis:
As per issue https://github.com/grpc/grpc-node/issues/2775, this pr adds support for overriding the authority on DNS resolver.

### DIFF
--- a/packages/grpc-js/gulpfile.ts
+++ b/packages/grpc-js/gulpfile.ts
@@ -91,11 +91,17 @@ const testWithAlternativeResolver = checkTask(() => {
   );
 });
 
+const resetEnv = () => {
+  process.env.GRPC_NODE_USE_ALTERNATIVE_RESOLVER = 'false';
+  return Promise.resolve();
+};
+
 const test = gulp.series(
   install,
   copyTestFixtures,
   runTests,
-  testWithAlternativeResolver
+  testWithAlternativeResolver,
+  resetEnv
 );
 
 export { install, lint, clean, cleanAll, compile, test };
--- a/packages/grpc-js/gulpfile.ts
+++ b/packages/grpc-js/gulpfile.ts
@@ -81,6 +81,21 @@ const runTests = checkTask(() => {
   );
 });
 
-const test = gulp.series(install, copyTestFixtures, runTests);
+const testWithAlternativeResolver = checkTask(() => {
+  process.env.GRPC_NODE_USE_ALTERNATIVE_RESOLVER = 'true';
+  return gulp.src(`${outDir}/test/**/test-resolver.js`).pipe(
+    mocha({
+      reporter: 'mocha-jenkins-reporter',
+      require: ['ts-node/register'],
+    })
+  );
+});
+
+const test = gulp.series(
+  install,
+  copyTestFixtures,
+  runTests,
+  testWithAlternativeResolver
+);
 
 export { install, lint, clean, cleanAll, compile, test };
--- a/packages/grpc-js/gulpfile.ts
+++ b/packages/grpc-js/gulpfile.ts
@@ -83,7 +83,7 @@ const runTests = checkTask(() => {
 
 const testWithAlternativeResolver = checkTask(() => {
   process.env.GRPC_NODE_USE_ALTERNATIVE_RESOLVER = 'true';
-  return gulp.src(`${outDir}/test/**/test-resolver.js`).pipe(
+  return gulp.src(`${outDir}/test/test-resolver.js`).pipe(
     mocha({
       reporter: 'mocha-jenkins-reporter',
       require: ['ts-node/register'],
--- a/packages/grpc-js/src/resolver-dns.ts
+++ b/packages/grpc-js/src/resolver-dns.ts
@@ -21,6 +21,7 @@ import {
   registerDefaultScheme,
 } from './resolver';
 import * as dns from 'dns';
+import * as dnsPromises from 'dns/promises';
 import * as util from 'util';
 import { extractAndSelectServiceConfig, ServiceConfig } from './service-config';
 import { Status } from './constants';
@@ -48,7 +49,6 @@ export const DEFAULT_PORT = 443;
 const DEFAULT_MIN_TIME_BETWEEN_RESOLUTIONS_MS = 30_000;
 
 const resolveTxtPromise = util.promisify(dns.resolveTxt);
-const dnsLookupPromise = util.promisify(dns.lookup);
 
 /**
  * Resolver implementation that handles DNS names and IP addresses.
@@ -63,7 +63,7 @@ class DnsResolver implements Resolver {
    * Failures are handled by the backoff timer.
    */
   private readonly minTimeBetweenResolutionsMs: number;
-  private pendingLookupPromise: Promise<dns.LookupAddress[]> | null = null;
+  private pendingLookupPromise: Promise<string[]> | null = null;
   private pendingTxtPromise: Promise<string[][]> | null = null;
   private latestLookupResult: Endpoint[] | null = null;
   private latestServiceConfig: ServiceConfig | null = null;
@@ -76,12 +76,17 @@ class DnsResolver implements Resolver {
   private isNextResolutionTimerRunning = false;
   private isServiceConfigEnabled = true;
   private returnedIpResult = false;
+  private resolver = new dnsPromises.Resolver();
+
   constructor(
     private target: GrpcUri,
     private listener: ResolverListener,
     channelOptions: ChannelOptions
   ) {
     trace('Resolver constructed for target ' + uriToString(target));
+    if (target.authority) {
+      this.resolver.setServers([target.authority]);
+    }
     const hostPort = splitHostPort(target.path);
     if (hostPort === null) {
       this.ipResult = null;
@@ -189,7 +194,7 @@ class DnsResolver implements Resolver {
        * because when looking up a single family, dns.lookup outputs an error
        * if the name exists but there are no records for that family, and that
        * error is indistinguishable from other kinds of errors */
-      this.pendingLookupPromise = dnsLookupPromise(hostname, { all: true });
+      this.pendingLookupPromise = this.resolver.resolve(hostname);
       this.pendingLookupPromise.then(
         addressList => {
           if (this.pendingLookupPromise === null) {
@@ -199,7 +204,7 @@ class DnsResolver implements Resolver {
           this.backoff.reset();
           this.backoff.stop();
           const subchannelAddresses: TcpSubchannelAddress[] = addressList.map(
-            addr => ({ host: addr.address, port: +this.port! })
+            addr => ({ host: addr, port: +this.port! })
           );
           this.latestLookupResult = subchannelAddresses.map(address => ({
             addresses: [address],
--- a/packages/grpc-js/src/resolver-dns.ts
+++ b/packages/grpc-js/src/resolver-dns.ts
@@ -20,9 +20,8 @@ import {
   registerResolver,
   registerDefaultScheme,
 } from './resolver';
-import * as dns from 'dns';
-import * as dnsPromises from 'dns/promises';
-import * as util from 'util';
+import { AnyRecord } from 'dns';
+import * as dns from 'dns/promises';
 import { extractAndSelectServiceConfig, ServiceConfig } from './service-config';
 import { Status } from './constants';
 import { StatusObject } from './call-interface';
@@ -48,8 +47,6 @@ export const DEFAULT_PORT = 443;
 
 const DEFAULT_MIN_TIME_BETWEEN_RESOLUTIONS_MS = 30_000;
 
-const resolveTxtPromise = util.promisify(dns.resolveTxt);
-
 /**
  * Resolver implementation that handles DNS names and IP addresses.
  */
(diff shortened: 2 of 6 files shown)


packages/grpc-js/src/resolver-dns.ts
```diff
@@ -189,7 +194,7 @@ class DnsResolver implements Resolver {
        * because when looking up a single family, dns.lookup outputs an error
        * if the name exists but there are no records for that family, and that
        * error is indistinguishable from other kinds of errors */
-      this.pendingLookupPromise = dnsLookupPromise(hostname, { all: true });
+      this.pendingLookupPromise = this.resolver.resolve(hostname);
```
murgatroid99, line 197: This should match the previous behavior, so it should be `this.resolver.lookup(hostname, { all: true })`. The other changes that this causes should also be reversed.

packages/grpc-js/src/resolver-dns.ts
```diff
@@ -48,7 +49,6 @@ export const DEFAULT_PORT = 443;
 const DEFAULT_MIN_TIME_BETWEEN_RESOLUTIONS_MS = 30_000;
 
 const resolveTxtPromise = util.promisify(dns.resolveTxt);
```
murgatroid99, line 51: This also needs to be changed, so that all resolution operations use the custom resolver.

gkampitakis: ah yes correct, missed that.

packages/grpc-js/src/resolver-dns.ts
gkampitakis: There is not `.lookup` for the Resolver class https://nodejs.org/api/dns.html#class-dnsresolver. I thought it resolves both ipv4 and ipv6 but checking the docs again is defaults to ipv4. We could use `resolveAny` and filter only `A` and `AAA`, wdyth?

murgatroid99: You're right, I thought I saw it there. Unfortunately, according to the `resolveAny` documentation,

> DNS server operators may choose not to respond to ANY queries. It may be better to call individual methods like [dns.resolve4()](https://nodejs.org/api/dns.html#dnsresolve4hostname-options-callback), [dns.resolveMx()](https://nodejs.org/api/dns.html#dnsresolvemxhostname-callback), and so on.

So we can't use `resolveAny`. As a result, `lookup` is the only function that will consistently get IPv4 and IPv6 results in a single operation.

In addition, as stated in the [Implementation considerations section](https://nodejs.org/api/dns.html#implementation-considerations), `lookup` uses `getaddrinfo` while `resolve` operations directly make DNS queries, probably using c-ares as the C core does. This means that switching between them can introduce other unexpected changes in behavior.

gkampitakis: hm I see. What would your suggestion then be for this case?

murgatroid99: I think this is part of the reason the C core has two different DNS configurations, and this behavior is only supported in one of them. One possibility is to mirror that here, with an equivalent to `GRPC_DNS_RESOLVER=ares`.

gkampitakis: Okay I see I can update the pr with something similar and see how the solution looks like? 

Thinking something (very basic pseudo), so change the behaviour if explicitly set like it. 

```js
if (process.env.GRPC_DNS_TODO_NAME) {
	this.resolver  = dns.ResolverWithDNSServer()
} else {
   this.resolver = dns.lookup()
}
```

gkampitakis: Have updated the pr where now we can select the behaviour. Also have left the env name as a placeholder.

gkampitakis: An update on my side with `Any` I don't get any response from our DNS also reading the docs
> DNS server operators may choose not to respond to ANY queries. It may be better to call individual methods like [dns.resolve4()](https://nodejs.org/api/dns.html#dnsresolve4hostname-options-callback), [dns.resolveMx()](https://nodejs.org/api/dns.html#dnsresolvemxhostname-callback), and so on. For more details, see [RFC 8482](https://tools.ietf.org/html/rfc8482).

Would be acceptable calling both methods (`resolve4`, `resolve6`)? Not sure what are the implications there though.

gkampitakis: @murgatroid99 any thoughts on this? To do something around 

```
merge(await promise(resolve4(), resolve6))
```

murgatroid99: I'm not sure what you mean by `promise(resolve4(), resolve6()))`. Calling both `resolve4` and `resolve6` should be OK, but I am concerned about one of them returning a result and the other one erroring. So, this should probably use `Promise.allSettled()`.

gkampitakis: Yes this is what I meant. Cool will work on updating the pr to use this logic and also add some unit tests. 

Also do I understand correct that the new behaviour is going to be used using `GRPC_DNS_RESOLVER=ares` ?

gkampitakis:
I have updated per recommendation the `lookup` method to resolve explicitly both ipv4 and ipv6. I see there are some unit tests but are skipped, what would be your suggestion for that matter? @murgatroid99

packages/grpc-js/src/resolver-dns.ts
```diff
@@ -45,10 +44,8 @@ function trace(text: string): void {
  */
 export const DEFAULT_PORT = 443;
 
-const DEFAULT_MIN_TIME_BETWEEN_RESOLUTIONS_MS = 30_000;
-
-const resolveTxtPromise = util.promisify(dns.resolveTxt);
-const dnsLookupPromise = util.promisify(dns.lookup);
+const DEFAULT_MIN_TIME_BETWEEN_RESOLUTIONS_MS = 30_000,
+  DNS_RESOLUTION_ENV = 'GRPC_DNS_RESOLVER_TODO';
```
murgatroid99, line 48: Please handle the environment variable in `environment.ts`, following the examples in that file.

packages/grpc-js/src/resolver-dns.ts
murgatroid99: I'm not sure about that. Since that setting already exists in the other library, people may have expectations about its behavior, and it's unclear if this implementation will meet that expectations. Maybe it could be something like `GRPC_NODE_USE_ALTERNATIVE_RESOLVER=true|false`

gkampitakis: Makes sense 👍

gkampitakis:
@murgatroid99 code has been update per your feedback, thank you. I have also tested locally and it does behave as expected. 

Are there any 
- tests that need to be added/updated 
- docs added/updated

MartinLoeper:
This feature is fantastic. Thanks so much for implementing it @gkampitakis! :partying_face:

gkampitakis:
@murgatroid99 I don't see an example of how to test this code path (using an env variable). 

To make this testable with no weird monkey patching the import I would have to rework the code and pass this env variable as a parameter but again I don't see any good pattern in the code base? Any suggestions for tackling the testing here?

murgatroid99:
OK, for testing I was thinking that a `gulpfile` target could be added that runs just that one test file with the environment variable set. Then the actual `test` target would run both tests in sequence.

gkampitakis:
> OK, for testing I was thinking that a gulpfile target could be added that runs just that one test file with the environment variable set. Then the actual test target would run both tests in sequence.

That makes sense. I am not very familiar with this setup. Thanks for the advice. I have added the new test but on my machine the timeouts are quite flaky, but didn't want to "touch" them without understanding what's happening there.

packages/grpc-js/test/test-resolver.ts
```diff
@@ -460,7 +460,7 @@ describe('Name Resolver', () => {
             }
           },
         },
-        {}
+        { 'grpc.dns_min_time_between_resolutions_ms': 2000 }
```
murgatroid99, line 463: This shouldn't be necessary. This test expects DNS resolution to fail, and this option only takes effect if it succeeds.

packages/grpc-js/gulpfile.ts
```diff
@@ -81,6 +81,21 @@ const runTests = checkTask(() => {
   );
 });
 
-const test = gulp.series(install, copyTestFixtures, runTests);
+const testWithAlternativeResolver = checkTask(() => {
+  process.env.GRPC_NODE_USE_ALTERNATIVE_RESOLVER = 'true';
+  return gulp.src(`${outDir}/test/**/test-resolver.js`).pipe(
```
murgatroid99, line 86: This string should just be `${outDir}/test/test-resolver.js`

murgatroid99 (commented):
The tests that try to resolve `localhost` need to be skipped when using the alternate resolver. You can do this by changing the test declaration to use `function(done) {...}` and adding this to the test body:

```ts
if (GRPC_NODE_USE_ALTERNATIVE_RESOLVER) {
  this.skip();
}
```

packages/grpc-js/test/test-resolver.ts
gkampitakis: For this test I am getting error without this value. I assume it allow us to "retry" resolution faster but probably it's another issue.

```
Uncaught AssertionError [ERR_ASSERTION]: resultCount 1 !== 2
      + expected - actual

      -1
      +2
      
      at Timeout._onTimeout (test/test-resolver.ts:477:16)
      at listOnTimeout (node:internal/timers:569:17)
      at processTimers (node:internal/timers:512:7)
```

Also getting timeouts on, should I increase the value? 

```
  Should resolve gRPC interop servers:
     Error: Timeout of 4000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/george.kampitakis/Dev/grpc-node/packages/grpc-js/build/test/test-resolver.js)
      at listOnTimeout (node:internal/timers:569:17)
      at processTimers (node:internal/timers:512:7)
```

packages/grpc-js/src/resolver-dns.ts
```diff
@@ -302,6 +295,41 @@ class DnsResolver implements Resolver {
     }
   }
 
+  private async lookup(hostname: string): Promise<TcpSubchannelAddress[]> {
+    if (GRPC_NODE_USE_ALTERNATIVE_RESOLVER) {
+      const records = await Promise.allSettled([
```
murgatroid99, line 300: If both of these are rejected, the whole `lookup` function should reject. This should fix the problem with the test that expects lookup failures.

gkampitakis: Ahhh yes. Totally missed that. Good there is the test there 🙇

gkampitakis:
Can't tell if the failing tests are related to my changes 🤔

packages/grpc-js/gulpfile.ts
```diff
@@ -81,6 +81,21 @@ const runTests = checkTask(() => {
   );
 });
 
-const test = gulp.series(install, copyTestFixtures, runTests);
+const testWithAlternativeResolver = checkTask(() => {
+  process.env.GRPC_NODE_USE_ALTERNATIVE_RESOLVER = 'true';
```
murgatroid99, line 85: Unfortunately, I think this is impacting other test jobs. I think this alternative will work: create a new file called something like `alternate-resolver.env` that just says `GRPC_NODE_USE_ALTERNATIVE_RESOLVER=true` and add an option in the `mocha` command a couple of lines down that says `nodeOption: 'env-file=alternate-resolver.env'`.

gkampitakis: If you are referring to this, I don't see a `nodeOption` or something similar from the documentation. 

```
mocha({
    reporter: 'mocha-jenkins-reporter',
    require: ['ts-node/register'],
  })
```

murgatroid99: That is what I'm talking about. The gulp-mocha documentation says

> Options are passed directly to the mocha binary, so you can use any its [command-line options](http://mochajs.org/#usage) in a camelCased form.

The mocha command line options documentation includes this line:

```
 -n, --node-option  Node or V8 option (no leading "--")                 [array]
```

gkampitakis: the node `--env-file` was added only on node 20 though right? will it work with the CI running on 14 and 16?

murgatroid99: You're right, that probably won't work. So, here's another alternative: add another target that just sets `process.env.GRPC_NODE_USE_ALTERNATIVE_RESOLVER = 'false';` and add it to the end of the `test` series.

murgatroid99: Unfortunately, I don't think this quite works because at the top level, the test suites are run in parallel. I think changing [this line](https://github.com/grpc/grpc-node/blob/master/gulpfile.ts#L44) to run in series should fix it.

murgatroid99 (approved):
OK, this seems to be in a good state now. Thank you for contributing this, and for working with me to get the tests working.

gkampitakis:
@murgatroid99 Thank you for the patience and all the help to bring it in a good state 🙇

murgatroid99:
This is now out in version 1.11.0.

PR #2801, opened 2024-07-18
fix(grpc-js): client crash on custom error code

hastom:
In version 1.11.1, grpc client crashes if receives custom error code with error `cannot read properties of undefined reading .toLowerCase()`

### DIFF
--- a/packages/grpc-js/test/test-retry.ts
+++ b/packages/grpc-js/test/test-retry.ts
@@ -333,6 +333,7 @@ describe('Retries', () => {
         metadata,
         (error: grpc.ServiceError, response: any) => {
           assert(error);
+          assert.strictEqual(error.code, 300);
           assert.strictEqual(error.details, 'Failed on retry 0');
           done();
         }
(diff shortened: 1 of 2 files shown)


packages/grpc-js/test/test-retry.ts
```diff
@@ -323,6 +323,21 @@ describe('Retries', () => {
...
+    it('Should not retry on custom error code', done => {
+      const metadata = new grpc.Metadata();
+      metadata.set('succeed-on-retry-attempt', '2');
+      metadata.set('respond-with-status', '300');
+      client.echo(
+        { value: 'test value', value2: 3 },
+        metadata,
+        (error: grpc.ServiceError, response: any) => {
+          assert(error);
+          assert.strictEqual(error.details, 'Failed on retry 0');
```
murgatroid99, line 336: This test should also assert that the error code that the application sees is equal to 300. Otherwise, this test could pass if the server doesn't send the custom error code at all, avoiding the relevant code path entirely.

hastom: ty, fixed

murgatroid99:
This is out in version 1.11.2

PR #2806, opened 2024-07-31
grpc-js: Simplify pick_first behavior

murgatroid99:
This includes a few related changes to the pick_first LB policy to simplify its behavior, in order to reduce the chances of reaching a bad state, and to make it easier to reason about:

 - In `pickSubchannel`, avoid assuming that the subchannel to be picked has any state related to this LB policy instance; in particular, don't assume that it has the state change listener registered, and don't have special case behavior if it's the same subchannel as the current pick. Instead, unconditionally reset the LB policy state and set up the relevant subchannel state from scratch.
 - In `resetSubchannelList`, remove the special case when handling the current pick.
 - In `connectToAddressList`, don't take any additional actions when immediately picking one of the new subchannels.
 - In `removeCurrentPick`, call `unref` last to completely avoid calling the connectivity state listener.
 - In `updateAddressList`, switch `requestedResolutionSinceLastUpdate` to `false` immediately instead of assuming that will be done elsewhere.

### DIFF
--- a/packages/grpc-js/src/load-balancer-pick-first.ts
+++ b/packages/grpc-js/src/load-balancer-pick-first.ts
@@ -320,21 +320,16 @@ export class PickFirstLoadBalancer implements LoadBalancer {
 
   private removeCurrentPick() {
     if (this.currentPick !== null) {
-      /* Unref can cause a state change, which can cause a change in the value
-       * of this.currentPick, so we hold a local reference to make sure that
-       * does not impact this function. */
-      const currentPick = this.currentPick;
-      this.currentPick = null;
-      currentPick.unref();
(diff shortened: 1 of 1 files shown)


murgatroid99:
The last commit removes `triedAllSubchannels` and `requestedResolutionSinceLastUpdate`.

`triedAllSubchannels` just conditionally skips `startNextSubchannelConnecting`, but calling that an extra time shouldn't be a problem. It might just cause some extra connection attempts.

`requestedResolutionSinceLastUpdate` was used to request resolution exactly once each time every subchannel fails a connection attempt. Now instead the `hasReportedTransientFailure` is just reset each time it becomes true for every subchannel.

gnossen (approved):

Issue #2811, opened 2024-08-22
Node.JS @grpc/grpc-js issue with disabling the verification of the server's TLS certificate


vinothsa4891:
### Problem description
We have a use case to access the gRPC service via Tunneling through a reverse proxy. The Proxy has a self-signed TLS certificate.

I see an option to disable the verification of the server's TLS certificate in client credentials. Check ServerIdentity: ()=> undefined. This is not working as I'm getting an error "Error: 14 UNAVAILABLE: No connection established. Last error: self-signed certificate in certificate chain"
   
 const verifyOptionsFalse = {
      checkServerIdentity: () => undefined,
    };
 
    const sslCreds = grpc.credentials.createSsl(
      null,
      null,
      null,
      verifyOptionsFalse 
    );

this._grpcObj.<< serviceName >>(
      "<< proxy_dns >>:8081",
      sslCreds,
    )

This is working only if I pass the server ssl_certificate ( ex : servercert.crt ) or set up process.env.NODE_TLS_REJECT_UNAUTHORIZED='0'.

Also in Golang its working by just setting "InsecureSkipVerify:false" -  
creds := credentials.NewTLS(&tls.Config{
        InsecureSkipVerify: true,
    })

Environment
OS name, version and architecture: Linux Debian
Node version 18.19.0
Package name and version gRPC@1.11.1

@murgatroid99

vinothsa4891:
Adding a fix to ignore self-signed certificates verification

File : grpc-node/packages/grpc-js/src/transport.ts.

connectionOptions = {
        ...connectionOptions,
        ...address,
        enableTrace: options['grpc-node.tls_enable_trace'] === 1,
        rejectUnauthorized: options['grpc-node.tls_reject_unauthorized'] ?? true,
      };

murgatroid99:
We already have an established way to make a client accept self-signed certificates: create a CA certificate file with the signing certificate information, and pass the contents of that file in the first argument of `credentials.createSsl`.

vinothsa4891:
@murgatroid99  Yes that works , But In our case, it is acceptable to disregard the certificate verification and  the http2.connect works only with rejectUnauthorized as false.

vinothsa4891:
[...]

PR #2812, opened 2024-08-25
Adding fix to ignore self-signed certificates verification

vinothsa4891:
Adding fix to ignore self-signed certificates verification

### DIFF
--- a/packages/grpc-js/src/channel-credentials.ts
+++ b/packages/grpc-js/src/channel-credentials.ts
@@ -200,7 +200,7 @@ class SecureChannelCredentialsImpl extends ChannelCredentials {
         verifyOptions.checkServerIdentity;
     }
 
-    if (verifyOptions?.rejectUnauthorized) {
+    if (verifyOptions?.rejectUnauthorized !== undefined) {
       this.connectionOptions.rejectUnauthorized =
         verifyOptions.rejectUnauthorized;
     }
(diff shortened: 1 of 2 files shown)


vinothsa4891:
https://github.com/grpc/grpc-node/issues/2811

murgatroid99:
In the referenced issue, you mention that the Go implementation accomplishes this with a TLS credentials option. I think we should take the same approach in Node, rather than introducing a new channel option. Both `credentials.createSsl` and `credetials.createFromSecureContext` have a `verifyOptions` argument that this option would fit in.

vinothsa4891:
I tried with verifyOptions checkServerIdentity: () => null, but with this, I was able to disable only the hostname/CN name verification. I'm still getting error for self signed certificate. I see only rejectUnauthorized: false to bypass the verification of self-signed certificates and allows the client to make SSL requests without validating the server’s SSL certificate @murgatroid99

murgatroid99:
I am suggesting that instead of adding a channel option, we add a field to `verifyOptions` called `rejectUnauthorized` that gets passed along to the connection options.

vinothsa4891:
@murgatroid99  - I tried that even before , But its not working. Getting "err Error: 14 UNAVAILABLE: No connection established. Last error: self-signed certificate in certificate chain"

const verifyOptions = {
      // Define checkServerIdentity function
      checkServerIdentity: (host, cert) => {
        // Here you can add custom certificate validation logic if needed
    
        // Returning undefined means accepting the certificate
        // Throwing an error will reject the connection
        return undefined; // Accept the certificate
      },
      rejectUnauthorized: false,
  };
  
  const sslCreds = grpc.credentials.createSsl(
      null,
      null,
      null,
      verifyOptions
    );

    return new this._grpcObj.RegistryApi(
      "<< url >>:443",
      sslCreds,
    );

murgatroid99 (approved):

PR #2816, opened 2024-09-04
allow to trace with lazy text calculation

ygalbel:
In `internal-channel.ts` file there is a trace after channel is constructed.
In this trace, the full stacktrace is added.

The trace is written only when the log level is on `trace`, but the stacktrace is calculated in any case.
We found in our usage, that if you have a big bundle file it can take few seconds to calculate the stacktrace, it can be reasonable in trace mode but not in production.

I suggest to made the text in trace function to allow also a `delegate` that returns a function.
The delegate will be called only when the trace is materialized.

### DIFF
--- a/packages/grpc-js/src/internal-channel.ts
+++ b/packages/grpc-js/src/internal-channel.ts
@@ -427,11 +427,11 @@ export class InternalChannel {
     trace(
       LogVerbosity.DEBUG,
       'channel_stacktrace',
-      '(' +
+     () => ('(' +
         this.channelzRef.id +
         ') ' +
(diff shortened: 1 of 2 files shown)


murgatroid99:
Simpler solution: surround that call to `trace` with `if (isTracerEnabled('channel_stacktrace')) { ... }`. This is how we are handling other cases where producing the trace log text is expensive.

Even simpler solution: just delete that trace log line. I don't think it's actually that valuable.

Either way, I recommend rebasing the change onto the `@grpc/grpc-js@1.11.x` branch to get the change out faster.

PR #2817, opened 2024-09-05
made the call to trace to be called only when tracer is enabled

ygalbel:
In internal-channel.ts file there is a trace after channel is constructed.
In this trace, the full stacktrace is added.

The trace is written only when the log level is on trace, but the stacktrace is calculated in any case.
We found in our usage, that if you have a big bundle file it can take few seconds to calculate the stacktrace, it can be reasonable in trace mode but not in production.

See https://github.com/grpc/grpc-node/pull/2816 for the conversation with @murgatroid99.

image

### DIFF
--- a/packages/grpc-js/src/internal-channel.ts
+++ b/packages/grpc-js/src/internal-channel.ts
@@ -31,7 +31,7 @@ import {
   getDefaultAuthority,
   mapUriDefaultScheme,
 } from './resolver';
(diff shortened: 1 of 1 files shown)


murgatroid99 (approved):
Thank you for the fix.

murgatroid99:
This is out in version 1.11.2.

PR #2819, opened 2024-09-05
Fixed generation of grpc services

orwoods:
This pull request fixes file generation to work with grpc-js. The bug is that the second argument of the makeClientConstructor function is missing, although it is required.

required_field


orwoods:
Hi @murgatroid99 Could you take a look at this PR?

murgatroid99:
Version 1.13.0 is now out with this change

PR #2827, opened 2024-09-13
Add dualstack kokoro test job config

murgatroid99:
This is the same as the other configs, with just the test suite value changed.


test/kokoro/psm-dualstack.cfg
```diff
@@ -0,0 +1,30 @@
+# Copyright 2022 gRPC authors.
```
sergiitk, line 1:
```suggestion
# Copyright 2024 gRPC authors.
```

😛

sergiitk (approved):

PR #2830, opened 2024-09-30
chore: proofread docs

JustinLawrenceMS:
trivial contribution.  DNS had inconsistent capitalization.


JustinLawrenceMS:
Please re-run easyCLA.  I fixed that.

murgatroid99:
It updated the comment but it still says the commit is not covered. I don't have visibility into what exactly the problem is.

PR #2840, opened 2024-10-10
Add option to suppress permissive types

### LINKED ISSUE #2839: Add a flag for proto-loader-gen-types to only output the restrictive type and remove the "__Output" from the name

We have several HTTP + gRPC services where we have the same shape objects coming in on requests. We abstract the request/response handling with koa controllers / unary grpc functions, and share the logic after that.

The interfaces we make for use with the koa controller and service functions overlap exactly with the proto messages we write coming from grpc. However the types generated (the "permissive" type) has all properties optional.

There is a "restrictive" type, but it's always the message name with "__Output" on the end, which makes it very long and clunky to use.

We _never_ want to use the permissive type. If we wanted optional properties, we would mark them as optional in the proto.

A flag to suppress the output of the "permissive" type and to make the "restrictive" type use the name the permissive type would have.

Telling developers to always import the `__Output` restrictive type is what we've been doing, but this has to be enforced in code review. If the permissive type wasn't generated at all, then it removes the need to look out for this.

### Additional context

__TypeScript doesn't think the types overlap__

Proto file:

```proto
message UpdateLocationRequest {
  int32 locationId = 1;
  string name = 2;
  string description = 3;
  string locationNumber = 4;
  string line1 = 5;
  string line2 = 6;
  string city = 7;
  string state = 8;
  string postal = 9;
  int32 latitude = 10;
  int32 longitude = 11;
  string timezone = 12;
}
```

Generated types (from proto-loader-gen-types):

```ts
export interface UpdateLocationRequest {
    'locationId'?: (number);
    'name'?: (string);
    'description'?: (string);
    'locationNumber'?: (string);
    'line1'?: (string);
    'line2'?: (string);
    'city'?: (string);
    'state'?: (string);
    'postal'?: (string);
    'latitude'?: (number);
    'longitude'?: (number);
    'timezone'?: (string);
}

export interface UpdateLocationRequest__Output {
    'locationId': (number);
    'name': (string);
    'description': (string);
    'locationNumber': (string);
    'line1': (string);
    'line2': (string);
    'city': (string);
    'state': (string);
    'postal': (string);
    'latitude': (number);
    'longitude': (number);
    'timezone': (string);
}
```

Our Typescript interface (for koa controllers/service):

```ts
export interface LocationUpdate {
    name: string;
    description: string;
    locationNumber: string;
    line1: string;
    line2?: string;
    city: string;
    state: string;
    postal: string;
    latitude: number;
    longitude: number;
    timezone: string;
}
```

Usage:

```ts
// service function
const updateLocation = async (locationId: number, location: LocationUpdate): Promise<void> => {
  // ...
};

// koa controller
const updateLocationController = async (ctx: Context<LocationUpdate, void>) => {
  const { locationId } = ctx.params;

  await updateLocation(locationId, ctx.request.body);
};

// grpc unary handler
const updateLocationGrpcHandler = async (req: UpdateLocationRequest): Promise<void> => {
  const { locationId, ...location } = req;

  // type error here
  await updateLocation(locationId, location);
};
```

Type error:

```txt
Argument of type '{ name?: string; description?: string; locationNumber?: string; line1?: string; line2?: string; city?: string; state?: string; postal?: string; latitude?: number; longitude?: number; timezone?: string; }' is not assignable to parameter of type 'LocationUpdate'.
  Property 'name' is optional in type '{ name?: string; description?: string; locationNumber?: string; line1?: string; line2?: string; city?: string; state?: string; postal?: string; latitude?: number; longitude?: number; timezone?: string; }' but required in type 'LocationUpdate'.ts(2345)
```

Using restrictive type eliminates the error since the properties are not optional:

```ts
const updateLocationGrpcHandler = async (req: UpdateLocationRequest__Output): Promise<void> => {
  const { locationId, ...location } = req;

  // no more error
  await updateLocation(locationId, location);
};
```

jdforsythe:
Fixes #2839

Allows suppressing the permissive types with `--

In combination with `--outputTemplate="%s"`, replaces the permissive types with the restrictive types using the default message name (without `__Output`).

### DIFF
--- a/packages/proto-loader/bin/proto-loader-gen-types.ts
+++ b/packages/proto-loader/bin/proto-loader-gen-types.ts
@@ -47,6 +47,7 @@ type GeneratorOptions = Protobuf.IParseOptions & Protobuf.IConversionOptions & {
   outputTemplate: string;
   inputBranded: boolean;
   outputBranded: boolean;
+  suppressPermissiveTypes: boolean;
 }
 
 class TextFormatter {
@@ -146,7 +147,11 @@ function getImportLine(dependency: Protobuf.Type | Protobuf.Enum | Protobuf.Serv
     }
   } else {
     if (dependency instanceof Protobuf.Type || dependency instanceof Protobuf.Enum) {
-      importedTypes = `${inputName(dependency.name)} as ${inputName(typeInterfaceName)}, ${outputName(dependency.name)} as ${outputName(typeInterfaceName)}`;
+      if (options.suppressPermissiveTypes) {
+        importedTypes = `${outputName(dependency.name)} as ${outputName(typeInterfaceName)}`;
+      } else {
+        importedTypes = `${inputName(dependency.name)} as ${inputName(typeInterfaceName)}, ${outputName(dependency.name)} as ${outputName(typeInterfaceName)}`;
+      }
     } else if (dependency instanceof Protobuf.Service) {
       importedTypes = `${dependency.name}Client as ${typeInterfaceName}Client, ${dependency.name}Definition as ${typeInterfaceName}Definition`;
     } else {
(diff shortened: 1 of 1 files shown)


murgatroid99:
This also impacts the output of the `generateService*` functions, because they currently use both the permissive and restrictive types. I think it might be enough to modify `useNameFmter` with a branch for this flag, but I'm not 100% sure.

PR #2842, opened 2024-10-28
Remove broken build badge from README

MasterOdin:
PR replaces the build badge in the README from one used from Travis (which the repo doesn't use) with one provided by GH actions.


murgatroid99:
I think that badge would also be potentially misleading. It only describes the grpc-tools package build, and doesn't represent anything about the primary library in this repository (grpc-js).

MasterOdin:
It doesn't look like there is a GH action for `grpc-js`, only for `grpc-tools`? Should there be a workflow that runs the tests for `grpc-js` or something?

murgatroid99:
The grpc-js tests are run elsewhere, and there isn't a badge for it. I think removing the badge entirely is the best option.

MasterOdin:
Done

murgatroid99 (approved):

murgatroid99:
Thank you for taking the time to improve this.

PR #2852, opened 2024-11-17
Declare generated GrpcType inheritance

### LINKED ISSUE #2851: Not declaring generated ProtoGrpcType inheritance
### Problem description
When using proto-loader-gen-types for ts-types generating, the root `ProtoGrpcType` type is not declared to `extends grpc.GrpcObject`.
Because of that, when I `grpc.loadPackageDefinition(...)`, I cannot cast it directly to `as MyProtoGrpcType` and I have to cast it to `as unknown` first.
Though root `ProtoGrpcType` is in fact the extension of `grpc.GrpcObject` -- the only problem is the absence of declaration.

### Reproduction steps
```typescript
import * as protoLoader from '@grpc/proto-loader';
import type { ProtoGrpcType as MyProtoGrpcType } from 'proto/my';

const myPackageDefinition = protoLoader.loadSync('my.proto', { includeDirs: [ 'proto' ] });
const myProtoDescriptor = grpc.loadPackageDefinition(myPackageDefinition) as unknown as MyGrpcType;
```

### Environment
 - Any OS name, version and architecture
 - Any Node version
 - Any Node installation method (nvm or others)
 - @grpc/proto-loader:0.7.13

### Additional context
@murgatroid99 , I strongly appreciate your creature: the proto-loader-gen-types!
It is a very helpful thing! And I even can go on with `as unknown` cast.
But this duct-tape of `as unknown` can be removed -- and please let's do it.
I created a PR addressing this problem, it's very simple. Please take a look.

enderteszla:
Generated interface ProtoGrpcType is in fact an extension of grpc.GrpcObject, but this inheritance wasn't declared until now.

This forced users to cast `grpc.loadPackageDefinition(...): grpc.GrpcObject` first to `as unknown` and only then to `as {GeneratedGrpcType}`.

With this fix users will be able to cast to `as {GeneratedGrpcType}` directly.

Fixes #2851

### DIFF
--- a/packages/proto-loader/bin/proto-loader-gen-types.ts
+++ b/packages/proto-loader/bin/proto-loader-gen-types.ts
@@ -760,12 +760,12 @@ function generateRootFile(formatter: TextFormatter, root: Protobuf.Root, options
   generateServiceImports(formatter, root, options);
   formatter.writeLine('');
 
-  formatter.writeLine('type SubtypeConstructor<Constructor extends new (...args: any) => any, Subtype> = {');
+  formatter.writeLine('type SubtypeConstructor<Constructor extends new (...args: any) => any, Subtype> = grpc.ServiceClientConstructor & {');
   formatter.writeLine('  new(...args: ConstructorParameters<Constructor>): Subtype;');
   formatter.writeLine('};');
   formatter.writeLine('');
 
-  formatter.writeLine('export interface ProtoGrpcType {');
+  formatter.writeLine('export interface ProtoGrpcType extends grpc.GrpcObject {');
   formatter.indent();
   for (const nested of root.nestedArray) {
     generateSingleLoadedDefinitionType(formatter, nested, options);


packages/grpc-js/test/test-channelz.ts
```diff
@@ -82,7 +82,7 @@ describe('Channelz', () => {
         channelzClient = new channelzGrpcObject.grpc.channelz.v1.Channelz(
           `localhost:${port}`,
           grpc.credentials.createInsecure()
-        );
+        ) as ChannelzClient;
```
murgatroid99, line 85: If the purpose of this change is to avoid typecasts, why do we need to add one here? If you change the typecast on line 37 to the new suggested usage, would this one still be required?

enderteszla: @murgatroid99 , fixed. Thank you!

murgatroid99 (commented):
Please revert the changes to the grpc-health-check package except for the ones that are direct consequences of the code generator change. In particular the published code should not interact with `@grpc/grpc-js` at all. This is a deliberate design decision to make the library compatible with both gRPC implementations. Modifications to that design can be discussed separately but they are out of scope for this change.

On a separate note, this change fully breaks compatibility with the old `grpc` package. That package does not export `ServiceClient` or `ServiceClientConstructor` types, and its grpc.GrpcObject is not compatible with the one this generator generates. That library still has 180K+ weekly downloads, so it's possible that this change will break people, and if we need to revert it that will be disruptive to a lot of people. If there is a way to avoid that breakage I would have a lot more confidence in publishing this change.

packages/proto-loader/README.md
```diff
@@ -122,9 +122,7 @@ const exampleServer: ExampleHandlers = {
 };
 
 const packageDefinition = protoLoader.loadSync('./proto/example.proto');
-const proto = (grpc.loadPackageDefinition(
-  packageDefinition
-) as unknown) as ProtoGrpcType;
+const proto: ProtoGrpcType = grpc.loadPackageDefinition(packageDefinition);
```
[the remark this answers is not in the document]
enderteszla: fixed.

PR #2864, opened 2024-12-06
feat(grpc-js): allow configurable http2 initial window size

### LINKED ISSUE #2429: Allow configuring HTTP2 initial/local window size
We are facing an issue when a Node app with gRPC client is getting responses slower than expected due to the fact server is awaiting for a `WINDOW_UPDATE` frame which is not being sent promptly by the client because the client is busy processing already received data.
This happens when the app is under a certain load, and is sending multiple concurrent calls to the server, while the responses are relatively big.

Allowing to control the initial window size and the local session window size would help tuning the settings for such use case.
Those configurations are available in Node's HTTP2 implementation:
- `initialWindowSize` - could be provided as part of the [settings](https://nodejs.org/api/http2.html#settings-object) passed to the [http2.connect](https://nodejs.org/api/http2.html#http2connectauthority-options-listener) method under `options.settings`
- `localWindowSize` - could be set using [http2session.setLocalWindowSize](https://nodejs.org/api/http2.html#http2sessionsetlocalwindowsizewindowsize) API

### Additional context
The default window size in Node's HTTP2 is 64KB.
The total amount of data to be received (for all the requests) in our example is ~5MB (with 100KB-150KB per request).

Would you consider accepting a PR for that?

rickihastings:
This PR adds support for setting an `initialWindowSize` in the http2 session. This fixes #2429, as described in the issue the default window size is 64Kb which is OK for typical http2 use cases but isn't so good for usage inside internal networks with larger response sizes.

I have chosen to add a channel option to implement this as it seemed the most sensible way to make this change. Please advise if there is a better way to achieve this.

### DIFF
--- a/packages/grpc-js/src/transport.ts
+++ b/packages/grpc-js/src/transport.ts
@@ -672,6 +672,11 @@ export class Http2SubchannelConnector implements SubchannelConnector {
       const session = http2.connect(`http://${targetPath}`, {
         createConnection: (authority, option) => {
           return underlyingConnection;
+        },
+        settings: {
+          initialWindowSize:
+            options['grpc-node.http2_initial_window_size'] ||
+            http2.getDefaultSettings().initialWindowSize,
         }
       });
       this.session = session;
--- a/packages/grpc-js/src/transport.ts
+++ b/packages/grpc-js/src/transport.ts
@@ -675,7 +675,7 @@ export class Http2SubchannelConnector implements SubchannelConnector {
         },
         settings: {
           initialWindowSize:
-            options['grpc-node.http2_initial_window_size'] ||
+            options['grpc-node.flow_control_window'] ??
             http2.getDefaultSettings().initialWindowSize,
         }
       });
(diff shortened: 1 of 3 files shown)


packages/grpc-js/src/transport.ts
```diff
@@ -672,6 +672,11 @@ export class Http2SubchannelConnector implements SubchannelConnector {
       const session = http2.connect(`http://${targetPath}`, {
         createConnection: (authority, option) => {
           return underlyingConnection;
+        },
+        settings: {
+          initialWindowSize:
+            options['grpc-node.http2_initial_window_size'] ||
```
murgatroid99, line 678: Please use `??` instead of `||` for consistency with the rest of the codebase.

murgatroid99 (commented):
I think `grpc-node.flow_control_window` would be a better option name, to match the Java implementation. If we implement dynamic window sizes in the future, "initial window size" would be a better name for that option, again keeping with the Java implementation.

rickihastings:
Thanks for the comments, I'll fix these as soon as I can, I'm just waiting for approval from the open source committee in my company.

rickihastings:
Hi, thanks for your patience, I have made the requested changes.

murgatroid99 (approved):

PR #2874, opened 2024-12-28
feat(grpc-tools): add omit_serialize_instanceof generator option

### LINKED ISSUE #2865: Make instanceof check optional
We were renaming some of our Protobuf messages to conform to the naming guidelines, ie. `FooRequest` and `FooResponse`, but it broke dependent libraries, even though we had included duplicate message declarations with the old names. This was because the generated JavaScript code checks the instance of the message that gets passed in:

https://github.com/grpc/grpc-node/blob/master/packages/grpc-tools/src/node_generator.cc#L132-L137

I understand why that check was added, and it makes a lot of sense in JavaScript. For us, we use `protoc-gen-ts` to give us type-checking at build time, so the check just gets in the way of migrating to the correct message names, since protobuf and gRPC don't actually care about the message names.

For ease of migration, I wish there was a way to omit the `instanceof` check in the generated JS code. Perhaps an additional flag: when the flag is set, the relevant lines get omitted from the generated code.

lionello:
Fixes #2865 

This adds a new flag `omit_serialize_instanceof` which, when set, omits the `instanceof` check from the generated JS code. It defaults to `false`, which retains the current behavior (ie. with the check).

### DIFF
--- a/packages/grpc-tools/README.md
+++ b/packages/grpc-tools/README.md
@@ -20,3 +20,7 @@ one of the following:
    gRPC library, and instead generates `PackageDefinition` objects that can
    be passed to the `loadPackageDefinition` function provided by both the
    `grpc` and `@grpc/grpc-js` libraries.
+ - `omit_serialize_instanceof`: Omit the `instanceof` check for messages in
+   client code. This is useful when the message was renamed or is from a
+   different package, and serialization would fail with
+   `Expected argument of type …`.
--- a/packages/grpc-tools/src/node_generator.cc
+++ b/packages/grpc-tools/src/node_generator.cc
@@ -120,7 +120,8 @@ grpc::string NodeObjectPath(const Descriptor* descriptor) {
 }
 
 // Prints out the message serializer and deserializer functions
-void PrintMessageTransformer(const Descriptor* descriptor, Printer* out) {
+void PrintMessageTransformer(const Descriptor* descriptor, Printer* out,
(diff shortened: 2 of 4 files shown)


murgatroid99:
The only difference between the separate `serialize_*` functions is the `instanceof` check, so with this option set, the generator should instead generate a single `serialize` method that is used for serializing every message type.

lionello:
> The only difference between the separate `serialize_*` functions is the `instanceof` check, so with this option set, the generator should instead generate a single `serialize` method that is used for serializing every message type.

@murgatroid99  I looked into this, and it's certainly possible to fix, but it adds a lot of complexity for little gain. It also no longer allows putting breakpoints on a particular instance of the `serialize_*` function during debugging, so IMHO it's not worth the effort. Unless you tell me it's a requirement to get this merged, I'd vote to merge this as-is.

lionello:
Tests appear to be failing with
```
FATAL ERROR: MarkCompactCollector: young object promotion failed Allocation failed - JavaScript heap out of memory
```

murgatroid99:
Ignore those failures. They're not related to this change.

murgatroid99 (approved):

lionello:
@murgatroid99 any idea when there'll be a new release of `grpc-tools` to NPM?

murgatroid99:
Version 1.13.0 is now out with this change

PR #2876, opened 2025-01-06
fix: support CIDR blocks in no_proxy env variable

melkouri:
**Issue created https://github.com/grpc/grpc-node/issues/2878**

**Issue:** 

The grpc-js lib doesn't support having CIDR blocks in the **NO_PROXY** variable (like 172.16.0.0/12
192.168.0.0/16). It only checks if the address we target is in the list of IPs in the NO_PROXY. Hence, it supposes that NO_PROXY is a list of single IPs.

For example if the NO_PROXY contains 192.168.0.0/16 and the IP you're trying to reach is 192.168.0.10. This latter is NOT considered in the range of the NO_PROXY because the lib just compares the IP to the elements of the NO_PROXY list.

We also want to support the case where we whitelist domain name suffixes in the NO_PROXY (like example.com, .internal., .local. ).

**What is done in this PR:**

- Parse NO_PROXY list to get the ip and the prefix length of the CIDR
- Check if targeted ip is in the range of the CIDR.
- Check if targeted host contains the domain suffix.

### DIFF
--- a/doc/environment_variables.md
+++ b/doc/environment_variables.md
@@ -11,10 +11,11 @@ can be set.
   checked in order, and the first one that has a value is used.
 
 * no_grpc_proxy, no_proxy
-  A comma separated list of hostnames to connect to without using a proxy even
-  if a proxy is set. These variables are checked in order, and the first one
+  A comma separated list of comma-separated list of hostnames, IP addresses,
+  or CIDR blocks to connect to without using a proxy even
+  if a proxy is set, for example: no_proxy=example.com,192.168.0.1,192.168.0.0/16.
+  These variables are checked in order, and the first one
   that has a value is used.
-  The `no_proxy` environment variable could be a comma-separated list of hostnames, IP addresses, or CIDR blocks (no_proxy=example.com,192.168.0.1,192.168.0.0/16).
 
 * GRPC_SSL_CIPHER_SUITES
   A colon separated list of cipher suites to use with OpenSSL
--- a/doc/environment_variables.md
+++ b/doc/environment_variables.md
@@ -14,6 +14,7 @@ can be set.
   A comma separated list of hostnames to connect to without using a proxy even
   if a proxy is set. These variables are checked in order, and the first one
   that has a value is used.
+  The `no_proxy` environment variable could be a comma-separated list of hostnames, IP addresses, or CIDR blocks (no_proxy=example.com,192.168.0.1,192.168.0.0/16).
 
 * GRPC_SSL_CIPHER_SUITES
   A colon separated list of cipher suites to use with OpenSSL
--- a/doc/environment_variables.md
+++ b/doc/environment_variables.md
@@ -11,7 +11,7 @@ can be set.
   checked in order, and the first one that has a value is used.
 
 * no_grpc_proxy, no_proxy
-  A comma separated list of comma-separated list of hostnames, IP addresses,
+  A comma separated list of hostnames, IP addresses,
   or CIDR blocks to connect to without using a proxy even
   if a proxy is set, for example: no_proxy=example.com,192.168.0.1,192.168.0.0/16.
   These variables are checked in order, and the first one
--- a/packages/grpc-js/src/http_proxy.ts
+++ b/packages/grpc-js/src/http_proxy.ts
@@ -162,15 +162,10 @@ function hostMatchesNoProxyList(serverHost: string): boolean {
     const parsedCIDR = parseCIDR(host);
     // host is a CIDR and serverHost is an IP address
     if (isIPv4(serverHost) && parsedCIDR && isIpInCIDR(parsedCIDR, serverHost)) {
-      trace('Not using proxy for target in no_proxy list: ' + serverHost);
       return true;
-    }
-    // host is a single IP or a domain name suffix
-    else {
-      if (serverHost.endsWith(host)) {
-        trace('Not using proxy for target in no_proxy list: ' + serverHost);
-        return true;
-      }
+    } else if (serverHost.endsWith(host)) {
+      // host is a single IP or a domain name suffix
+      return true;
     }
   }
   return false;
--- a/packages/grpc-js/src/http_proxy.ts
+++ b/packages/grpc-js/src/http_proxy.ts
@@ -151,13 +151,10 @@ function isIpInCIDR(cidr: CIDRNotation, serverHost: string) {
   const mask = -1 << (32 - cidr.prefixLength);
   const hostIP = ipToInt(serverHost);
 
-  const networkAddress = ip & mask;
-  const broadcastAddress = networkAddress | ~mask;
-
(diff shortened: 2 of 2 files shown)


packages/grpc-js/src/http_proxy.ts
```diff
@@ -119,6 +119,56 @@ function getNoProxyHostList(): string[] {
...
+  const broadcastAddress = networkAddress | ~mask;
+
+  return hostIP >= networkAddress && hostIP <= broadcastAddress;
+}
+
+function checkHostInNoProxyHostList(serverHost: string): boolean {
+  for (const host of getNoProxyHostList()) {
+    const parsedCIDR = parseCIDR(host);
+    // host is a single IP address or a CIDR notation or a domain
+    if (parsedCIDR && isIpInCIDR(parsedCIDR, serverHost) || serverHost.endsWith(host)) {
```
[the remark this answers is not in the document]
melkouri: The original logic where we handle single IPs is handled by `isIpInCIDR(parsedCIDR, serverHost)` as this latter transforms single IPs to CIDR notations ([here](https://github.com/grpc/grpc-node/pull/2876/files#diff-c02b4ca7d3052a54c207670d78f12655a208dcdb014f61172d84ff7d6e83ec16R139)).
This `serverHost.endsWith(host)` was to support the case where we include domain names/suffixes in the NO_PROXY (for example `NO_PROXY=example.com,.local.,.local.`).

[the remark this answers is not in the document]
melkouri: This case wasn't stated in the issue, I'll update it and make sure it's well handled in the code. Thank you for raising this 🙇

melkouri:
Thank you @murgatroid99 for your review, I made some changes accordingly to your comments. I hope it's clearer 🙂

packages/grpc-js/src/http_proxy.ts
```diff
@@ -119,6 +119,63 @@ function getNoProxyHostList(): string[] {
...
+
+  return (hostIP & mask) === (ip & mask);
+}
+
+function hostMatchesNoProxyList(serverHost: string): boolean {
+  for (const host of getNoProxyHostList()) {
+    const parsedCIDR = parseCIDR(host);
+    // host is a single IP or a domain name suffix
+    if (!parsedCIDR) {
+      if (host === serverHost || serverHost.includes(host)) {
```
murgatroid99, line 165: I checked and I see that other implementations support suffix matching, so I am OK with keeping the `endsWith` check here, but `includes` is definitely not correct. The equality check is also redundant with the substring check, so we should not have both.

packages/grpc-js/src/http_proxy.ts
```diff
@@ -119,6 +119,63 @@ function getNoProxyHostList(): string[] {
...
+  const hostIP = ipToInt(serverHost);
+
+  return (hostIP & mask) === (ip & mask);
+}
+
+function hostMatchesNoProxyList(serverHost: string): boolean {
+  for (const host of getNoProxyHostList()) {
+    const parsedCIDR = parseCIDR(host);
+    // host is a single IP or a domain name suffix
+    if (!parsedCIDR) {
```
murgatroid99, line 164: `if...else` blocks should not use negated conditions. Please switch the two cases.

packages/grpc-js/src/http_proxy.ts
```diff
@@ -119,6 +119,63 @@ function getNoProxyHostList(): string[] {
...
+      if (host === serverHost || serverHost.includes(host)) {
+        trace('Not using proxy for target in no_proxy list: ' + serverHost);
+        return true;
+      }
+    }
+    // host is a CIDR or a domain
+    else if (isIpInCIDR(parsedCIDR, serverHost)) {
```
murgatroid99, line 171: This still doesn't handle the case where `serverHost` is not an IP address.

packages/grpc-js/src/http_proxy.ts
```diff
@@ -119,6 +119,63 @@ function getNoProxyHostList(): string[] {
...
+  for (const host of getNoProxyHostList()) {
+    const parsedCIDR = parseCIDR(host);
+    // host is a single IP or a domain name suffix
+    if (!parsedCIDR) {
+      if (host === serverHost || serverHost.includes(host)) {
+        trace('Not using proxy for target in no_proxy list: ' + serverHost);
+        return true;
+      }
+    }
+    // host is a CIDR or a domain
```
murgatroid99, line 170: I think this comment is incorrect. `host` should be a CIDR range here, not a domain.

doc/environment_variables.md
```diff
@@ -11,8 +11,10 @@ can be set.
   checked in order, and the first one that has a value is used.
 
 * no_grpc_proxy, no_proxy
-  A comma separated list of hostnames to connect to without using a proxy even
-  if a proxy is set. These variables are checked in order, and the first one
+  A comma separated list of comma-separated list of hostnames, IP addresses,
```
murgatroid99, line 14:
```suggestion
  A comma separated list of hostnames, IP addresses,
```

packages/grpc-js/src/http_proxy.ts
```diff
@@ -119,6 +119,63 @@ function getNoProxyHostList(): string[] {
...
+
+  return (hostIP & mask) === (ip & mask);
+}
+
+function hostMatchesNoProxyList(serverHost: string): boolean {
+  for (const host of getNoProxyHostList()) {
+    const parsedCIDR = parseCIDR(host);
+    // host is a CIDR and serverHost is an IP address
+    if (isIPv4(serverHost) && parsedCIDR && isIpInCIDR(parsedCIDR, serverHost)) {
+      trace('Not using proxy for target in no_proxy list: ' + serverHost);
```
murgatroid99, line 165: This trace line and the one in the else case are both redundant with the one where this function is called. Please remove these.

packages/grpc-js/src/http_proxy.ts
```diff
@@ -119,6 +119,63 @@ function getNoProxyHostList(): string[] {
...
+      return true;
+    }
+    // host is a single IP or a domain name suffix
+    else {
+      if (serverHost.endsWith(host)) {
```
murgatroid99, line 170: These lines are formatted in an unnecessarily unusual way. There's no reason to make this `if` statement separate from the `else`, just use `else if`. And the line split before the `else` to accommodate the comment is just jarring, instead move the `else` to the same line of the `}` as usual and move the comment inside the block.

In other words, these four lines should just be

```ts
} else if (serverHost.endsWith(host)) {
  // host is a single IP or a domain name suffix
```

murgatroid99 (approved):
Thank you for your contribution, and for being patient through all of the reviews.

melkouri:
Thank you for the reviews!

murgatroid99:
This is now out in version 1.13.0.

PR #2891, opened 2025-01-28
fix: xds-client LrsCallState statsTimer memory leak

gfrancz:


gfrancz:
Ok sorry about that -- CLA is signed, I've confirmed it builds.

ws-gregm:
@murgatroid99 It appears that this PR and [this one](https://github.com/grpc/grpc-node/pull/2896) are blocked on Windows Tests which is failing because of timeout with test. I'm looking into it but I suspect it's a false fail.

murgatroid99 (approved):
Thank you for the fix.

Issue #2893, opened 2025-01-29
Potential memory leak when re-initializing client


karliky:
### Problem description

Hello, and thanks for all your hard work on this library! 

I want to share a situation that might look like a memory leak in `@grpc/grpc-js`, but I want to stress that this isn't the library’s fault. Instead, it's caused by the way the client is being re-initialized in application code. 

I'm creating this issue to help others who might run into similar issues and to suggest potential improvements that could make it easier to detect or prevent.

We noticed an ongoing increase in memory usage when our application handled `UNAVAILABLE` or `DEADLINE_EXCEEDED` errors. In those cases, we were re-initializing a new client without closing the existing one. Over time, the detached references to the old clients accumulated in memory, resulting in a crash.

It was difficult to pinpoint the source of the leak because the attached memory references aren’t always obvious in devtools, especially if your monitoring tools (like Datadog) don’t highlight “detached nodes.” As soon as we added `client.close()` before creating a new client, memory usage stabilized and the issue was resolved.

### Reproduction steps
server.js
```
const grpc = require('@grpc/grpc-js');
const protoLoader = require('@grpc/proto-loader');
const path = require('path');

// Load proto definition
const PROTO_PATH = path.join(__dirname, 'ping.proto');
const packageDefinition = protoLoader.loadSync(PROTO_PATH);
const pingProto = grpc.loadPackageDefinition(packageDefinition).pingpong;

let callCount = 0;

function ping(call, callback) {
  callCount++;
  console.log('Received ping request #', callCount);

  // Every 10th call, return an error with code = UNAVAILABLE
  if (callCount % 10 === 0) {
    const error = {
      code: grpc.status.UNAVAILABLE,
      message: 'Simulated server error: UNAVAILABLE',
    };
    console.log('Sending error for request #', callCount);
    return callback(error, null);
  }

  // For normal requests, just echo back a "pong"
  callback(null, { message: `pong: ${call.request.message}` });
}

function main() {
  const server = new grpc.Server();

  server.addService(pingProto.PingService.service, { Ping: ping });

  const address = '0.0.0.0:50051';
  server.bindAsync(address, grpc.ServerCredentials.createInsecure(), (err, port) => {
    if (err) {
      return console.error(err);
    }
    console.log(`Server running at http://${address}`);
    server.start();
  });
}

main();
```

client.js
```
const grpc = require('@grpc/grpc-js');
const protoLoader = require('@grpc/proto-loader');
const path = require('path');

// Load proto definition
[...]

PR #2896, opened 2025-01-31
grpc-js: Allow garbage collection of IDLE channels

murgatroid99:
This addresses #2893 to the greatest extent reasonable.

This PR includes 3 changes to allow garbage collection of IDLE channels:

 - Modify the subchannel wrapper used by the channel so that it does not register itself until it has a ref, to handle the fact that pick_first may never ref some subchannels.
 - Don't run the `callRefTimer` until a call is started, and stop it when the channel goes IDLE.
 - Move the channelz info into a separate object, so that the reference from channelz only keeps that object alive instead of the whole channel.

A caveat here is that repeatedly creating and discarding channels without closing them will still leak some memory with channelz, because the channel is never unregistered with channelz, so channelz needs to retain the associated data. This does not happen if channelz is disabled for a channel.

### DIFF
--- a/packages/grpc-js/package.json
+++ b/packages/grpc-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/grpc-js",
-  "version": "1.12.5",
+  "version": "1.12.6",
   "description": "gRPC Library for Node - pure JS implementation",
   "homepage": "https://grpc.io/",
   "repository": "https://github.com/grpc/grpc-node/tree/master/packages/grpc-js",
--- a/packages/grpc-js/src/internal-channel.ts
+++ b/packages/grpc-js/src/internal-channel.ts
@@ -125,10 +125,13 @@ class ChannelSubchannelWrapper
     ) => {
       channel.throttleKeepalive(keepaliveTime);
     };
-    childSubchannel.addConnectivityStateListener(this.subchannelStateListener);
   }
 
   ref(): void {
+    if (this.refCount === 0) {
+      this.child.addConnectivityStateListener(this.subchannelStateListener);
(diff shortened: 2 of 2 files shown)


karliky (approved):

igorescobar-moonpay:
Do you think this issue could be related somehow? https://github.com/googleapis/nodejs-pubsub/issues/2016

karliky:
@igorescobar-moonpay are you properly closing the clients?

igorescobar-moonpay:
> @igorescobar-moonpay The purpose of this change was to minimize the memory impact of continuously creating and discarding client objects. However, the best strategy is to not do that at all. For any given service on a particular server, you should only create one client object, and use it for the lifetime of the process. If for whatever reason your only option is to keep creating new clients, you should always explicitly `close` them when you discard them.
> 
> The heap dump screenshots in the linked issue show that a large amount of retained memory is in the `Client` class, which implies that you are in fact creating many client objects.

I appreciate you taking the time to look into the screenshots and giving us some pointers! I will have a look!

igorescobar-moonpay:
> @igorescobar-moonpay The purpose of this change was to minimize the memory impact of continuously creating and discarding client objects. However, the best strategy is to not do that at all. For any given service on a particular server, you should only create one client object, and use it for the lifetime of the process. If for whatever reason your only option is to keep creating new clients, you should always explicitly `close` them when you discard them.
> 
> The heap dump screenshots in the linked issue show that a large amount of retained memory is in the `Client` class, which implies that you are in fact creating many client objects.

Thanks a lot for the pointer @murgatroid99! The issue was fixed, more here: https://github.com/googleapis/nodejs-pubsub/issues/2016#issuecomment-2691671022

PR #2902, opened 2025-02-11
grpc-js-xds: NACK with detailed validation errors

murgatroid99:
Previously, when validating resources, the validation function would short circuit on the first error, and the NACK would contain a generic message indicating that the resource failed to validate. Now the validation functions find every validation error, and report the details of all of them.


murgatroid99:
- [ ] [grpc/node/master/xds_k8s_lb](https://source.cloud.google.com/results/invocations/5908a846-4dc1-4750-b90c-572c51940e5b)

murgatroid99:
The interop test run failed, but the failure matches the one I fixed in the last commit.

PR #2912, opened 2025-03-01
feat(proto-loader-gen-types): add options for specifying file extensions

### LINKED ISSUE #2907: [Feature Request]: `proto-loader-gen-types`: More options for generated code

After extensive trial and error, I found that `proto-loader-gen-types` is the simplest solution for generating type definitions for code that uses `@grpc/grpc-js` in TypeScript.

But the TypeScript code generated by `proto-loader-gen-types` has the following issues:

1. The file extension is `.ts`, but since it is meant for **type declarations**, it is supposed to be `.d.ts` instead.
2. With [`module": "nodenext"`](https://www.typescriptlang.org/docs/handbook/modules/reference.html#node16-nodenext-1) set in `tsconfig.json` and `"type": "module"` set in `package.json` for Node projects, VS Code raises warnings about **missing file extensions in import module specifiers**. When using ESM in Node, both [Node.js](https://nodejs.org/api/esm.html#mandatory-file-extensions) and [TypeScript's module resolution](https://www.typescriptlang.org/docs/handbook/modules/theory.html#module-resolution) require explicit file extensions in import paths. However, the code generated by `proto-loader-gen-types` does not automatically include them. This results in a manual update step after code generation, which is inconvenient and an antipattern.

1. Provide an option to configure which file extension (`.ts` or `.d.ts`) is used for generated code. Maybe something like [this](https://github.com/bufbuild/protobuf-es/blob/main/MANUAL.md#target)
2. Provide an option to configure whether file extensions are appended to module specifiers in generated code. Maybe something like [this](https://github.com/bufbuild/protobuf-es/blob/main/MANUAL.md#import_extension)

### Additional context

There are related discussions, such as #2693 and #2401. This issue aims to revive the topic and encourage further discussion.

tenkirin:
Add the following two options to `proto-loader-gen-types`:

- `targetFileExtension`: File extension for generated files. Defaults to `.ts`
- `importFileExtension`: File extension for import specifiers in generated code. Defaults to none (omitted)

The default behavior of `proto-loader-gen-types` is kept unchanged to avoid any possible breaking on existing usage, and the output is only modified when options are provided as needed.

Closes #2907

### DIFF
--- a/packages/proto-loader/bin/proto-loader-gen-types.ts
+++ b/packages/proto-loader/bin/proto-loader-gen-types.ts
@@ -107,8 +107,8 @@ function getImportPath(to: Protobuf.Type | Protobuf.Enum | Protobuf.Service): st
   return stripLeadingPeriod(to.fullName).replace(/\./g, '/');
 }
 
-function getPath(to: Protobuf.Type | Protobuf.Enum | Protobuf.Service, extension: string = '.ts') {
-  return stripLeadingPeriod(to.fullName).replace(/\./g, '/') + extension;
+function getPath(to: Protobuf.Type | Protobuf.Enum | Protobuf.Service, options: GeneratorOptions) {
+  return stripLeadingPeriod(to.fullName).replace(/\./g, '/') + options.targetFileExtension;
 }
 
 function getPathToRoot(from: Protobuf.NamespaceBase) {
@@ -155,7 +155,7 @@ function getImportLine(dependency: Protobuf.Type | Protobuf.Enum | Protobuf.Serv
       throw new Error('Invalid object passed to getImportLine');
     }
   }
-  return `import type { ${importedTypes} } from '${filePath}${options.importFileExtension ?? ''}';`
+  return `import type { ${importedTypes} } from '${filePath}${options.importFileExtension}';`
 }
 
 function getChildMessagesAndEnums(namespace: Protobuf.NamespaceBase): (Protobuf.Type | Protobuf.Enum)[] {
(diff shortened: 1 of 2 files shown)


packages/proto-loader/bin/proto-loader-gen-types.ts
```diff
@@ -105,8 +107,8 @@ function getImportPath(to: Protobuf.Type | Protobuf.Enum | Protobuf.Service): st
   return stripLeadingPeriod(to.fullName).replace(/\./g, '/');
 }
 
-function getPath(to: Protobuf.Type | Protobuf.Enum | Protobuf.Service) {
-  return stripLeadingPeriod(to.fullName).replace(/\./g, '/') + '.ts';
+function getPath(to: Protobuf.Type | Protobuf.Enum | Protobuf.Service, extension: string = '.ts') {
```
murgatroid99, line 110: I would prefer to just pass `GeneratorOptions` here, for consistency with the other functions.

packages/proto-loader/bin/proto-loader-gen-types.ts
```diff
@@ -877,6 +879,8 @@ async function runScript() {
     .option('outputTemplate', { string: true, default: `${templateStr}__Output` })
     .option('inputBranded', boolDefaultFalseOption)
     .option('outputBranded', boolDefaultFalseOption)
+    .option('targetFileExtension', { string: true, default: '.ts' })
+    .option('importFileExtension', { string: true })
```
murgatroid99, line 883: Set the empty string default here, instead of where it is used.

murgatroid99 (changes requested):

tenkirin:
@murgatroid99 
Thanks for the suggestions! That’s a great point—I’ve committed an update.

packages/proto-loader/bin/proto-loader-gen-types.ts
```diff
@@ -916,6 +920,8 @@ async function runScript() {
       outputTemplate: 'Template for mapping output or "restricted" type names',
       inputBranded: 'Output property for branded type for  "permissive" types with fullName of the Message as its value',
       outputBranded: 'Output property for branded type for  "restricted" types with fullName of the Message as its value',
+      targetFileExtension: 'File extension for generated files. Defaults to .ts',
+      importFileExtension: 'File extension for import specifiers in generated code. Defaults to none (omitted)'
```
murgatroid99, line 924: For consistency with other description messages, don't state the default values. As the README changes show the defaults are included automatically. And update the README again with that change.

murgatroid99 (approved):

simonhaenisch:
Nice to see this has landed, however it seems like a new version of `@grpc/proto-loader` hasn't been published yet? At least it's not working for me with `--targetFileExtension='.d.ts'`.

murgatroid99:
This is now out in version 0.7.14

PR #2919, opened 2025-03-08
fix: return valid status if missing response message

polRk:


murgatroid99:
The current behavior is correct, as specified in [the status codes specification](https://github.com/grpc/grpc/blob/master/doc/statuscodes.md):

> | Case | Code | Generated at Client or Server |
> |-|-|-|
> |...|...|...|
> | Response cardinality violation (method requires exactly one response but server sent some other number of responses) | UNIMPLEMENTED | Client |

In addition, the library is never allowed to generate the DATA_LOSS code, as specified in the same document:

> The following status codes are never generated by the library:
> - INVALID_ARGUMENT
> - NOT_FOUND
> - ALREADY_EXISTS
> - FAILED_PRECONDITION
> - ABORTED
> - OUT_OF_RANGE
> - **DATA_LOSS**

PR #2930, opened 2025-03-25
Default values are added in channel-options

Anitha24597:
Raised PR for this issue https://github.com/grpc/grpc/issues/21417 
Added default values and info regarding channel arguments in channel_options.ts file


murgatroid99:
That issue was for channel options that I didn't implement in this library because their functionality was unclear. It can't be addressed by documentation in this library.

Also, I think it's going to be confusing rather than helpful to document options that aren't declared or implemented.

Anitha24597:
am closing this PR as these channel options are not implemented

PR #2938, opened 2025-04-09
feat(proto-loader): export utils for loading protos

DakEnviy:
The functions `loadProtosWithOptions` and `loadProtosWithOptionsSync` are very useful for loading proto files because they have the path resolver with the `includeDirs` option. I think they should be exported.


murgatroid99:
The existing `load` and `loadSync` functions have the same functionality as those "WithOptions" functions, because they just call those functions. I don't understand what functionality you think this would add.

DakEnviy:
> The existing `load` and `loadSync` functions have the same functionality as those "WithOptions" functions, because they just call those functions. I don't understand what functionality you think this would add.

With `loadProtosWithOptions` and `loadProtosWithOptionsSync`, you can get a proto-root object to later use all its functions like `lookup`. This is useful for decoding proto-messages by yourself and much convenient than using `loadSync` from `protobufjs` because you don't need to write `addIncludePathResolver` by yourself.

murgatroid99:
I see. This library intentionally does not expose any parts of the `protobufjs` API in its own API, meaning that we do not want to export a function that returns a `protobufjs` `Root` object.

PR #2950, opened 2025-05-02
added allowedOrigin to ServerOptions

emmveqz:


murgatroid99:
The [gRPC over HTTP/2 spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) does not say that gRPC servers should support the "options" method, and gRPC clients certainly don't support it, so I don't think this change would be correct.

In addition, I don't see what the value would be of having origin enforcement in gRPC, especially if you just tell the client what the required origin header is.

emmveqz:
@murgatroid99 

Sorry I didn't add details.
There's a somewhat particular use case that might use `origin` / `Access-Control-Allow-Origin`  
and the HTTP method `OPTIONS`,  
when you are a gRPC client directly from a web browser (front-end) application.  
Right now the defacto style to do that is setting up a proxy, combined with [grpc-web](https://github.com/grpc/grpc-web)  
the proxy setup takes care of the origin, hence `OPTIONS` hasn't been needed,
but using a style/approach with [protobuf.js](https://github.com/protobufjs/protobuf.js) that's no longer the case, since we can skip using a proxy.

Front-end (web browser) applications have a different origin than the gRPC server's.
Web browsers don't have a response when the origin is different and not specified in a response `Access-Control-Allow-Origin` header from a HTTP call with method `OPTIONS`

murgatroid99:
You cannot use gRPC directly from a web browser. You need a proxy when communicating with a gRPC server from the browser because browsers don't support the gRPC protocol. grpc-web uses a different protocol, and the proxy translates. Protobuf.js has nothing to do with it.

emmveqz:
It has to do with [protobuf.js](https://github.com/protobufjs/protobuf.js) in the sense that it provides an encoder/decoder using `Uint8Array` (body) same that can be upstreamed/downstreamed from web browsers, using the web browser native `fetch` therefore `HTTP/2`  
protobuf.js also helps for the `application/grpc`  `content-type`  
Thus the freedom to remove the proxy (if needed)

murgatroid99:
The binary encoding is not the obstacle to using gRPC in the browser. The main problem is that fetch doesn't support trailers or full bidirectional streaming.

So, to resolve this conversation, we do not consider gRPC directly in the browser to be supported, so I will not currently accept a change to expand that support. You can formally propose that we provide this support in [the proposal repository](https://github.com/grpc/proposal).

emmveqz:
got it,
I missed the `trailers` part in the rpc call life cycle. thank you

emmveqz:
@murgatroid99 
Hi again, sorry to bother you,
given that it's not useful, can you delete this PR? (for personal reasons). Thank you for the understanding.

murgatroid99:
No, I can't delete it, and from what I can find, GitHub will only delete a PR if it contains sensitive information. You can delete your branch so that the code doesn't show up in the PR anymore.

PR #2951, opened 2025-05-02
Add List method to Health service

MatiasManevi:
# Context

Clients cannot get a list of the services a certain server is watching to and their respective status.

In the context of Kubernetes, we have livez and readyz endpoints, by listing all the services a server is checking, we could return a response like the `/readyz?verbose` Kubernetes API returns:

```
[+]ping ok
[+]log ok
[+]etcd ok
[+]poststarthook/start-kube-apiserver-admission-initializer ok
[+]poststarthook/generic-apiserver-start-informers ok
[+]poststarthook/start-apiextensions-informers ok
[+]poststarthook/start-apiextensions-controllers ok
[+]poststarthook/crd-informer-synced ok
[+]poststarthook/bootstrap-controller ok
[+]poststarthook/rbac/bootstrap-roles ok
[+]poststarthook/scheduling/bootstrap-system-priority-classes ok
[+]poststarthook/start-cluster-authentication-info-controller ok
[+]poststarthook/start-kube-aggregator-informers ok
[+]poststarthook/apiservice-registration-controller ok
[+]poststarthook/apiservice-status-available-controller ok
[+]poststarthook/kube-apiserver-autoregistration ok
[+]autoregister-completion ok
[+]poststarthook/apiservice-openapi-controller ok
healthz check passed
```

# Use case
This is not only useful for having status reports like Kubernetes readyz endpoints; it also allows services to simplify the set up when creating microservices that publish their status to status page systems such as:

- https://cachethq.io/
- https://www.atlassian.com/software/statuspage

# Change
This PR introduces a new `List` endpoint to the `Health` service in order to list all the services that a certain server is watching to.

# Alternative solution
We could extract the `List` endpoint into a new service to avoid breaking the `v1` API. A new `HealthLister` (or any other name) must be created.

# Proposal
https://github.com/grpc/proposal/pull/468

### DIFF
--- a/packages/grpc-health-check/proto/health/v1/health.proto
+++ b/packages/grpc-health-check/proto/health/v1/health.proto
@@ -25,6 +25,13 @@ option java_multiple_files = true;
 option java_outer_classname = "HealthProto";
 option java_package = "io.grpc.health.v1";
 
+message HealthListRequest {}
+
+message HealthListResponse {
+  // statuses contains all the services and their respective status.
+  map<string, HealthCheckResponse> statuses = 1;
+}
+
 message HealthCheckRequest {
   string service = 1;
 }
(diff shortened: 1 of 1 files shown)


murgatroid99 (approved):
Thank you for adding this.

marcoshuck:
Thanks for adding these changes, Mati! Happy to work with you 💯

Issue #2959, opened 2025-05-30
Random jitter is taken for next retry backoff - should be +- 0.2 according to spec


ekscentrysytet:
According to grpc client retries spec - a jitter of +- 0.2 should be taken when calculating [next retry delay](https://github.com/grpc/proposal/blob/master/A6-client-retries.md#exponential-backoff).

During my tests I've noticed current lib generates random jitters, and after observing source code I can see it's caused by using `Math.random()` [here](https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/retrying-call.ts#L420):

```ts
private getNextRetryBackoffMs() {
    const retryPolicy = this.callConfig?.methodConfig.retryPolicy;
    if (!retryPolicy) {
      return 0;
    }
    const nextBackoffMs = Math.random() * this.nextRetryBackoffSec * 1000;
    const maxBackoffSec = Number(
      retryPolicy.maxBackoff.substring(0, retryPolicy.maxBackoff.length - 1)
    );
    this.nextRetryBackoffSec = Math.min(
      this.nextRetryBackoffSec * retryPolicy.backoffMultiplier,
      maxBackoffSec
    );
    return nextBackoffMs;
  }
```

According to spec `random(0.8, 1.2)` should be used.

Please confirm is this is expected

murgatroid99:
You're right, this should be changed. The spec you linked to was changed in https://github.com/grpc/proposal/pull/452, and I never went back and adjusted the behavior to match.

PR #2960, opened 2025-06-04
fix(grpc-js): fix jitter for client retries

ekscentrysytet:
Resolves: https://github.com/grpc/grpc-node/issues/2959


murgatroid99 (approved):
Thank you for the fix.

PR #2972, opened 2025-06-28
removed the .start() method as it is deprecated

virajdoshi11:
.start() method is directly integrated in the bindAsync function


virajdoshi11:
Please let me know if this can be added and I will then also update the documentation and the examples in other files

PR #3016, opened 2025-09-17
Add TLS examples

murgatroid99:
These examples are based on the corresponding [Go examples](https://github.com/grpc/grpc-go/tree/master/examples/features/encryption). The data files are copied from [this directory in the Go repository](https://github.com/grpc/grpc-go/tree/master/examples/data/x509). The ALTS example is omitted because ALTS is not implemented in Node.js.


examples/data/x509/README.md
```diff
@@ -0,0 +1,6 @@
+This directory contains x509 certificates and associated private keys used in
```
emil10001, line 1: Would it be possible to not include the certificates in the repo directly, and instead just have the script generate the certs. Reason is that I worry if we include certs that folks might take them and use them instead of generating their own.

examples/encryption/README.md
```diff
@@ -0,0 +1,48 @@
+# Encryption
+
+This directory contains two related examples: one for TLS and one for mTLS.
+
+## Try it
```
emil10001, line 5: If you remove the certs and key files, please add the instructions to generate them here.

emil10001 (commented):
Left a couple comments. High level, I think it would be good to keep certs and keys out of the repo, and simply provide an easy way for folks to generate them with the script you have.

Issue #2953, opened 2025-05-08
GRPC server side keepalive doesn't work as expected


zabranskiy:
### Problem description
GRPC server doesn't destroy a connection after 'Ping timeout passed without response', instead it runs session.close() and the connection keeps hanging. 

### Reproduction steps
I started server, sent 1 request to grpc and moved client process to SIGSTOP state after first 'Received ping response'. I expect that after 'Ping timeout passed without response' underlying http2 session should be destroyed.

server.js:1157
```
keepaliveTimer = setTimeout(() => {
    clearKeepaliveTimeout();
    this.keepaliveTrace('Ping timeout passed without response');
    this.trace('Connection dropped by keepalive timeout');
    sessionClosedByServer = true;
    session.close(); // <-- here the problem
}, this.keepaliveTimeoutMs);
```

### Environment
 - OS name, version and architecture: macOS Sequoia/arm64 
 - Node version v22.11.0
 - Node installation method npm
 - Package name and version @grpc/grpc-js@1.13.3

### Additional context
server options:
'grpc.keepalive_time_ms': 10e3, 
'grpc.keepalive_timeout_ms': 5e3,      
'grpc.keepalive_permit_without_calls': 1,

```
D 2025-05-08T14:09:12.787Z | v1.13.3 94805 | server | (1) Server constructed
D 2025-05-08T14:09:12.789Z | v1.13.3 94805 | server | (1) bindAsync port=0.0.0.0:8803
D 2025-05-08T14:09:12.790Z | v1.13.3 94805 | dns_resolver | Resolver constructed for target dns:0.0.0.0:8803
D 2025-05-08T14:09:12.790Z | v1.13.3 94805 | backoff | {0} constructed initialDelay=1000 multiplier=1.6 jitter=0.2 maxDelay=120000
D 2025-05-08T14:09:12.790Z | v1.13.3 94805 | backoff | {0} runOnce()
D 2025-05-08T14:09:12.790Z | v1.13.3 94805 | backoff | {0} runTimer(delay=1000)
D 2025-05-08T14:09:12.790Z | v1.13.3 94805 | dns_resolver | Returning IP address for target dns:0.0.0.0:8803
D 2025-05-08T14:09:12.791Z | v1.13.3 94805 | backoff | {0} stop()
D 2025-05-08T14:09:12.791Z | v1.13.3 94805 | backoff | {0} reset() running=false
[...]

PR #3022, opened 2025-10-24
Fix server keep alive timeout not properly destroying connections

mattias-wiberg:
## Issue
**GitHub Issue:** #2953  
**Reporter:** @zabranskiy  
**Package:** @grpc/grpc-js  
**Severity:** Bug - Server connections hang indefinitely when clients don't respond to keepalive pings

## Root Cause
The server-side keepalive ping timeout handler uses `session.close()` to terminate connections when clients fail to respond to keepalive pings. However, `session.close()` performs a graceful shutdown that does not forcibly terminate the underlying connection, causing it to hang indefinitely.

## Fix Summary
Changed keepalive error handlers to use `session.destroy()` instead of `session.close()` to forcibly close connections when:
- Keepalive ping timeout expires without response
- Ping callback receives an error
- Sending the ping fails

## Files Modified
- `packages/grpc-js/src/server.ts` (6 occurrences changed)
  - 3 changes in `_sessionHandler` method (lines ~1609, ~1635, ~1643)
  - 3 changes in `_channelzSessionHandler` method (lines ~1807, ~1830, ~1843)

## Testing Recommendations
To verify the fix:
1. Start a gRPC server with these options:
   ```javascript
   {
     'grpc.keepalive_time_ms': 10000,
     'grpc.keepalive_timeout_ms': 5000,
     'grpc.keepalive_permit_without_calls': 1
   }
   ```
2. Connect a client and send a request
3. Stop the client process (SIGSTOP) after receiving the first ping response
4. Observe server logs - after the timeout, the connection should be destroyed (not hanging)

## Consistency
This change aligns server-side keepalive handling with the client-side transport implementation (`packages/grpc-js/src/transport.ts`), which already uses `session.destroy()` via `handleDisconnect()` for keepalive timeouts.

## Backward Compatibility
This is a bug fix that should not affect normal operation:
- Proper keepalive implementations will continue to work normally
- Only affects scenarios where clients fail to respond to keepalive pings
- Changes do not affect the public API or configuration options

### DIFF
--- a/packages/grpc-js/src/server.ts
+++ b/packages/grpc-js/src/server.ts
@@ -1609,7 +1609,7 @@ export class Server {
               if (err) {
                 this.keepaliveTrace('Ping failed with error: ' + err.message);
                 sessionClosedByServer = true;
-                session.close();
+                session.destroy();
               } else {
                 this.keepaliveTrace('Received ping response');
                 maybeStartKeepalivePingTimer();
@@ -1631,7 +1631,7 @@ export class Server {
             'Connection dropped due to ping send error: ' + pingSendError
           );
           sessionClosedByServer = true;
-          session.close();
+          session.destroy();
           return;
         }
 
@@ -1640,7 +1640,7 @@ export class Server {
           this.keepaliveTrace('Ping timeout passed without response');
           this.trace('Connection dropped by keepalive timeout');
           sessionClosedByServer = true;
-          session.close();
+          session.destroy();
         }, this.keepaliveTimeoutMs);
         keepaliveTimer.unref?.();
       };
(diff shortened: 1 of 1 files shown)


OlegStrokan:
@mattias-wiberg I think we’re the only two people on Earth who have discovered this issue. I was trying to implement internet disconnection detection for streaming, using only bare keepalive, without any additional healthcheck/ heartbeat.  I even created PR to nest library because it doesn’t map keepalive settings to grpc-js. But later I discovered that the errors still didn’t propagate to the application layer in gprc-js library (I checked with gRPC tracing).
So I ended up generating more CO2 using Claude than all soccer moms combined to understand was ist das
It wasn’t a priority task, so I never proposed a change to the library. Now I see your solution and you’re a lifesaver.
I’m too scared to use a custom grpc-js fork in prod, but it seems like the maintainers aren’t very interested in code reviewing. (You’ll probably get a GitHub notification and think “Oh a maintainer finally commented on my pr, but then you will see it’s just some guy with the same issue, sorry 🙃)

Issue #2968, opened 2025-06-20
Please consider updating grpc-tools @mapbox/node-pre-gyp to something more resent


aaltat:
### Problem description
The grpc-tools uses pretty old @mapbox/node-pre-gyp https://github.com/grpc/grpc-node/blob/master/packages/grpc-tools/package.json#L27 which in in favor uses dependencies which are outdated and not anymore maintained.  Could you consider updating @mapbox/node-pre-gyp to more recent version?

aaltat:
Thank you.

PR #3026, opened 2025-11-10
feat: bump grpc-tools @mapbox/node-pre-gyp version

robcresswell:
fixes: #2968

### DIFF
--- a/packages/grpc-tools/package.json
+++ b/packages/grpc-tools/package.json
@@ -1,6 +1,6 @@
 {
   "name": "grpc-tools",
-  "version": "1.13.0",
+  "version": "1.13.1",
   "author": "Google Inc.",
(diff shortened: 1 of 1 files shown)


robcresswell:
Note this would raise the minimum supported node version for this library to `>=18.x`

18 itself is already EOL, but typically Linux-based operating systems only carry very, _very_ old versions of Node, so it might be problematic in some cases.

robcresswell:
@murgatroid99 before I burn a bunch of time figuring out how to test this on Windows, could you confirm those are unexpected failures? I can't see much helpful in the logs, but I'm also not sure what to look out for 🤓

robcresswell:
The library maintainers seem to be unsure how they can solve their publishing woes from glancing through https://github.com/mapbox/node-pre-gyp/issues/907

PR #3028, opened 2025-11-21
feat: support "http_proxy" channel option

notsapinho:


packages/grpc-js/src/http_proxy.ts
```diff
@@ -57,6 +57,9 @@ function getProxyInfo(): ProxyInfo {
   } else if (process.env.http_proxy) {
     envVar = 'http_proxy';
     proxyEnv = process.env.http_proxy;
+  } else if(proxy) {
+    envVar = 'proxy';
```
murgatroid99: The only place this variable is used is the parse error, to state which environment variable it attempted to parse. With this change, the value no longer always comes from an environment variable. So, this value would just be completely misleading. Instead, the parse error should be modified to indicate whether it tried to parse an environment variable or channel arg.

notsapinho:
It should be better now! Thanks for the quick response.

murgatroid99 (approved):

PR #3048, opened 2026-04-18
fix: bump protobufjs to ^7.5.5 to address GHSA-xq3m-2v4x-88gg

tawni-dev:
**Summary**

Bumps protobufjs from ^7.5.3 to ^7.5.5 to address the arbitrary code execution vulnerability in protobufjs < 7.5.5.
Security advisory: GHSA-xq3m-2v4x-88gg

**Details**

Attackers can inject arbitrary code into the type fields of protobuf definitions, which executes during object decoding. Fixed in protobufjs 7.5.5.

**Testing**

8/8 unit tests pass

### DIFF
--- a/packages/proto-loader/package.json
+++ b/packages/proto-loader/package.json
@@ -47,7 +47,7 @@
   "dependencies": {
     "lodash.camelcase": "^4.3.0",
     "long": "^5.0.0",
-    "protobufjs": "^7.5.3",
+    "protobufjs": "^7.5.5",
     "yargs": "^17.7.2"
   },
   "devDependencies": {


coldfannn:
@murgatroid99 @nicolasnoble PTAL.

murgatroid99 (approved):

youngseo-choi:
Hi @murgatroid99, Thank you for landing this fix!

I noticed a small catch for downstream users: @grpc/grpc-js currently depends on @grpc/proto-loader version ^0.7.x. Since this version bump was released in proto-loader@0.8.1, projects using grpc-js are still pulling in the vulnerable protobufjs version via the 0.7.x branch.

To help downstream projects completely resolve this vulnerability, would it be possible to either:

1. Backport this fix and release a patch version for the @grpc/proto-loader 0.7.x line
2. Update the @grpc/proto-loader dependency in @grpc/grpc-js to 0.8.x

Any guidance or help on this would be greatly appreciated. Thank you!

murgatroid99:
I see a dependency on `@grpc/proto-loader` version `^0.8.0` in the latest version of `@grpc/grpc-js`, version 1.14.4. Where are you seeing the dependency on version `^0.7.x`?

youngseo-choi:
Hi @murgatroid99,

Thank you for the reply!

My current project has the @grpc/grpc-js dependency set to ^1.10.9, so I mistakenly only checked the latest releases of the 1.10.x and 1.11.x branches. In those versions, I saw it still depends on @grpc/proto-loader version ^0.7.x:

- [1.10.12](https://github.com/grpc/grpc-node/blob/2c99fbddc969c3f377297d400e7b5a320e5c5de2/packages/proto-loader/package.json#L50)
- [1.11.4](https://github.com/grpc/grpc-node/blob/3f6b0e3a3d301c0e9f6783023e8701e6bcccd794/packages/proto-loader/package.json#L50)

As you pointed out, the latest versions use ^0.8.0. I will review upgrading our dependency to the 1.14.4 version to resolve this issue.

Thanks again for your help and guidance!

PR #3054, opened 2026-05-18
grpc-js: Bump version to 1.12.7

murgatroid99:
To publish some changes


murgatroid99:
The test jobs are failing because of an `@grpc/grpc-js-xds` build error, that seems to be related to `@types/node` versions, but the fix that works locally doesn't seem to work in the CI. I only care about the `@grpc/grpc-js` build, so I think it's OK to accept that failure.

Issue #2030, opened 2022-01-23
"long" types need to be listed as dependency - yarn v2


fflores97:
### Problem description
Need to list `@types/long` as dependency

### Reproduction steps
Typescript build fails when using Yarn v2+ in pnp mode

### Environment
 - OS name, version and architecture: Arch Linux
 - Node version 14.18.2
 - Node installation method: nvm
 - Package name and version: @grpc/grpc-js@1.5.0

### Additional context
Building with yarn and Typescript gives:

```
error TS2688: Cannot find type definition file for 'long'.

/// <reference types="long" />
```

I solved it by adding this to my `.yarnrc.yml`, which I believe means `@types/long` needs to be added as an explicit dependency in this package

```
packageExtensions:
  "@grpc/grpc-js@*":
    dependencies:
      "@types/long": "*"
```

murgatroid99:
I don't understand why it would need that dependency. `@grpc/grpc-js` does not import the package `long` anywhere. Instead, it imports the type `Long` from `@grpc/proto-loader`. That package does have a dependency on `@types/long`, which it uses to define that type. It should be the case that if a package exports a type and declares whatever dependencies are necessary to define it, that should be sufficient to provide that type.

Since this problem only seems to happen with Yarn's new PnP, I think that is where the problem is, and you should file an issue with Yarn, or with [this package](https://github.com/yarnpkg/berry/tree/master/packages/plugin-compat), which appears to be where they monkey-patch PnP support into `typescript`.

jeffrson:
Well, grpc-js references "long" in 
build\src\generated\google\protobuf\Int64Value.d.ts
and more than 10 other files. yarn 2 or 3 report this as
```
.yarn/cache/@grpc-grpc-js-npm-1.5.3-4edae7a246-19d75ac09e.zip/node_modules/@grpc/grpc-js/build/src/generated/google/protobuf/Int64Value.d.ts:1:23 - error TS2688: Cannot find type definition file for 'long'.

1 /// <reference types="long" />
                        ~~~~
```

This is not related to yarn pnp.

With nodeLinker node_modules @types/long is pulled in via
```
>yarn why @types/long
├─ @grpc/proto-loader@npm:0.6.9
│  └─ @types/long@npm:4.0.1 (via npm:^4.0.1)
│
└─ protobufjs@npm:6.11.2
   └─ @types/long@npm:4.0.1 (via npm:^4.0.1)
```
so it doesn't matter, that grpc-js doesn't.

[...]

PR #3067, opened 2026-07-07
proto-loader: Update protobufjs dependency to 8.x

### LINKED ISSUE #3062: Upgrade to protobufjs version 8

ProtobufJS 8 introduces support for edition 2024. Currently, proto-loader does not support version 2024. I'm not sure why that is a breaking change, however

Upgrade to protobufjs v8

Manually forcing v8 resolution in the upstream project. This is not a google solution as it introduces unsafe behavior

### Additional context

n/a

slagiewka:
The only [breaking change](https://github.com/protobufjs/protobuf.js/blob/eaefa1f9cb7bb888b16f53de22215de1c35643aa/CHANGELOG.md#800-2025-12-16) in the 8.x train is the Edition 2024 support

I took the liberty to bump the version number by a minor the same way 7b4704cc921cddf7c6084c69704ac940c4450b98 did.

Fixes #3062

### DIFF
--- a/packages/proto-loader/package.json
+++ b/packages/proto-loader/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@grpc/proto-loader",
-  "version": "0.8.1",
+  "version": "0.9.0",
   "author": "Google Inc.",
   "contributors": [
     {
@@ -47,7 +47,7 @@
   "dependencies": {
     "lodash.camelcase": "^4.3.0",
     "long": "^5.0.0",
-    "protobufjs": "^7.5.5",
+    "protobufjs": "^8.7.0",
     "yargs": "^17.7.2"
   },
   "devDependencies": {
--- a/packages/proto-loader/src/index.ts
+++ b/packages/proto-loader/src/index.ts
@@ -366,6 +366,7 @@ function createPackageDefinitionFromDescriptorSet(
 
   const root = Protobuf.Root.fromDescriptor(
     decodedDescriptorSet,
+    {keepCase: options.keepCase}
   );
   root.resolveAll();
   return createPackageDefinition(root, options);
--- a/packages/proto-loader/src/index.ts
+++ b/packages/proto-loader/src/index.ts
@@ -59,40 +59,6 @@ export function isAnyExtension(obj: object): obj is AnyExtension {
   return ('@type' in obj) && (typeof (obj as AnyExtension)['@type'] === 'string');
 }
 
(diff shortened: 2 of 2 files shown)


slagiewka:
Since v8 has vastly improved typings that `proto-loader` can now use. This surfaces the difference in `enumType.toDescriptor()` as well as allows us to pass `keepCase` to `Root.fromDescriptor`.

CC: @dcodeIO

slagiewka:
@murgatroid99 I've pushed one extra commit for the Windows build - seems like it's just due to the runner image update. This should confirm that.

ScottMansfield:
One argument for this that hasn't been made on either this PR or #3062: the 8.x train also removes `protobufjs`'s entire transitive dependency set, which is a meaningful install-size win for everything downstream of `proto-loader`.

| protobufjs | dependencies | installed |
|---|---|---:|
| `7.6.5` | `long`, **`@types/node`**, + 9 × `@protobufjs/*` | 5.66 MB, 13 pkgs |
| `8.7.2` | `long` | 3.67 MB, 2 pkgs |
| | | **−1.99 MB, −11 pkgs** |

Measured with a clean `npm i protobufjs@<v>` into an empty directory, Node v26.7.0.

The `@types/node` half of that has its own history — protobufjs/protobuf.js#1026, #1558, #2019 and #2030 — and it wasn't only about the ~2.7 MB. From #2019:

> Adding @types/node to my web projects ends up polluting the global scope… `setTimeout` infers its return type as `Timeout` instead of `number`.

It was resolved upstream in 8.4.0, but every package still on `^7` keeps its consumers pinned behind that fix, since no `^7.x` release can reach it.

I ran into this from the other direction while auditing a dependency tree that had both `proto-loader` and another `^7` consumer: as long as *any* package in the tree requires `^7`, a second consumer moving to `^8` just installs both majors side by side rather than saving anything —

```
v8.7.2  3.8M  node_modules/protobufjs
v7.6.5  3.1M  node_modules/.../node_modules/protobufjs
```

so the ecosystem benefit only materialises once the remaining `^7` holdouts move. `proto-loader` is one of the more widely-depended-on ones, which is why this PR matters beyond Edition 2024 support.

No action needed on my part — just adding the numbers in case they're useful for prioritising the review. Happy to test against a real tree if that helps.

hypnoce:
Is there anything blocking this PR ? It's needed by other tools like Bruno or Insomnia to correctly handle reflection API with proto editions. Thanks !

PR #3070, opened 2026-07-28
grpc-js-xds: Fix weighted cluster total calculation

### LINKED ISSUE #3069: grpc-js-xds: WeightedCluster with omitted `total_weight` can select no cluster and crash in resolver
### Problem description

`@grpc/grpc-js-xds` can crash when an xDS route uses `weighted_clusters`, omits the deprecated `total_weight` field, and the cluster weights do not add up to 100.

For example, a route containing one cluster with weight `1` fails with:

```text
TypeError: Cannot read properties of undefined (reading 'ref')
    at Object.invoke (.../node_modules/@grpc/grpc-js-xds/build/src/resolver-xds.js:343:36)
```
The same xDS route works with `grpcurl`/gRPC Go.

The resolver currently defaults an absent `total_weight` to `100`:

```ts
routeAction = new WeightedClusterRouteAction(
  weightedClusters,
  route.route!.weighted_clusters!.total_weight?.value ?? 100,
  // ...
);
```
`WeightedClusterRouteAction` therefore generates a random value in `[0, 100)`. With one cluster of weight `1`, only values in `[0, 1)` select that cluster. The remaining ~99% return an empty cluster name.

The resolver then looks up that empty name and dereferences `undefined`:

```ts
const clusterResult = action.getCluster();
const clusterRef = this.clusterRefs.get(clusterResult.name)!;
clusterRef.ref();
```

The generated Envoy API documentation says that `total_weight` is deprecated and clients should use the sum of all cluster weights.

Relevant code:

- https://github.com/grpc/grpc-node/blob/master/packages/grpc-js-xds/src/resolver-xds.ts
- https://github.com/grpc/grpc-node/blob/master/packages/grpc-js-xds/src/route-action.ts
- https://github.com/grpc/grpc-node/blob/master/packages/grpc-js-xds/src/generated/envoy/config/route/v3/WeightedCluster.ts

### Reproduction steps

1. Configure an RDS route similar to:

```yaml
route:
  weighted_clusters:
    clusters:
      - name: test-cluster
        weight:
          value: 1
    # total_weight intentionally omitted
```

2. Register `@grpc/grpc-js-xds` and create an xDS client:

```js
const grpc = require('@grpc/grpc-js');
require('@grpc/grpc-js-xds').register();

const client = new grpc.Client(
  'xds:///test-service',
);
```

3. Make repeated RPCs through the route.

4. Most calls fail with:

```text
TypeError: Cannot read properties of undefined (reading 'ref')
```

Setting the cluster weight to `100`, explicitly setting `total_weight: 1`, or using a non-weighted single-cluster route avoids the failure.

### Expected behavior

When `total_weight` is omitted, the client should use the sum of the supplied cluster weights.

For one cluster with weight `1`, that cluster should receive 100% of requests.

The resolver should also return a controlled error instead of throwing if cluster selection somehow produces an unknown or empty cluster name.

### Environment

- OS: Linux x86_64, Kubernetes
- Node version: **[fill in]**
- `@grpc/grpc-js` version: **[fill in]**
- `@grpc/grpc-js-xds` version: **[fill in]**
- Installation method: npm

### Suggested fix

Calculate the denominator from the cluster weights:

```ts
const totalWeight = weightedClusters.reduce(
  (sum, cluster) => sum + cluster.weight,
  0
);
```

Use that value when creating `WeightedClusterRouteAction`, rather than defaulting an omitted `total_weight` to `100`.

It would also be helpful to guard the `clusterRefs.get()` result so an invalid selection produces a gRPC status error instead of an uncaught `TypeError`.

### Additional context

A configuration with weights `1` and `99` works because the weights happen to total 100. The failure occurs when the supplied weights have a nonzero sum other than 100 and `total_weight` is omitted.
```

### PR DESCRIPTION
**Pull request by gunsch**
Fixes #3069.

The xDS resolver previously defaulted to a total weight of `100` if the `total_weight` field is empty (which is [deprecated](https://github.com/envoyproxy/envoy/blob/742a3b297d7cbc31bd9266ef6adcce186b271df2/api/envoy/config/route/v3/route_components.proto#L483)).

Instead, sum the cluster weights to get the total before using it in calculations.

Without this, resolution of a cluster with total weights of "1" would fail with:

```
TypeError: Cannot read properties of undefined (reading 'ref')
    at Object.invoke (.../node_modules/@grpc/grpc-js-xds/build/src/resolver-xds.js:343:36)
```

Testing:

- added regression test

### DIFF
--- a/packages/grpc-js-xds/src/resolver-xds.ts
+++ b/packages/grpc-js-xds/src/resolver-xds.ts
@@ -332,7 +332,16 @@ class XdsResolver implements Resolver {
             }
             weightedClusters.push({name: clusterWeight.name, weight: clusterWeight.weight?.value ?? 0, dynamicFilterFactories: extraFilterFactories});
           }
-          routeAction = new WeightedClusterRouteAction(weightedClusters, route.route!.weighted_clusters!.total_weight?.value ?? 100, {name: [], timeout: timeout, retryPolicy: retryPolicy}, hashPolicies);
+          const totalWeight = weightedClusters.reduce(
+            (sum, cluster) => sum + cluster.weight,
+            0
+          );
+          routeAction = new WeightedClusterRouteAction(
+            weightedClusters,
+            totalWeight,
+            {name: [], timeout: timeout, retryPolicy: retryPolicy},
+            hashPolicies
+          );
           break;
         }
         default:
@@ -356,7 +365,15 @@ class XdsResolver implements Resolver {
         for (const {matcher, action} of matchList) {
           if (matcher.apply(methodName, metadata)) {
             const clusterResult = action.getCluster();
-            const clusterRef = this.clusterRefs.get(clusterResult.name)!;
+            const clusterRef = this.clusterRefs.get(clusterResult.name);
+            if (!clusterRef) {
+              return {
+                methodConfig: clusterResult.methodConfig,
+                pickInformation: {cluster: '', hash: ''},
+                status: status.UNAVAILABLE,
+                dynamicFilterFactories: clusterResult.dynamicFilterFactories,
+              };
+            }
             clusterRef.ref();
             const onCommitted = () => {
               clusterRef.unref();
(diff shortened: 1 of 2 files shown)


gunsch:
hey @murgatroid99 , would you be able to take a look at this? looks like it's mostly your file!

murgatroid99 (approved):

murgatroid99:
This fix is out in version 1.14.1

PR #3074, opened 2026-08-13
Fixed Resource timer and cache issue

zarinn3pal:
Implemented the following fixes to resolve the issue regarding furthur updates ignored after LDS resource deletion.

1. Zombie Timers: Cancel timers on stream termination to prevent false `DOES_NOT_EXIST` states.
2. Dormant Timers: Start timers immediately for ADS calls on already `READY` clients.
3. Invalid Cache: Clear caches on `DOES_NOT_EXIST` so identical restored resources are properly recognized.
4. Added a test for the LDS deletion and restoration recovery flow.

### DIFF
--- a/packages/grpc-js-xds/src/xds-client.ts
+++ b/packages/grpc-js-xds/src/xds-client.ts
@@ -150,6 +150,7 @@ class ResourceTimer {
     if (!resourceState) {
       return;
     }
+    resourceState.cachedResource = null;
     resourceState.meta.clientStatus = 'DOES_NOT_EXIST';
     for (const watcher of resourceState.watchers) {
       watcher.onResourceDoesNotExist();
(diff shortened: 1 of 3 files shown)


packages/grpc-js-xds/src/xds-client.ts
```diff
@@ -150,6 +150,7 @@ class ResourceTimer {
     if (!resourceState) {
       return;
     }
+    resourceState.cachedResource = null;
```
murgatroid99: This should instead be

```ts
if (resourceState.cachedResource !== null) {
  return;
}
```

The purpose of the timer is to say "the server has not sent this resource yet, so we treat it as nonexistent". It is not intended to invalidate resources we have already seen.

murgatroid99 (approved):

Issue #2785, opened 2024-07-04
gRPC-JS client hangs when target server restarts


sheinbergon:
### Problem description

gRPC js client halts on server disconnections

### Reproduction steps

We have a server (server API is gRPC, using malijs), let's call it **A** which calls a gRPC target proxied by an NLB and nginx gRPC ingress, let's call it **B**. If we go ahead and restart the target instances of **B**, The client might sometime hang, resulting in the application **A** failing to function. All flows calling to A that will result in **A** calling **B** will hang

The moment we restart **A**, it becomes functioning, meaning it's a client state issue.

### Environment

 - Debian 11/12 based containers 
 - Node version: 18.20.x and 22.2.x
 - NVM based installation
 - gRPC-JS 1.19.15
-  Client options: `{ 'grpc.keepalive_time_ms': 10_000, 'grpc.max_receive_message_length': 64 * 1024 * 1024 }`

### Additional context

We've activated gRPC tracing on server **A**

here's what a normal server selection trace looks like: 

```
D 2024-07-04T09:37:26.487Z | v1.9.15 29 | subchannel | (157) 10.41.10.107:443 READY -> IDLE
D 2024-07-04T09:37:26.487Z | v1.9.15 29 | subchannel_refcount | (157) 10.41.10.107:443 refcount 2 -> 1
D 2024-07-04T09:37:26.487Z | v1.9.15 29 | pick_first | READY -> IDLE
D 2024-07-04T09:37:26.487Z | v1.9.15 29 | resolving_load_balancer | dns:service-b.internal.com:443 READY -> IDLE
D 2024-07-04T09:37:26.487Z | v1.9.15 29 | connectivity_state | (2) dns:service-b.internal.com:443 READY -> IDLE
D 2024-07-04T09:37:26.487Z | v1.9.15 29 | dns_resolver | Looking up DNS hostname service-b.internal.com
D 2024-07-04T09:37:26.487Z | v1.9.15 29 | resolving_load_balancer | dns:service-b.internal.com:443 IDLE -> CONNECTING
D 2024-07-04T09:37:26.487Z | v1.9.15 29 | connectivity_state | (2) dns:service-b.internal.com:443 IDLE -> CONNECTING
D 2024-07-04T09:37:26.492Z | v1.9.15 29 | dns_resolver | Resolved addresses for target dns:service-b.internal.com:443: [10.41.33.32:443,10.41.22.168:443,10.41.10.107:443]
[...]

PR #3078, opened 2026-09-09
grpc-js: Remove calls from the pick queue immediately when they end

murgatroid99:
This is a defense-in-depth fix for the bug reported in https://github.com/grpc/grpc-node/issues/2785#issuecomment-5607201959. If, due to another bug, a channel gets stuck in a non-READY state for a long period of time, call references can accumulate in the pick queue, and they aren't removed immediately if they end locally due to cancellation or a deadline. With this change, calls are always immediately removed from the queue when they end.

Also, as suggested in that comment, I made the pick queue a set to make that removal efficient.

### DIFF
--- a/packages/grpc-js/src/internal-channel.ts
+++ b/packages/grpc-js/src/internal-channel.ts
@@ -342,7 +342,7 @@ export class InternalChannel {
       },
       updateState: (connectivityState: ConnectivityState, picker: Picker) => {
         this.currentPicker = picker;
-        const queueCopy = {...this.pickQueue};
+        const queueCopy = this.pickQueue;
         this.pickQueue = new Set();
         if (queueCopy.size > 0) {
           this.callRefTimerUnref();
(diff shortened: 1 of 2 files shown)


packages/grpc-js/src/internal-channel.ts
```diff
@@ -342,9 +342,9 @@ export class InternalChannel {
       },
       updateState: (connectivityState: ConnectivityState, picker: Picker) => {
         this.currentPicker = picker;
-        const queueCopy = this.pickQueue.slice();
-        this.pickQueue = [];
-        if (queueCopy.length > 0) {
+        const queueCopy = {...this.pickQueue};
+        this.pickQueue = new Set();
```
Sermelyan: do we need deep copy here? maybe just switch the ref? should be safe in this context

```suggestion
        const queueCopy = this.pickQueue;
        this.pickQueue = new Set();
```

murgatroid99:
This is out in version 1.14.5

PR #3083, opened 2026-09-14
feat(grpc-js): add option to disable caller stack trace capture

olavloite:
Currently, Client constructs a new Error on every RPC invocation to capture the caller's stack trace in case the call fails. Constructing an Error captures a full V8 stack trace synchronously on every call, which adds noticeable CPU and GC overhead in high-throughput applications.

This adds a channel option `grpc-node.enable_caller_stack_traces` (enabled by default) that allows clients to opt out of caller stack trace capture. When disabled, the client omits caller stack frame capture during call dispatch, reducing per-RPC overhead.

The default behavior of the gRPC client remains unchanged by this PR.


packages/grpc-js/src/channel-options.ts
```diff
@@ -67,6 +67,12 @@ export interface ChannelOptions {
   'grpc-node.retry_max_attempts_limit'?: number;
   'grpc-node.flow_control_window'?: number;
   'grpc.server_call_metric_recording'?: number;
+  /**
+   * Whether to construct an Error object on every call to capture the caller's
+   * stack trace. Enabled by default (1). Can be set to 0 to disable and avoid
+   * stack trace capture overhead for every RPC.
+   */
+  'grpc.enable_caller_stack_traces'?: number;
```
murgatroid99, line 75: For consistency with other channel options that are specific to this implementation, and have no reason to ever be shared with the C++ implementation, this option name should be `grpc-node.enable_caller_stack_traces`.

olavloite: Done

PR #3084, opened 2026-09-14
chore(grpc-js): unify call numbers and avoid disabled trace allocations

olavloite:
Previously, each layer of an RPC (ResolvingCall, RetryingCall, LoadBalancingCall, and Http2SubchannelCall) allocated its own separate call number via getNextCallNumber(). This made correlating logs across layers difficult and consumed multiple IDs per logical RPC.

Additionally, dynamic trace arguments such as string concatenations, JSON.stringify(), and deadline formatting were evaluated eagerly at call sites even when tracing was disabled, adding unnecessary garbage collection pressure on the fast path.

This change:
1. Passes the initial callNumber from ResolvingCall down through RetryingCall, LoadBalancingCall, and Http2SubchannelCall so that all layers of an attempt share the same call number. Subsequent retries and hedged attempts allocate a new call number per attempt.
2. Short-circuits isTracerEnabled() when no tracers are active and guards dynamic trace argument evaluation behind traceEnabled checks so that disabled tracers incur zero string or formatting allocations.


murgatroid99 (approved):

murgatroid99:
This is out in version 1.14.5.

PR #3085, opened 2026-09-14
Added Node version switcher instead of nvm-ps to fix windows ci issue

zarinn3pal:
Windows ci failures occured for grpc-node because the legacy [nvm-ps ](https://github.com/grumpycoders/nvm-ps)installation command in run-tests.bat attemped to use PowerShell's Invoke-WebRequest without the `UseBasicParsing` flag. Replaced nvm-ps with [nvs](https://github.com/jasongin/nvs)(maintained node version switcher library) . [nvm-windows](https://github.com/nvm-windows/nvm) was considered and avoided because of its architectural dependence on .exe installers since v2.0.0.


murgatroid99 (approved):
This change seems to have solved the relevant problem. The job is still failing, but I believe it will be fixed by #3030.

